PR #1, opened 2024-08-30
chore: add depenabot configuration

kjaskiewiczz:


.github/dependabot.yml
```diff
@@ -0,0 +1,50 @@
...
+    groups:
+      backend-golang-dependencies:
+        patterns:
+          - "*"
+    reviewers:
+      - "mendersoftware/backend-dependabot-reviewers"
+  - commit-message:
+      prefix: chore
+    directory: /backend
+    package-ecosystem: docker
```
mzedel: we can try, but I'm afraid this won't pick up the versions in the per service Dockerfiles... looking at https://github.com/dependabot/dependabot-core/issues/2178 we will likely have to list the directories here.

kjaskiewiczz: yes, I'm preparing config with all `directories` listed

.github/dependabot.yml
```diff
@@ -46,3 +46,61 @@ updates:
...
+      - "mendersoftware/backend-dependabot-reviewers"
+  - commit-message:
+      prefix: chore
+    directories:
+      - /backend/tests
+      - /backend/services/deviceconnect/tests
+      - /backend/services/iot-manager/tests
+      - /backend/services/reporting/tests
+      - /backend/services/deployments/tests/mock-workflows
+    package-ecosystem: pip
```
merlin-northern, line 98: does this mean updating the deps in the tests? as in: keeping newest requests library?

olehermanse: This looks correct.

@merlin-northern Please keep in mind that it's required to keep all dependencies up to date, there is no general exception for test dependencies or infra dependencies. If you have some specific folder or repo or test which cannot be upgraded, that should be an explicit decision with a documented reason (i.e. talk to me and the security team).

alfrunes (approved):

mzedel (approved):

PR #2, opened 2024-08-30
Add support for multiplatform builds

alfrunes:
See `README.make`.


.gitlab-ci.yml
```diff
@@ -44,6 +48,22 @@ build:backend:dist:
...
+    - job: build:binaries
+      artifacts: true
+  rules:
+    - changes:
+        paths: ["backend/**/*"]
+        compare_to: "${RULES_CHANGES_COMPARE_TO_REF}"
+  tags:
+    - hetzner-amd-beefy
+  script:
+    - make docker
```
mzedel: do we really want to rebuild all the services for a single service change? This will unnecessarily slow down the pipeline

frontend/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	docker build $(DOCKER_BUILD_ARGS) \
+		--build-arg BIN_FILE=$(DOCKER_BIN) \
+		--platform $(DOCKER_PLATFORM) \
+		-f Dockerfile \
+		-t $(DOCKER_TAG) \
+		.
+
+build:
+	npm install && \
+    npm run build
```
mzedel:
```suggestion
build:
	npm ci && \
    npm run build
```
otherwise you won't get reliably reproducible results.

.gitlab-ci.yml
```diff
@@ -44,6 +48,22 @@ build:backend:dist:
     paths:
       - dist.tar.gz
 
+build:docker:
```
mzedel:
```suggestion
build:backend:docker:
```

mzedel (changes requested):

.gitlab-ci.yml
oldgiova: Additionally, it could be wonderful if we can use parallel jobs when we have to build multiple changed services.

PR #3, opened 2024-08-30
ci: added generators for build, test & helm sync

mzedel:
@alfrunes as RFC, resulting in: https://gist.github.com/mzedel/690a2fc1a3bde898ac17b10f2c912b22 - to be included via https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html#dynamic-child-pipelines and potentially extended with the per-service coverage publish jobs...
BUT! I just assume we would get the Dockerfile/ pkg issue to work somehow 🤞

### DIFF
--- a/ci-generate-build-n-test.js
+++ b/ci-generate-build-n-test.js
@@ -1,0 +1,127 @@
+const fs = require("fs");
+
+const services = [
+  "auditlogs",
+  "create-artifact-worker",
+  "deployments",
+  "deviceauth",
+  "deviceconfig",
+  "deviceconnect",
+  "devicemonitor",
+  "inventory",
+  "mender-gateway",
+  "tenantadm",
+  "useradm",
+  "workflows",
+];
+
+const pipelineTemplate = (jobs) => `stages:
+  - build
+  - test
+${jobs.join("")}
+`;
+
+const getChangesRule = ({
+  RULES_CHANGES_COMPARE_TO_REF,
+  service,
+}) => `- changes:
+        paths: ["backend/services/${service}/**/*", "backend/pkg/**/*"]
+        compare_to: "${RULES_CHANGES_COMPARE_TO_REF}"`;
+
+const buildTemplate = ({
+  CI_COMMIT_REF_NAME,
+  CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX,
+  CI_PIPELINE_ID,
+  CI_REGISTRY_IMAGE,
+  DOCKER_PUBLISH_COMMIT_TAG,
+  RULES_CHANGES_COMPARE_TO_REF,
+  service,
+}) => `
+build:${service}:docker-multiplatform:
+  image: "registry.gitlab.com/northern.tech/mender/mender-test-containers:docker-multiplatform-buildx-v1-master"
+  stage: build
+  rules:
+    ${getChangesRule({ RULES_CHANGES_COMPARE_TO_REF, service })}
+    - if: '$CI_COMMIT_TAG =~ /^saas-[a-zA-Z0-9.]+$/'
+      when: never
+    - when: on_success
+  tags:
+    - hetzner-amd-beefy
+  services:
+    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:20.10.21-dind
+      alias: docker
+  needs: []
+  variables:
+    DOCKER_BUILDKIT: 1
+    GITLAB_REGISTRY_TAG: '${CI_REGISTRY_IMAGE}:${CI_PIPELINE_ID}'
+  before_script:
+    - *dind-login
+    - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+  script:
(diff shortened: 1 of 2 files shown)


ci-generate-build-n-test.js
```diff
@@ -0,0 +1,127 @@
...
+      --tag \${GITLAB_REGISTRY_TAG}
+      --tag ${CI_REGISTRY_IMAGE}:${CI_COMMIT_REF_NAME}
+      --file backend/services/${service}/Dockerfile
+      --build-arg GIT_COMMIT_TAG="${DOCKER_PUBLISH_COMMIT_TAG}"
+      --platform $MULTIPLATFORM_PLATFORMS
+      --provenance false
+      --push
+      \${EXTRA_DOCKER_ARGS}
+      .
+`;
```
alfrunes, line 76: Could we instead use the `make` targets? That makes it easier to reproduce locally and make it more obvious to compare local vs CI builds.
```
make -C backend/services/${service} docker
```

ci-generate-build-n-test.js
```diff
@@ -0,0 +1,127 @@
...
+const buildTemplate = ({
+  CI_COMMIT_REF_NAME,
+  CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX,
+  CI_PIPELINE_ID,
+  CI_REGISTRY_IMAGE,
+  DOCKER_PUBLISH_COMMIT_TAG,
+  RULES_CHANGES_COMPARE_TO_REF,
+  service,
+}) => `
+build:${service}:docker-multiplatform:
```
alfrunes, line 40: Building services in different jobs, although parallel is more wasteful as it cannot leverage the local go cache.

ci-generate-build-n-test.js
```diff
@@ -0,0 +1,127 @@
...
+  "deployments",
+  "deviceauth",
+  "deviceconfig",
+  "deviceconnect",
+  "devicemonitor",
+  "inventory",
+  "mender-gateway",
+  "tenantadm",
+  "useradm",
+  "workflows",
```
alfrunes, line 15: This list contains enterprise services and `mender-gateway` which is not part of this repository.

ci-generate-build-n-test.js
```diff
@@ -0,0 +1,127 @@
...
+    - if: '$CI_COMMIT_TAG =~ /^saas-[a-zA-Z0-9.]+$/'
+      when: never
+    - when: on_success
+  tags:
+    - hetzner-amd-beefy
+  needs:
+    - job: build:${service}:docker-multiplatform
+      artifacts: true
+  script:
+    - go test -trimpath -ldflags -s -w ./...
```
alfrunes, line 97: How do we deal with the test with the same toolchain version as we're building. Today we're not doing a very good job at that - versions managed inside gitlab tend to lag major versions (Go term - not semver) behind the Dockerfile.
```suggestion
    - go test -trimpath -ldflags='-s -w' ./...
```

ci-generate-build-n-test.js
```diff
@@ -0,0 +1,127 @@
...
+  stage: build
+  rules:
+    ${getChangesRule({ RULES_CHANGES_COMPARE_TO_REF, service })}
+    - if: '$CI_COMMIT_TAG =~ /^saas-[a-zA-Z0-9.]+$/'
+      when: never
+    - when: on_success
+  tags:
+    - hetzner-amd-beefy
+  services:
+    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:20.10.21-dind
```
alfrunes, line 51: This version of docker is over 3 years old. Could we make the docker version more transparent in the global `variable` section?
```suggestion
    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-dind
```

ci-generate-sync.js
```diff
@@ -0,0 +1,97 @@
...
+  CONTAINER_NAME,
+  RULES_CHANGES_COMPARE_TO_REF,
+  service,
+}) => `
+trigger:mender-helm-version-bump:${service}:staging:
+  extends:
+    - .trigger:mender-helm-version-bump
+  rules:
+    ${getChangesRule({ RULES_CHANGES_COMPARE_TO_REF, service })}
+    - if: $CI_COMMIT_BRANCH =~ /^(staging)$/
```
alfrunes, line 58: Maybe this is a topic for the QA meeting, but IMO we could drop the staging branch and deploy to staging from main.
```suggestion
  rules:
    ${getChangesRule({ "HEAD~1", service })}
    - if: $CI_COMMIT_BRANCH == main
```

alfrunes (commented):
There are some jobs missing (none for frontend, build and run backend acceptance and integration tests), as well as publishing so the trigger jobs would not work.
Another issue we will eventually have to deal with is how to deal with test coverage. By excluding jobs, it will also exclude test coverage unless we have a cleaver way of caching coverage. Or should we drop coverage collection?

oldgiova:
With regard to the Helm Chart sync, although I consider this a great rework, some key points have to be addressed before:

1. the idea is to use release tags from `main`; this mean that the helm:sync job should be extended to run just once, and increment every services in the helm chart (requires rework on the helm chart side)
2. you create a new release tag for every services, then the `deviceconnect` service is always restarted on every release, even without changes on that service. This is a concern for long living websocket connections @alfrunes. 
3. The helm sync should run from the `mender-server-enterprise` repo only ;)

PR #6, opened 2024-09-02
MEN-7461 - fix: fixed an issue that prevented retrieving group devices & related reports

mzedel:
Ticket: MEN-7461
Changelog: Title


mineralsfree (approved):
It looks reasonable; it's better to separate fixes and refactoring into different commits.

PR #10, opened 2024-09-02
MEN-6894 - feat: made search results reopen whenever the search field is clicked again and has a search term

mzedel:
Ticket: MEN-6894
Changelog: Title


olehermanse (approved):
I tested the current behavior on hosted Mender, and I agree that this would be a good improvement.

PR #17, opened 2024-09-02
Build images in standalone mode

alfrunes:
Loosely based on #2 
TODOs:
- [x] Update CI pipelines
  - [x] Replace build binary job with docker jobs
  - [x] Use Gitlab registry
  - [x] Update test:unit target
    - Can we use an image built by the pipeline in a later job?
      - Can export the builder context and use it in gitlab template to make sure we test using the same Go toolchain.
- [x] Update compose setup
  - [x] Image references from environment variables (aligned with make)

I know there's a lot of commits and changes, so I'l try to summarize a TL;DR below of the changes (I tried to document what I do in the commit messages too).
 * The `Dockerfile`s now has a build stage
   * As a side-effect, all the images can be built inside compose: `docker compose -f dev/docker-compose.yml build`
 * Changed the build interface:
    * `docker` is the default build target
    * `docker` does not depend on the build stage, but the `Dockerfile`s themselves are self contained.
 * New targets:
   * `docker-pull` to pull all the images
     * Gitlab test targets no longer rebuild the images
   * `docker-publish` to publish the images upstream
     * Copies the images built from the local registry to the upstream registry.
 * New gitlab stage to publish the images
   * Runs on all pipelines for **protected** branches and tags
     * For Go semver tags (with `v` prefix) it publishes images: `vX.Y.Z`, `vX.Y`, `vX`.
     * All other protected references are published using the same reference name.


alfrunes:
I'm not sure what to do about the `test:backend:unit` job. To ensure we are testing using the same toolchain we build, I can see two options:
1. Add an `ARG GOTOOLCHAIN` to all the `Dockerfile`s which we assign to the `GOLANG_VERSION` used in the CI pipelines.
  * The version defined in `.gitlab-ci.yml` will become the source of truth.
  * Dependabot will not be able to manage golang versions.
  * Trickier to maintain consistent version when building for development (locally) and building for production.
2. Build any of the `Dockerfile`s and target the build stage and publish it internally in the gitlab registry and use that image to run tests.
  * More overhead in terms of building and exporting it as well as storing the artifact.
  * Dockerfiles will remain source of truth
    * Dependabot can update the versions.
Any other suggestions are welcome.

alfrunes:
I went ahead and implemented (1) above 7f98fefb546cb4300c3424177ad2d4c7fbe6b806

oldgiova:
> 2. Build any of the `Dockerfile`s and target the build stage and publish it internally in the gitlab registry and use that image to run tests.

Although this requires more overhead, it seems to me the most clean solution. Also the overhead is decreased because of caching previous builds.

mzedel (approved):
Just for the sake of not even risking missing an update I would vote for option 2, export the build stage for the services as artifacts (re-running the build command for this should be very quick as all layers would be cached) and then run the tests from/ in there, but we can still change this down the line as well - so IMO not anything to worry too much about.
Apart from that, with the quicker pipeline overall I don't think the local/ CI pipeline discrepancy will be much of an issue, as it likely will be much easier than before to just run CI for a quick check 🙈.

alfrunes:
Ok, I'll try with (2) above :whale:

alfrunes:
@oldgiova @mzedel Check it out: 72e1a5f87f7dc8bc20475432258e610b02d7791e
It's a little hacky, but I didn't want to pull in the rest of the build stage so I added a small test to check that they're consistent. But it seems to be working fine: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/7764002898

oldgiova:
> @oldgiova @mzedel Check it out: [72e1a5f](https://github.com/mendersoftware/mender-server/commit/72e1a5f87f7dc8bc20475432258e610b02d7791e) It's a little hacky, but I didn't want to pull in the rest of the build stage so I added a small test to check that they're consistent. But it seems to be working fine: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/7764002898

Actually I like this kind of hacky! For sure is consistent, and preserves the single source of truth.

mzedel:
that looks indeed hacky... I would have expected something more along the lines of:
```
    - docker buildx build
      --tag ${DOCKER_REPOSITORY}:tests
      --target build
      --output type=docker,dest=testsImage.tar
  artifacts:
    paths:
      - testsImage.tar
```
and then in the unit tests job:
```
  needs:
    - job: <build-job>
      artifacts: true
  script:
    - docker load -i testsImage.tar
    - docker run --entrypoint "/bin/sh" testsImageTag -c "go test ......"
```

alfrunes:
> that looks indeed hacky... I would have expected something more along the lines of:
> 
> ```
>     - docker buildx build
>       --tag ${DOCKER_REPOSITORY}:tests
>       --target build
>       --output type=docker,dest=testsImage.tar
>   artifacts:
>     paths:
>       - testsImage.tar
> ```
> 
> and then in the unit tests job:
> 
> ```
>   needs:
>     - job: <build-job>
>       artifacts: true
>   script:
>     - docker load -i testsImage.tar
>     - docker run --entrypoint "/bin/sh" testsImageTag -c "go test ......"
> ```

The problem is which Dockerfile to pick? We have one Dockerfile per service and if we pick one we are ignoring all the others. And building the `builder` target will also build the binary among other things (we can of course restructure the dockerfiles to isolate this to a separate stage).

mzedel:
Then this would likely also have to go through the makefiles as the Dockerfile is picked in there, then another go through the makefiles to load and run the images for the services 🤨

alfrunes:
@mzedel Something like this 7685da57d976238ae0557b87b397da8c7c559f37?

mzedel:
Sure, if the services are so similar & deployments can cover all the variations already that seems like a pretty nice middle ground solution to me 👍

oldgiova (approved):
Really impressive job! Nice one!

alfrunes:
> Sure, if the services are so similar & deployments can cover all the variations already that seems like a pretty nice middle ground solution to me 👍

It makes no difference which service I'd pick. It is just the `golang:1.23.1` image with the binary. Because of the caching, none of the modules pulled at build time are stored in the layer created by the `RUN` (compile) command in the Dockerfile.

tranchitella (approved):

merlin-northern (approved):
you are my hero. what is the story behind the choice of "localhost:5000" ?

alfrunes:
> what is the story behind the choice of "localhost:5000" ?

I didn't want docker compose to try to attempt to pull from an upstream repository by default, and port 5000 seems to be the default insecure port number for self-hosted docker registry.

PR #24, opened 2024-09-05
Chore/gui ci

mzedel:


frontend/pipeline.yml
```diff
@@ -0,0 +1,302 @@
...
+    - changes:
+        paths: ['frontend/**/*']
+        compare_to: '${RULES_CHANGES_COMPARE_TO_REF}'
+      when: always
+  script:
+    - cd frontend
+    - docker context create ci;
+    - docker builder create --name ci-builder ci;
+    - unset DOCKER_HOST;
+    - docker buildx create --name gui-builder --driver=docker-container ci
```
alfrunes: Isn't this recreating the builder from the `before_script` in the template?

frontend/pipeline.yml
```diff
@@ -0,0 +1,302 @@
...
+    # needed for the license generation step
+    - docker build
+      --tag ${FRONTEND_REPOSITORY}:base
+      --target base
+      --builder=gui-builder
+      --output type=docker,dest=baseImage.tar
```
alfrunes: Could also be pushed to the internal registry to allow reuse of existing layers.
```suggestion
    - docker build
      --tag ${FRONTEND_REPOSITORY}:base
      --target base
      --builder=gui-builder
      --tag ${CI_REGISTRY_IMAGE}:frontend-license-${CI_PIPELINE_ID}
```

frontend/pipeline.yml
```diff
@@ -0,0 +1,302 @@
...
+      .
+    - echo ${CI_REGISTRY_IMAGE}
```
alfrunes: ?

frontend/pipeline.yml
```diff
@@ -0,0 +1,302 @@
...
+      --provenance false
+      --push
+      .
+  artifacts:
+    paths:
+      - baseImage.tar
+
+.template:test:frontend:acceptance:
+  stage: e2e-test
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/tiangolo/docker-with-compose
```
alfrunes: This is an old discontinued image. Please use `docker:${DOCKER_VERSION}-cli`.

frontend/pipeline.yml
```diff
@@ -0,0 +1,302 @@
...
+  tags:
+    - hetzner-amd-beefy
+
+publish:frontend:docker:
+  extends: publish:backend:docker
+  variables:
+    MENDER_PUBLISH_REGISTRY: docker.io
+  before_script:
+    - export DOCKER_TAG=${CI_REGISTRY_IMAGE}:frontend-${CI_PIPELINE_ID}
+    - docker pull ${DOCKER_TAG}
```
alfrunes:
```suggestion
    - skopeo login ${CI_REGISTRY} ...
    - skopeo login ${MENDER_PUBLISH_REGISTRY} ...
```

frontend/pipeline.yml
```diff
@@ -0,0 +1,300 @@
...
+  needs:
+    - job: build:frontend:docker
+      artifacts: true
+  services:
+    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:dind
+      alias: docker
+  script:
+    - docker load -i baseImage.tar
+    - docker run --rm --entrypoint "/bin/sh" -v $(pwd):/extract ${FRONTEND_REPOSITORY}:base -c "npm run disclaim && cp disclaimer.txt /extract/"
+    - docker image rm disclaimer
```
alfrunes: Following the suggestion above. This can be simplified significantly...
```suggestion
  image: ${CI_REGISTRY_IMAGE}/frontend-licenses-${CI_PIPELINE_IID}
  needs:
    - job: build:frontend:docker
      artifacts: false
  script:
    - npm run disclaim
```

frontend/pipeline.yml
```diff
@@ -0,0 +1,300 @@
...
+    - skopeo copy --multi-arch all \
+      docker://${DOCKER_TAG} \
+      docker://${MENDER_PUBLISH_REGISTRY}/${FRONTEND_REPOSITORY}:${MENDER_PUBLISH_TAG} # covers vX.Y.Z + -fragment/ -build tags
+    - |
+      if echo -n "${MENDER_PUBLISH_TAG}" | grep -E '^v[0-9]+\.v[0-9]+\.[0-9]+$'; then
+        skopeo copy --multi-arch all docker://${DOCKER_TAG} docker://${MENDER_PUBLISH_REGISTRY}/gui:$(echo -n $MENDER_PUBLISH_TAG | cut -d . -f-2)
+        skopeo copy --multi-arch all docker://${DOCKER_TAG} docker://${MENDER_PUBLISH_REGISTRY}/gui:$(echo -n $MENDER_PUBLISH_TAG | cut -d . -f-1)
+        skopeo copy --multi-arch all docker://${DOCKER_TAG} docker://${MENDER_PUBLISH_REGISTRY}/gui:latest
+        ;;
+      fi
```
alfrunes, line 292: Why is this here?

frontend/pipeline.yml
```diff
@@ -0,0 +1,300 @@
...
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/tiangolo/docker-with-compose
+  rules:
+    - changes:
+        paths: ['frontend/**/*']
+        compare_to: '${RULES_CHANGES_COMPARE_TO_REF}'
+  needs:
+    - job: build:frontend:docker
+      artifacts: true
+  services:
+    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:dind
```
alfrunes, line 163: Can we use a consistent version of docker?
```suggestion
    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-dind
```

frontend/tests/e2e_tests/run
```diff
@@ -95,13 +81,13 @@ run_tests() {
         token=$(docker exec $containerid /usr/bin/tenantadm get-tenant --id $tenant | jq -r .tenant_token)
         jq --arg tenantToken "$token" '. + { "TenantToken": $tenantToken }' ${GUI_REPOSITORY}/tests/e2e_tests/dockerClient/mender.json > ${GUI_REPOSITORY}/tests/e2e_tests/dockerClient/mender-test.json
         # start a new client, using the obtained tenant token - starting it later (in the test suite) would require docker access in the test runner
-        docker run --name connect-client -d --network=gui-tests_mender \
+        docker run --name connect-client -d --network=gui-tests_default \
             -v ${GUI_REPOSITORY}/tests/e2e_tests/dockerClient/mender-test.json:/etc/mender/mender.conf \
             -v ${GUI_REPOSITORY}/tests/e2e_tests/dockerClient/mender-connect.json:/etc/mender/mender-connect.conf \
             mendersoftware/mender-client-docker-addons:mender-3.6.x
```
alfrunes: I don't know about much about this script, but isn't this redefining what's already in the compose file?
```suggestion
       $COMPOSE_CMD run --name connect-client --detach client
```

frontend/tests/e2e_tests/run
```diff
@@ -2,16 +2,10 @@
 set -e
 
 # by default just add minio, with COMPOSE_CMD_BASE this creates the standard onprem ST setup
-COMPOSE_CMD="docker compose -p gui-tests \
-             -f ${INTEGRATION_PATH}/docker-compose.yml \
-             -f ${INTEGRATION_PATH}/docker-compose.client.yml \
-             -f ${INTEGRATION_PATH}/docker-compose.demo.yml \
-             -f ${INTEGRATION_PATH}/docker-compose.storage.minio.yml"
-ENTERPRISE_COMPOSE=" -f ${INTEGRATION_PATH}/docker-compose.enterprise.yml"
+COMPOSE_CMD="docker compose -p gui-tests -f ${SERVER_ROOT}/dev/docker-compose.yml"
```
alfrunes: The `-p` argument could also be set in the compose file `name` top level attribute.

.gitlab-ci.yml
```diff
@@ -27,44 +27,41 @@ variables:
...
-    file: ".gitlab-ci-check-commits.yml"
-  - project: "Northern.tech/Mender/mendertesting"
-    file: ".gitlab-ci-github-status-updates.yml"
+    file:
+      - ".gitlab-ci-check-commits.yml"
+      - ".gitlab-ci-github-status-updates.yml"
+  - local: "/frontend/pipeline.yml"
 
 stages:
+  - pre-test
```
alfrunes: What is the purpose of this stage, and why are all the frontend test jobs running in it?

.gitlab-ci.yml
```diff
@@ -27,44 +27,41 @@ variables:
...
   - build
   - test
+  - e2e-test
   - publish
+  - version-bump
+  - test-deploy
```
alfrunes: These stages are unused.

.gitlab-ci.yml
```diff
@@ -27,44 +27,41 @@ variables:
...
+    file:
+      - ".gitlab-ci-check-commits.yml"
+      - ".gitlab-ci-github-status-updates.yml"
+  - local: "/frontend/pipeline.yml"
 
 stages:
+  - pre-test
   - build
   - test
+  - e2e-test
```
alfrunes: Isn't end-to-end tests a test?

frontend/pipeline.yml
mzedel: no, the other builder doesn't support docker image extraction and doesn't exist reliably

frontend/pipeline.yml
mzedel: stale

frontend/tests/e2e_tests/run
mzedel: without the assigned network the client won't reach the backend as it gets a different network assigned + depending on the test setup the client is torn down based independently of the composition (for staging e.g.)

frontend/pipeline.yml
mzedel: considering that we have missed upgrading these and they won't be tracked through dependabot I would rather not.

frontend/pipeline.yml
mzedel: because the gui is an OS enterprise component and was always published from OS as such, we can skip publishing it in enterprise

.gitlab-ci.yml
mzedel: The frontend tests used to be in a test stage that precedes the build stage (same as commit checks etc.), since most run quicker - so the pipeline can be stopped earlier. Since the backend changed this to move the test stage to post-build, this is running the tests pre-build 🤷‍♂️

.gitlab-ci.yml
mzedel: yep, related to the changed execution order from the general pipeline setup... will adjust.

frontend/pipeline.yml
alfrunes: That is exactly what I was intending to solve by introducing one version that gets reused instead of each job declaring their own image that is not looked after. Pulling the image through the cache may lead to stale versions of this tag and then you could have version skew between the Docker client and server.

frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
```diff
@@ -33,18 +35,43 @@ services:
...
+  inventory:
+    image: mendersoftware/inventory:main
+  iot-manager:
+    image: mendersoftware/iot-manager:main
+  useradm:
+    image: mendersoftware/useradm:main
+  workflows:
+    image: mendersoftware/workflows:main
+  workflows-worker:
+    image: mendersoftware/workflows:main
```
alfrunes: Just one thing I forgot last round.
Have you considered setting the environment variables?
```bash
MENDER_IMAGE_PREFIX="docker.io/mendersoftware"
MENDER_IMAGE_TAG="${CI_COMMIT_TAG:-main}"
```

frontend/tests/e2e_tests/docker-compose.e2e-tests.enterprise.yml
```diff
@@ -1,20 +1,43 @@
...
+  tenantadm:
+    image: mendersoftware/tenantadm:main
+  useradm:
+    image: mendersoftware/useradm-enterprise:main
     environment:
       USERADM_BASE_URL: "https://docker.mender.io"
+  workflows:
+    image: mendersoftware/workflows-enterprise:main
+  workflows-worker:
+    image: mendersoftware/workflows-enterprise-worker:main
```
alfrunes: Same as above.

.gitlab-ci.yml
alfrunes: You can use an empty `needs` to start a job immediately.

frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
```diff
@@ -1,25 +1,27 @@
+name: gui-tests
 services:
-  mender-gui-tests-runner:
+  gui-tests-runner:
     image: mendersoftware/mender-test-containers:gui-e2e-testing
     command: tail -f /dev/null
```
alfrunes, line 5: If this service is run-only you can set `scale: 0`.

frontend/tests/e2e_tests/run
alfrunes: The container is assigned to the `default` network by default where the backend now also lives.
Since run just starts a new container you can similarly kill it with `stop`, `rm` or even `kill` (if you're feeling brutal) compose commands.

mzedel:
can't reply to the network comment for some reason, so here:
The compose network will be project scoped, so you have to assign it for the client to reach the backend.

My main goal is to have the client interactions aligned across the different environments the client is running in - staging won't have a compose attached.

frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
mzedel: does it make a difference if this is omitted?

.gitlab-ci.yml
mzedel: or I can leave the jobs as they were and get the same behaviour with a smaller change 🤷‍♂️

frontend/tests/e2e_tests/run
mzedel: The compose network will be project scoped, so you have to assign it for the client to reach the backend.

My main goal is to have the client interactions aligned across the different environments the client is running in - staging won't have a compose attached.

frontend/pipeline.yml
```diff
@@ -0,0 +1,293 @@
...
+  script:
+    - cd frontend
+    - docker context create ci;
+    - docker buildx create --name gui-builder --driver=docker-container ci
+    # needed for the license generation step
+    - docker build
+      --tag ${FRONTEND_REPOSITORY}:base
+      --target base
+      --builder=gui-builder
+      --tag ${CI_REGISTRY_IMAGE}:frontend-base-${CI_PIPELINE_ID}
```
alfrunes, line 139: Sorry, I forgot the `--push` argument.
```suggestion
      --tag ${CI_REGISTRY_IMAGE}:frontend-base-${CI_PIPELINE_ID}
      --push
```

frontend/pipeline.yml
```diff
@@ -0,0 +1,293 @@
...
+      --tag ${CI_REGISTRY_IMAGE}:frontend-${CI_PIPELINE_ID}
+      --builder=gui-builder
+      --build-arg GIT_COMMIT_TAG="${CI_COMMIT_TAG}"
+      --platform ${DOCKER_PLATFORM:-linux/amd64}
+      --provenance false
+      --push
+      .
+  artifacts:
+    paths:
+      - baseImage.tar
```
alfrunes: This won't be needed anymore.
```suggestion
```

frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
alfrunes: Just makes it more apparent that it is a one-off job and not a service.

frontend/tests/e2e_tests/run
alfrunes: Ok, if you believe it is complicating things, then please disregard.

.gitlab-ci.yml
alfrunes: Up to you

frontend/pipeline.yml
mzedel: remove the publishing for now?

frontend/pipeline.yml
mzedel: Sure, OTOH we have had several pipelines that had it defined in the ci file and went stale, whereas others were diligently updated by hand and again others were left with the mutable tag and signaled problematic changes early... if you want to guarantee that we get this kept up to date, we can set it here as well - the state in the majority of the older repos says otherwise IMO.
As to the skewing: I don't think there will be version skew for any reasonable amount of time on this proxy and this image in particular.

oldgiova (approved):
LGTM, for the parts that I could check

PR #44, opened 2024-09-13
chore(workflows): Clean up unused MaybeExecuteGoTemplate

alfrunes:
There are no Go templates used in any of the workflow definitions. Let's skip the processing.

### DIFF
--- a/backend/services/workflows/app/worker/http.go
+++ b/backend/services/workflows/app/worker/http.go
@@ -57,9 +57,7 @@ func processHTTPTask(
 		form := url.Values{}
 		for key, value := range httpTask.FormData {
 			key = ps.ProcessJobString(key)
-			key = ps.MaybeExecuteGoTemplate(key)
 			value = ps.ProcessJobString(value)
-			value = ps.MaybeExecuteGoTemplate(value)
 			form.Add(key, value)
 		}
 		payloadString = form.Encode()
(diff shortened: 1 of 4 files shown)


backend/services/workflows/app/worker/http.go
```diff
@@ -57,9 +57,7 @@ func processHTTPTask(
 		form := url.Values{}
 		for key, value := range httpTask.FormData {
 			key = ps.ProcessJobString(key)
-			key = ps.MaybeExecuteGoTemplate(key)
 			value = ps.ProcessJobString(value)
-			value = ps.MaybeExecuteGoTemplate(value)
```
merlin-northern, line 62: how can it be unused, when it is used here?

alfrunes: There are no Go template directives (`{{ ... }}`) in any of the workflow definitions.

merlin-northern: I know that, but the commit message is misleading. that was added for some purpose, right? although I do not remember which.

merlin-northern (approved):
I approve not to block, but I would like to have a second opinion, this was added for a reason.

alfrunes:
> I approve not to block, but I would like to have a second opinion, this was added for a reason.

I added it, but never ended up using it.

backend/services/workflows/app/worker/http.go
tranchitella: I agree with @merlin-northern, this is a feature we have which we are removing because we are not currently using it.

kjaskiewiczz (approved):

PR #60, opened 2024-09-30
MEN-7551 - fix: added path for all devices on device page

mineralsfree:
Ticket: MEN-7551
Changelog: None

### DIFF
--- a/frontend/src/js/components/devices/device-groups.js
+++ b/frontend/src/js/components/devices/device-groups.js
@@ -146,7 +146,7 @@ export const DeviceGroups = () => {
     }
     // preset selectedIssues and selectedId with empty values, in case if remain properties are missing them
     let listState = { ...remainder };
-    if (statusParam && Object.values(DEVICE_STATES).some(state => state === statusParam)) {
+    if (statusParam && (Object.values(DEVICE_STATES).some(state => state === statusParam) || statusParam === 'any')) {
       listState.state = statusParam;
     }
 
--- a/frontend/src/js/utils/locationutils.js
+++ b/frontend/src/js/utils/locationutils.js
@@ -24,7 +24,6 @@ import {
   emptyFilter
 } from '@northern.tech/store/constants';
 
-import { routes } from '../components/devices/base-devices';
 import { deepCompare, getISOStringBoundaries } from '../helpers';
 
 const SEPARATOR = ':';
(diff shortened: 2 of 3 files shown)


frontend/src/js/components/leftnav.js
```diff
@@ -28,7 +28,7 @@ import DocsLink from './common/docslink';
 
 const listItems = [
   { route: '/', text: 'Dashboard', canAccess },
-  { route: '/devices', text: 'Devices', canAccess: ({ userCapabilities: { canReadDevices } }) => canReadDevices },
+  { route: '/devices/accepted', text: 'Devices', canAccess: ({ userCapabilities: { canReadDevices } }) => canReadDevices },
```
mzedel: this should remain `/devices` as it is expected to be set by the previously selected state, if that wasn't selected before the devices should always default to `accepted`

frontend/src/js/components/devices/device-groups.js
```diff
@@ -147,7 +147,7 @@ export const DeviceGroups = () => {
     }
     // preset selectedIssues and selectedId with empty values, in case if remain properties are missing them
     let listState = { ...remainder };
-    if (statusParam && Object.values(DEVICE_STATES).some(state => state === statusParam)) {
+    if (statusParam && Object.values(DEVICE_STATES).some(state => state === statusParam || statusParam === 'any')) {
```
mzedel:
```suggestion
    if (statusParam && (Object.values(DEVICE_STATES).some(state => state === statusParam) || statusParam === routes.allDevices.key)) {
```
to not have the check in every iteration there

frontend/src/js/utils/locationutils.js
```diff
@@ -219,7 +219,7 @@ export const formatDeviceSearch = ({ pageState, filters, selectedGroup }) => {
 export const generateDevicePath = ({ pageState }) => {
   const { state: selectedState } = pageState;
   const path = ['/devices'];
-  if (![routes.allDevices.key, ''].includes(selectedState)) {
+  if (![''].includes(selectedState)) {
```
mzedel:
```suggestion
  if (selectedState !== '') {
```
I think

mzedel (changes requested):

mzedel (approved):

PR #71, opened 2024-10-02
feat: Add `version` command to all Go binaries

alfrunes:
The `version` command will display the app version (linked at build time) as well as runtime version and commit SHA1.

Changelog: Title
Ticket: None

Sample output (created a local v4.0.0-alpha0 tag for demo purposes).
```bash
deviceauth version
```
> ```
> Version:	v4.0.0-alpha0
> Go version:	go1.23.1
> Git commit:	0701064f0ee1784feea9178fc11d08c033805cca
> Revision time:	2024-10-02T10:58:16Z
> Os/Arch:	linux/amd64
> ```

```bash
deviceauth --version
```
> ```
> deviceauth version v4.0.0-alpha0
> ```

### DIFF
--- a/backend/pkg/version/version.go
+++ b/backend/pkg/version/version.go
@@ -1,0 +1,60 @@
+package version
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"runtime"
+	"runtime/debug"
+	"text/tabwriter"
+)
+
+var (
+	version string = "v0.0.0-dev"
+)
+
+type Version struct {
+	Version      string
+	GoVersion    string
+	GitCommit    string `json:",omitempty"`
+	RevisionTime string `json:",omitempty"`
+	OS           string
+	Arch         string
+}
+
+func Get() Version {
+	v := Version{
+		Version:   version,
+		GoVersion: runtime.Version(),
+		OS:        runtime.GOOS,
+		Arch:      runtime.GOARCH,
+	}
+	if bldInfo, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range bldInfo.Settings {
+			switch setting.Key {
+			case "vcs.revision":
+				v.GitCommit = setting.Value
+			case "vcs.time":
+				v.RevisionTime = setting.Value
+			}
+		}
+	}
+	return v
+}
+
+func (v Version) String() string {
+	var buf bytes.Buffer
+	w := tabwriter.NewWriter(&buf, 0, 8, 0, '\t', 0)
+	fmt.Fprintf(w, "Version:\t%s\n", version)
+	fmt.Fprintf(w, "Go version:\t%s\n", v.GoVersion)
+	fmt.Fprintf(w, "Git commit:\t%s\n", v.GitCommit)
+	fmt.Fprintf(w, "Revision time:\t%s\n", v.RevisionTime)
+	fmt.Fprintf(w, "Os/Arch:\t%s/%s\n", runtime.GOOS, runtime.GOARCH)
+	w.Flush()
+	return buf.String()
+}
+
+func (v Version) MarshalJSON() ([]byte, error) {
+	type version Version
+	return json.Marshal((version)(v))
+}
--- a/backend/services/Makefile.common
+++ b/backend/services/Makefile.common
@@ -79,12 +79,12 @@ docker-publish:
 endif
 
 $(binfile): $(GOFILES)
-	# TODO: Add -ldflags "-X ...$(Version)"
 	env CGO_ENABLED=$(CGO_ENABLED) \
 		GOOS=$(GOOS) \
 		GOARCH=$(GOARCH) \
 		go build -o $(binfile) \
 			-ldflags '$(LDFLAGS)' \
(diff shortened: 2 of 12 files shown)


merlin-northern (approved):
this is _exactly_ what I wanted, but there was a decision to remove all the versions from all the services. I fought for something like that, but lost. this is a must have for so many support requests I was having, you have no idea. I will give it another read-through, but this is great. thank you!

backend/pkg/version/version.go
```diff
@@ -0,0 +1,60 @@
...
+		OS:        runtime.GOOS,
+		Arch:      runtime.GOARCH,
+	}
+	if bldInfo, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range bldInfo.Settings {
+			switch setting.Key {
+			case "vcs.revision":
+				v.GitCommit = setting.Value
+			case "vcs.time":
+				v.RevisionTime = setting.Value
```
merlin-northern, line 38: great use of the vcs which is populated by default since some time now. I just love this patch.

backend/pkg/version/version.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+		}
+	}
+	return v
+}
+
+func (v Version) String() string {
+	var buf bytes.Buffer
+	w := tabwriter.NewWriter(&buf, 0, 8, 0, '\t', 0)
+	fmt.Fprintf(w, "Version:\t%s\n", version)
```
merlin-northern, line 48:
```suggestion
	fmt.Fprintf(w, "Version:%8s\n", version)
```
`8` is just an example -- but for cool alignment in column you can use it here.

backend/services/Makefile.common
```diff
@@ -78,12 +78,12 @@ docker-publish:
 endif
 
 $(binfile): $(GOFILES)
-	# TODO: Add -ldflags "-X ...$(Version)"
 	env CGO_ENABLED=$(CGO_ENABLED) \
 		GOOS=$(GOOS) \
 		GOARCH=$(GOARCH) \
 		go build -o $(binfile) \
 			-ldflags '$(LDFLAGS)' \
+			-ldflags '-X github.com/mendersoftware/mender-server/pkg/version.version=$(VERSION)' \
```
merlin-northern, line 87: just to double check, VERSION comes from where here?

backend/services/deployments/main.go
```diff
@@ -129,6 +135,28 @@ func doMain(args []string) {
 			},
 			Action: cmdStorageDaemon,
 		},
+		{
+			Name:  "version",
+			Usage: "Show version of the program",
```
merlin-northern:
```suggestion
			Usage: "Show version information",
```
maybe?

backend/services/workflows/main.go
merlin-northern, line 1: shall we distinguish here and print -worker and -server somewhere?
super bonus points: connect to nats and get the runtime version and print it as well! mongo would be nice too somewhere. as in:
"
running with mongo driver vxx mongo db vxx
running with nats driver vxx server vxx
"

backend/services/create-artifact-worker/cmd/root.go
merlin-northern, line 1: are these all the places we want the version? generate delta worker?

merlin-northern (approved):
after the second read-through I still love it. left some minor comments.

backend/pkg/version/version.go
alfrunes: The `w` is a `tabwriter.Writer` which formats the information into a (left-aligned) table. The tab `\t` character is used as a separator, but may actually result in two tabs if the column is too wide.

backend/services/Makefile.common
alfrunes: It's almost at the top of the file: https://github.com/mendersoftware/mender-server/blob/0701064f0ee1784feea9178fc11d08c033805cca/backend/services/Makefile.common#L14

backend/services/workflows/main.go
alfrunes: > shall we distinguish here and print -worker and -server somewhere?

I don't think so. They are the same binary - so the same version. We're also decommissioning the `workflows-worker` image, only keeping `workflows` (server) using a different command for the worker container.

> super bonus points: connect to nats and get the runtime version and print it as well! mongo would be nice too somewhere. as in:
> ```
> running with mongo driver vxx mongo db vxx
> running with nats driver vxx server vxx
> ```

I was considering it, but that would require a lot more work as different services have different mechanisms/dependencies. Maybe we can revisit this later?
At least by knowing the version of the binary, we can use git to look up all dependencies (/driver) versions.

backend/services/create-artifact-worker/cmd/root.go
alfrunes: The enterprise services will be covered after merging this. Note that the workflows binary pulled in to the docker image will also implicitly be versioned.

backend/services/workflows/main.go
merlin-northern: makes sense, lets revisit later.

merlin-northern (approved):

PR #74, opened 2024-10-03
Add enterprise composition and fix frontend acceptance tests

alfrunes:


mzedel (changes requested):
no. Please move the enterprise compose files here as well instead - so that users can evaluate the enterprise solution in a configuration we defined. As a user this change makes the product even less accessible.

mzedel (changes requested):
The last commit is misleading as it implies the enterprise pipeline for the gui image would be any different from what is run here - this should never happen.

mzedel (approved):
Thank you very much for the changes, the description and the e2e tests adjustments!! it should now be a lot easier to get started with the development 🤞

PR #75, opened 2024-10-03
chore: Use the correct license (Apache 2.0)

alfrunes:


tranchitella (approved):

vpodzime (changes requested):
This is not a simple fix of the license file. This is a major change.

vpodzime (approved):
Oh, I got confused and thought this was a different thing than it is. Let's just please double-check and make sure that all the repos the code had been moved in here from are Apache 2.0

alfrunes:
> Oh, I got confused and thought this was a different thing than it is. Let's just please double-check and make sure that all the repos the code had been moved in here from are Apache 2.0

I can confirm, they were all Apache 2.0. The initial license was added by mistake.

PR #80, opened 2024-10-04
Update backend tests dependencies and ignore bravado updates in backend tests dependencies

kjaskiewiczz:


merlin-northern (approved):
you are my hero with all those deps updates you do

tranchitella (approved):

PR #91, opened 2024-10-08
fix: remove Audit log tab when tenant does not have audit log capability

mineralsfree:
Ticket: MEN-7600
Changelog: None

### DIFF
--- a/frontend/src/js/components/leftnav.js
+++ b/frontend/src/js/components/leftnav.js
@@ -21,7 +21,7 @@ import { makeStyles } from 'tss-react/mui';
 
 import storeActions from '@northern.tech/store/actions';
 import { TIMEOUTS, canAccess } from '@northern.tech/store/constants';
-import { getFeatures, getUserCapabilities, getVersionInformation } from '@northern.tech/store/selectors';
+import { getFeatures, getTenantCapabilities, getUserCapabilities, getVersionInformation } from '@northern.tech/store/selectors';
 import copy from 'copy-to-clipboard';
 
 import DocsLink from './common/docslink';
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/leftnav.js
```diff
@@ -130,11 +134,12 @@ export const LeftNav = () => {
   const { classes } = useStyles();
 
   const userCapabilities = useSelector(getUserCapabilities);
+  const tenantCapabilities = useSelector(getTenantCapabilities);
   return (
     <div className={`leftFixed leftNav ${classes.list}`}>
       <List style={{ padding: 0 }}>
         {listItems.reduce((accu, item, index) => {
-          if (!item.canAccess({ userCapabilities })) {
+          if (!item.canAccess({ userCapabilities }, { tenantCapabilities })) {
```
mzedel:
```suggestion
          if (!item.canAccess({ userCapabilities, tenantCapabilities })) {
```
Passing just a single options object with named attributes... just to prevent having a long list of options/ working around the absence of named parameters in JS

frontend/src/js/components/leftnav.js
```diff
@@ -33,7 +33,11 @@ const listItems = [
   { route: '/devices', text: 'Devices', canAccess: ({ userCapabilities: { canReadDevices } }) => canReadDevices },
   { route: '/releases', text: 'Releases', canAccess: ({ userCapabilities: { canReadReleases, canUploadReleases } }) => canReadReleases || canUploadReleases },
   { route: '/deployments', text: 'Deployments', canAccess: ({ userCapabilities: { canDeploy, canReadDeployments } }) => canReadDeployments || canDeploy },
-  { route: '/auditlog', text: 'Audit log', canAccess: ({ userCapabilities: { canAuditlog } }) => canAuditlog }
+  {
+    route: '/auditlog',
+    text: 'Audit log',
+    canAccess: ({ userCapabilities: { canAuditlog } }, { tenantCapabilities: { hasAuditlog } }) => hasAuditlog && canAuditlog
```
mzedel:
```suggestion
    canAccess: ({ userCapabilities: { canAuditlog }, tenantCapabilities: { hasAuditlog } }) => hasAuditlog && canAuditlog
```

mzedel (changes requested):
I slightly wonder if a user with auditlog access, but a role that shouldn't allow it will also see the entry (they shouldn't)...

mineralsfree:
> I slightly wonder if a user with auditlog access, but a role that shouldn't allow it will also see the entry (they shouldn't)...

Isn't `canAuditlog` be `false` in that case?

mzedel:
> Isn't `canAuditlog` be `false` in that case?

Exactly, but it should be the case already (in this test).

mzedel (approved):
having looked at the capabilities code, this should work fine and seems to more be a case of "how could this ever work" 🙈

mzedel:
Argh, I'm sorry for wasting your time @mineralsfree we made this change specifically in [MEN-5414](https://github.com/mendersoftware/gui/pull/3875) and I completely forgot about it 😞.

[MEN-5414]: https://northerntech.atlassian.net/browse/MEN-5414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #94, opened 2024-10-08
feat: new endpoint for listing deployments

kjaskiewiczz:


kjaskiewiczz:
@mzedel I added you as a reviewer, but my thinking was that you can/will skip the implementation part, but you may still be interested in the API docs

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -309,6 +312,18 @@ var (
 			Name:       &IndexArtifactProvidesName,
 		},
 	}
+
+	// 1.2.17
+	IndexDeploymentName = mongo.IndexModel{
+		Keys: bson.D{
+			{Key: StorageKeyDeploymentName,
+				Value: 1},
```
alfrunes:
```suggestion
			{Key: StorageKeyDeploymentName, Value: 1},
			{Key: "created", Value: 1},
```

alfrunes: We also need new indexes with secondary index on `created`, possibly dropping (replacing) previous ones, to avoid collection/index scans and in-memory sorts.

mzedel (approved):
Looks good to me, the only concern I have is users relying on the search/ partial match behaviour - but OTOH I think it's fair to break this through the version 👌

backend/services/deployments/app/app.go
```diff
@@ -1852,6 +1854,21 @@ func (d *Deployments) LookupDeployment(ctx context.Context,
...
+func (d *Deployments) LookupDeploymentV2(ctx context.Context,
+	query model.Query) ([]*model.Deployment, error) {
+	list, err := d.db.FindDeployments(ctx, query)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "searching for deployments")
+	}
+
+	if list == nil {
+		return make([]*model.Deployment, 0), nil
```
tranchitella: I would write it this way:

```suggestion
		list = make([]*model.Deployment, 0)
```

backend/services/deployments/app/app_test.go
```diff
@@ -1749,3 +1749,229 @@ func TestReindexDeployment(t *testing.T) {
...
+		deploymentsQuery model.Query
+		deployments      []*model.Deployment
+		deploymentsCount int64
+		deploymentsErr   error
+
+		res      []model.DeviceDeploymentListItem
+		resCount int
+		resErr   error
+	}{
+		"ok": {
```
tranchitella, line 1765: No test case for "not deployments found"

backend/services/deployments/docs/management_api_v2.yml
```diff
@@ -48,6 +48,102 @@ responses:
...
 
 paths:
+  /deployments:
+    get:
+      operationId: List Deployments
+      tags:
+        - Management API
+      security:
+        - ManagementJWT: []
+      summary: Find all deployments
```
tranchitella:
```suggestion
      summary: List all the deployments matching the specified filter parameters
```

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -2661,6 +2676,99 @@ func (db *DataStoreMongo) Find(ctx context.Context,
...
+	query := bson.M{}
+	if len(andq) != 0 {
+		// use search criteria if any
+		query = bson.M{
+			"$and": andq,
+		}
+	}
+
+	if match.CreatedAfter != nil && match.CreatedBefore != nil {
+		query["created"] = bson.M{
```
tranchitella: "created" is specified many times, maybe it should be a const somewhere

tranchitella (approved):

backend/services/deployments/store/mongo/datastore_mongo.go
kjaskiewiczz: I updated the one I'm creating with additional `created` key, but I'm not sure about other ones. Looks like we only have one index with `created` field only. Should I add index with all the fields we are using in the query?

PR #98, opened 2024-10-11
fix: fixed an issue that prevented enterprise demo tenant creation

mzedel:
FYI @mineralsfree

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ To initialize the admin user, use the following snippet:
 MENDER_NAME=Admin
 MENDER_USERNAME=admin@mender.local
 MENDER_PASSWORD=password123
-docker compose exec tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME"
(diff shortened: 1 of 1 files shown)


alfrunes (approved):

README.md
```diff
@@ -48,7 +48,7 @@ To initialize the admin user, use the following snippet:
 MENDER_NAME=Admin
 MENDER_USERNAME=admin@mender.local
 MENDER_PASSWORD=password123
-docker compose exec tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME"
+docker compose exec tenantadm tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME"
```
alfrunes, line 51: I would even consider using `run` instead of `exec` so that we have the logs in case it fails.
```suggestion
docker compose run --rm tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME"
```

alfrunes (approved):

PR #110, opened 2024-10-17
Docker compose environment improvements

alfrunes:
* Fixed mount path for `mongo` volume
* Added self-signed certificate covering hostnames `mender.local`, `localhost` and `docker.mender.io` for backward compatibility.
* Added client to the composition
  * Updated README.md

Ticket: QA-729


kjaskiewiczz (approved):
tested, works like a charm!

mzedel (commented):
In the spirit of being backward compatible: could we default to docker.mender.io? I can see the motivation of moving towards a `.local` address, but there are many other areas pointing to docker.mender.io already (docs, api docs, ui parts, hub entries, ...) - all of these will need extra clarifications/ adjustments/ time spent.

tranchitella (approved):

oldgiova:
Just a note: the `.local` TLD is [reserved for mDNS](https://www.rfc-editor.org/rfc/rfc6762.html) and it's usage it's preferibly to avoid. Better to use `.lan`.
That being said, I used the `.local` for years without any problem, so I'm not blocking this PR ;)

alfrunes:
For the sake of not creating a stir in the existing CI infra, I imported the old certificate and updated the hostnames to `docker.mender.io`. Let's review the domain name for the compose environment once the certificate renewal approaches.

PR #111, opened 2024-10-17
fix(deviceconfig): enable multiplatform build

oldgiova:
The required TARGETOS and TARGETARCH variables were missing from the Dockerfile.

Ticket: QA-673

### DIFF
--- a/backend/services/deviceconfig/Dockerfile
+++ b/backend/services/deviceconfig/Dockerfile
@@ -1,4 +1,6 @@
 FROM --platform=$BUILDPLATFORM golang:1.23.1 AS builder
+ARG TARGETOS
+ARG TARGETARCH
 ARG LDFLAGS="-s -w"
 ARG BUILDFLAGS="-trimpath"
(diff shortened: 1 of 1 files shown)


alfrunes (approved):

alfrunes:
Hmm... Why is CI trying to run all those jobs that are only supposed to run on protected references? :thinking:

oldgiova:
> Hmm... Why is CI trying to run all those jobs that are only supposed to run on protected references? 🤔

Because yesterday I protected the branch `pr_111` to build the `arm64` image and test it (then I unprotected it right before the publishing). By default it builds `amd64` only.

alfrunes:
> > Hmm... Why is CI trying to run all those jobs that are only supposed to run on protected references? 🤔
> 
> Because yesterday I protected the branch `pr_111` to build the `arm64` image and test it (then I unprotected it right before the publishing). By default it builds `amd64` only.

Aha, thanks for the explanation :)

PR #113, opened 2024-10-18
MEN-7568, MEN-7566 - feat: added service provider tenant UI and routes, SP header

mineralsfree:
Ticket: MEN-7568, MEN-7566
Changelog: None

### DIFF
--- a/frontend/src/js/components/header/header.js
+++ b/frontend/src/js/components/header/header.js
@@ -46,6 +46,7 @@ import {
   getDeviceLimit,
   getFeatures,
   getIsEnterprise,
+  getIsServiceProvider,
   getOrganization,
   getShowHelptips,
   getUserRoles,
@@ -54,6 +55,7 @@ import {
 import { useAppInit } from '@northern.tech/store/storehooks';
 import {
   getAllDeviceCounts,
+  getUserOrganization,
   initializeSelf,
   logoutUser,
   setAllTooltipsReadState,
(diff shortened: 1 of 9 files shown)


frontend/src/js/components/header/header.js
```diff
@@ -142,6 +145,17 @@ const useStyles = makeStyles()(theme => ({
...
+    padding: theme.spacing(1),
+    color: theme.palette.grey[800],
+    background: theme.palette.tooltip.tierTipBackground,
+    borderRadius: theme.spacing(3),
+    marginRight: theme.spacing(2),
+    fontSize: '11px'
+  },
+  spTenantBadgeContainer: {
+    borderRight: `1px solid ${theme.palette.grey[300]}`
   }
```
mzedel: there is the `.header-section` class that handles this for the existing widgets in the header - I guess that would be a good reason to keep the header related content in a less file in the header folder to make it more apparent there is some related stuff to be found

frontend/src/js/components/leftnav.js
```diff
@@ -35,6 +35,10 @@ const listItems = [
   { route: '/deployments', text: 'Deployments', canAccess: ({ userCapabilities: { canDeploy, canReadDeployments } }) => canReadDeployments || canDeploy },
   { route: '/auditlog', text: 'Audit log', canAccess: ({ userCapabilities: { canAuditlog } }) => canAuditlog }
 ];
+const spTenantItems = [
+  { route: '/tenants', text: 'Tenants', canAccess },
+  { route: '/auditlog', text: 'Audit log', canAccess }
+];
```
mzedel: these should go into the regular `listItems` and have their access checks adjusted

mzedel (changes requested):

mzedel (approved):
Thank you!

PR #118, opened 2024-10-21
deprectate(deployments): deprecate v1 endpoint for listing deployments

kjaskiewiczz:
We deprecated GET v1 /deployments/deployments endpoint because of an issue with "search" query parameter behavior. As a replacement we introduce v2 version of the endpoint, where we replaced "search" parameter with "id" and "name" parameters.

### DIFF
--- a/backend/services/deployments/docs/management_api.yml
+++ b/backend/services/deployments/docs/management_api.yml
@@ -54,6 +54,7 @@ responses:
 paths:
   /deployments:
     get:
+      deprecated: true
       operationId: List Deployments
       tags:
         - Management API
@@ -64,6 +65,10 @@ paths:
         Returns a filtered collection of deployments in the system,
         including active and historical. If both 'status' and 'query' are
         not specified, all devices are listed.
+
+        DEPRECATED: we deprecated the endpoint due to an issue with the "search" query
+          behavior. Please use the v2 /deployments/deployments endpoint instead.
+          In the new endpoint, we replaced search parameter with the "id" and "name" parameters.
       parameters:
         - name: status
           in: query


kjaskiewiczz:
Looking for an advise from reviewers - we don't support `deprecate` commit type, so I need to use one of
(fix|feat|build|chore|ci|docs|perf|refac|revert|style|test)
any preferences?

olehermanse:
> Looking for an advise from reviewers - we don't support deprecate commit type, so I need to use one of
(fix|feat|build|chore|ci|docs|perf|refac|revert|style|test)
any preferences?

@kjaskiewiczz probably irrelevant, but there is a typo in commit message and PR title; _deprectate_

merlin-northern:
> Looking for an advise from reviewers - we don't support `deprecate` commit type, so I need to use one of (fix|feat|build|chore|ci|docs|perf|refac|revert|style|test) any preferences?

maybe fix? as in the slightly other meaning of the fix word.

backend/services/deployments/docs/management_api.yml
```diff
@@ -64,6 +65,10 @@ paths:
         Returns a filtered collection of deployments in the system,
         including active and historical. If both 'status' and 'query' are
         not specified, all devices are listed.
+
+        DEPRECATED: we deprecated the endpoint due to an issue with "search" query
```
merlin-northern:
```suggestion
        DEPRECATED: we deprecated the endpoint due to an issue with the "search" query
```

backend/services/deployments/docs/management_api.yml
```diff
@@ -64,6 +65,10 @@ paths:
...
+          behavior. Please use the v2 /deployments/deployments end-point instead.
+          In the new endpoint, we replaced search parameter with "id" and "name" parameters.
```
merlin-northern: either endpoint in both places or end-point. histogram of usage:
```bash
#grep -rniHE end.*point backend/services/*/docs/*yml | grep -i [^a-zA-Z0-9/_]endpoint |wc -l
29
#grep -rniHE end.*point backend/services/*/docs/*yml | grep -i [^a-zA-Z0-9/_]end-point |wc -l
12
```

merlin-northern: also: maybe "with the id and name parameters" would sound better?

backend/services/deployments/docs/management_api.yml
kjaskiewiczz: done

backend/services/deployments/docs/management_api.yml
kjaskiewiczz: done

merlin-northern (approved):
very lovely.

PR #124, opened 2024-10-23
Fix/improve SeaweedFS docker compose setup

alfrunes:
The command passed to the `s3fs` service messed up the entrypoint as all the argument was passed as a single string. The entrypoint adds default argument using a weak test on the first argument passed, since it didn't match the expectation the persistence configuration was messed up and would break on restarts (only the file index would be persisted without the actual data).
Since we are using path-style URLs for accessing the artifacts bucket, we can make the rule independent of the hostname.


alfrunes:
Superseded by #136

PR #128, opened 2024-10-23
Feat/frontend license check

mzedel:


frontend/pipeline.yml
```diff
@@ -30,6 +30,21 @@ test:frontend:lint:
   tags:
     - hetzner-amd-beefy
 
+test:frontend:license-headers:
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-2.0.2
+  stage: test
+  rules:
```
oldgiova, line 36: The runner tag is missing, I fear

oldgiova (approved):
Optional comment, otherwise ok

frontend/pipeline.yml
mzedel: That gets inherited from the defaults in the root gitlab - so it runs on there already 👌 :
Screenshot 2024-10-23 at 15 20 29

PR #133, opened 2024-10-23
Add context in case of e2e auditlogs test failures

mzedel:
this particular failure (failing to parse the received terminal recording because of faulty non-JSON text) seems to occur more in the enterprise repo, but not reliably and I can't seem to reproduce it locally or anywhere else

### DIFF
--- a/frontend/tests/e2e_tests/integration/06-auditlogs.spec.ts
+++ b/frontend/tests/e2e_tests/integration/06-auditlogs.spec.ts
@@ -23,6 +23,7 @@ import { selectors, storagePath, timeouts } from '../utils/constants.ts';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+const search = `const transfer = '[{"content":`;
 const checkDownloadedReplayForSecret = async (path, secret) => {
   const fileStream = fs.createReadStream(path);
   const lines = readline.createInterface({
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e_tests/integration/06-auditlogs.spec.ts
```diff
@@ -23,23 +23,30 @@ import { selectors, storagePath, timeouts } from '../utils/constants.ts';
...
       const encodedText = line.substring(line.indexOf(search) + search.length, line.lastIndexOf(`}]';`));
-      const content = JSON.parse(Buffer.from(encodedText, 'base64').toString());
+      const jsonContent = Buffer.from(encodedText, 'base64').toString();
+      let content;
+      try {
+        content = JSON.parse(jsonContent);
+      } catch {
+        console.log('checkDownloadedReplayForSecret', jsonContent);
+      }
       const decodedContent = String.fromCharCode(...content.data);
```
mineralsfree, line 43: Wouldn't content be undefined in case of exception here which will cause `Cannot read properties of undefined (reading 'data')`

mzedel: Yes, this is not supposed to really fix whatever is happening there, but rather help track down what is received here in the first place for failures like [here](https://gitlab.com/Northern.tech/Mender/mender-server-enterprise/-/jobs/8060571820#L2327)
If the parsing fails, the test should fail, if the parsing passes, content should be an object with `data`.

mineralsfree: than it's ok, I believe that following exception wouldn't create much noise in the logs, good improvement for e2e debuging indeed :+1:

mineralsfree (approved):

PR #142, opened 2024-10-25
feat(deployments): add filter field to deployment object

kjaskiewiczz:
The filter field contains information about devices targeted by the deployment.
Ticket: MEN-7416

### DIFF
--- a/backend/services/deployments/app/app.go
+++ b/backend/services/deployments/app/app.go
@@ -57,6 +57,7 @@ const (
 	InventoryIdentityScope           = "identity"
 	InventoryGroupAttributeName      = "group"
 	InventoryStatusAttributeName     = "status"
+	InventoryIdAttributeName         = "id"
 	InventoryStatusAccepted          = "accepted"
 
 	fileSuffixTmp = ".tmp"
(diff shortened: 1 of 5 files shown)


backend/services/deployments/app/app.go
```diff
@@ -1245,6 +1247,50 @@ func (d *Deployments) getDeploymentGroups(
...
+				},
+			},
+		}
+	} else if len(constructor.Devices) > 0 {
+		filter = &model.Filter{
+			Terms: []model.FilterPredicate{
+				{
+					Scope:     InventoryIdentityScope,
+					Attribute: InventoryIdAttributeName,
+					Type:      "$in",
```
alfrunes, line 1284: This is not a valid filtering operator according to inventory, should we add it? https://github.com/mendersoftware/mender-server/blob/f9ce3d325be98697ab7cee464cbbcba6d5683b1c/backend/services/inventory/model/filters.go#L21-L24

kjaskiewiczz: looks like it's a mess
it's not here - https://github.com/mendersoftware/mender-server/blob/f9ce3d325be98697ab7cee464cbbcba6d5683b1c/backend/services/inventory/docs/management_api_v2.yml#L278 (btw, here $nin is also missing)
it's here - https://github.com/mendersoftware/mender-server/blob/f9ce3d325be98697ab7cee464cbbcba6d5683b1c/backend/services/inventory/docs/internal_api_v2.yml#L219 together with almost every possible operator
now I'm wondering how does it work...

kjaskiewiczz: @alfrunes I created a follow up ticket - https://northerntech.atlassian.net/browse/MEN-7667.
With this PR we don't execute the filter. Filter will be executed later by the UI (there will be separate UI task to start using this functionality), so it will be blocked by the inventory one.

tranchitella (approved):

alfrunes (approved):

PR #145, opened 2024-10-29
fix: fixed an issue that prevented onboarding tips from showing

mzedel:
With the RTK approach destructuring the entire slice state will not produce a proper object and break the change tracking of the js proxy and instead the change will just fall back to the previous value... but if instead of relying on the proxy tracking the entire state is returned as a new object, this will be picked up as a new state altogether, so all the selectors get updated as well 👌.

### DIFF
--- a/frontend/src/js/store/onboardingSlice/index.ts
+++ b/frontend/src/js/store/onboardingSlice/index.ts
@@ -30,7 +30,7 @@ export const onboardingSlice = createSlice({
   initialState,
   reducers: {
     setOnboardingState: (state, action) => {
-      state = { ...state, ...action.payload };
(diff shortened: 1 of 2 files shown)


alfrunes (approved):
The tool tips reappeared! However, they were still hidden on the main dashboard, they only resurfaced when I went to the devices tab.

alfrunes:
> The tool tips reappeared! However, they were still hidden on the main dashboard, they only resurfaced when I went to the devices tab.

Sorry, disregard. The tooltip wasn't showing because it skipped ahead since I had a device. After starting fresh it is certainly there. We only need to modify the snippets involving the `demo` script (and possibly also uploading the demo artirfact).

PR #150, opened 2024-10-29
feat(deviceconfig): support for the hidden Configure Add-on fields

merlin-northern:
The set configuration for deviceconfing is extended to accept fields in the form of an object with `value` and a `hidden` attrbutes

The v2 get configuration for deviceconfig endpoint returns the new attributes without values, if `hidden` is set to true.

Changelog: It is possible to define a hidden configuration attribute, which will lead to values that will never be openly presented to the user, and can be seen as hidden with queries to the v2 endpoint.
Ticket: MEN-7614


merlin-northern:
@alfrunes could I ask for an early review, before I add the tests? the `attributes2Map` is perhaps not the best piece I have created, but it works.

backend/services/deviceconfig/model/attribute.go
```diff
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
 type Attribute struct {
```
alfrunes, line 25: May I suggest we put the hidden flag here (inside `Attribute`)?
That way we don't have to do the bson dance inside `attributes2Map`.
Instead you may have to make some slight modifications to the `Attributes.UnmarshalJSON` (hint: try to unmarshal `map[string]Attribute`).

backend/services/deviceconfig/api/http/management.go
```diff
@@ -99,6 +100,16 @@ func (api *ManagementAPI) GetConfiguration(c *gin.Context) {
...
+	if strings.HasPrefix(c.Request.RequestURI, URIManagement) {
+		// in the v1 endpoint we do not return other attributes that the flat json ones
+		var flatAttributes model.Attributes
+		for _, a := range device.ConfiguredAttributes {
+			if _, isString := a.Value.(string); isString {
+				flatAttributes = append(flatAttributes, a)
+			}
+		}
+		device.ConfiguredAttributes = flatAttributes
+	}
```
alfrunes, line 112: This would break the PUT semantics - hidden attributes would be removed if you try to use the v1 endpoint in this case. I'm not sure if it's a good idea to introduce a new endpoint after all.

backend/services/deviceconfig/model/attribute.go
```diff
@@ -74,7 +76,38 @@ func (a *Attributes) UnmarshalJSON(b []byte) error {
 func attributes2Map(attributes []Attribute) map[string]interface{} {
```
alfrunes, line 76: Side note (not related to your changes): We should sort the attributes inside `map2Attributes` to prevent randomizing the order of attributes which may lead to unnecessary writes to the DB.

backend/services/deviceconfig/model/validation.go
```diff
@@ -30,7 +30,10 @@ var (
 		switch value.(type) {
 		case string:
 			return nil
-
+		case map[string]interface{}:
+			// MEN-7614: added support for
+			// PUT {"key1":"value1","key8":{"value":"value8","hidden":true}}
```
alfrunes, line 35: Missing schema validation for this case (this case is treated as free form. See my suggestion above - in that case we wouldn't need another case in this switch.

alfrunes (commented):
I have some doubts about introducing the new API endpoint.

There's also missing some crucial features of this ticket:
* Hidden attribute values are returned from the GET API - they should be hidden.
* There's logic missing at the database level for preserving hidden values on PUT.
* The devices API should not be affected by the new schema in the management API.

backend/services/deviceconfig/api/http/management.go
merlin-northern: that was the idea: to not return hidden attributes (this is GET). if you use v1 PUT, or any PUT, all the attributes are replaced always. this block only ensure that the v1 works as it used to. hidden attributes are hidden. with v2 they are returned as ticket said: 
```json
{
  "id": "8eb63348-f63d-4b9d-802d-85ea1d4fbc5b",
  "configured": {
    "key12": {
      "hidden": true
    },
    "key2": "value2",
    "key4": "value4"
  },
  "reported": {
    "timezone": "UTC"
  },
  "updated_ts": "2024-10-30T09:09:12.122Z",
  "reported_ts": "2024-10-28T18:19:21.827Z"
}
```

backend/services/deviceconfig/model/attribute.go
merlin-northern: okey.

merlin-northern:
> I have some doubts about introducing the new API endpoint.
> 
> There's also missing some crucial features of this ticket:
> 
> * Hidden attribute values are returned from the GET API - they should be hidden.

 wait the ticket says: to return them as `{"hidden_config": {"hidden": true}}` and thats how they are returned.

> * There's logic missing at the database level for preserving hidden values on PUT.

the endpoint always worked like that: replacing all you supply. so you want to always preserve the hidden attributes? so each PUT will technically leave all the hidden attributes in the db as they are? so how do I replace them, or remove?

> * The devices API should not be affected by the new schema in the management API.

wait, I have not touched the device API, there maybe some hidden features introduced, but the API should be untouched.

PR #154, opened 2024-10-30
ci: Frontend acceptance job use local artifacts for protected refs

alfrunes:
The Frontend acceptance tests fails for commit tags because the reference is published after the test pass. Also, we want to test the build artifacts for protected branches. Therefore it will only use upstream main for PRs.


alfrunes:
See failing job: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/8230040255

The job fails because the release tag does not (yet) exist. Eventually we can run the pipeline because publishing the backend components does not depend on the frontend e2e tests (yet).

EDIT: It fails because it uses the raw tag (not slug).

alfrunes:
I also discovered that the published Docker tags were translated to `v4-0-0-rc-1` which is not intended. We should instead use the ref name and make sure that all protected references use a name that is compatible with container tags.

oldgiova (approved):

PR #156, opened 2024-10-31
Use utils from nt-gui

aleksandrychev:


mzedel (changes requested):
Thank you, I like it a lot! can you please also remove the helpers.js? just to be sure there's no leftover reference possible 🙃

aleksandrychev:
> Thank you, I like it a lot! can you please also remove the helpers.js? just to be sure there's no leftover reference possible 🙃

yes, sure. I missed this one. I thought that I was removed together with utils/*

mzedel (approved):
🚀 - also FYI @olehermanse

aleksandrychev:
I additionally tested UI manually, seems like everything is correct. so, going to merge it

PR #158, opened 2024-10-31
chore(compose): remove configs top-level element from evaluation composition

alfrunes:
The `configs` attribute requires a newer version of Docker Compose. Removing it to lower the entry barrier.
NOTE: S3 credentials cannot be changed using the project environment
      (`.env`).


alfrunes:
Decided to bump the compose version requirement instead (#172).

PR #182, opened 2024-11-06
chore(ci): install git before running `npm run disclaim`

tranchitella:
Changelog: None
Ticket: None


mzedel:
This won't work, as the nt-gui package isn't released yet - there's https://github.com/mendersoftware/mender-server/pull/181 to fix it.

PR #185, opened 2024-11-06
ci: Add trigger job for full integration pipeline

alfrunes:


lluiscampos (commented):
Could we have this in a template so that it can be included from other repos? Ideally I'd like to have it in `mendertesting`, but I am also okey with leaving it in `mender-server` as long as it is a separate file :smiley:

.gitlab-ci.yml
```diff
@@ -268,6 +268,28 @@ test:backend:integration:
...
 
+test:integration:
+  stage: test
+  needs:
+    - job: build:backend:docker
+      artifacts: false
+    - job: build:frontend:docker
+      artifacts: false
+  rules:
+    - if: $CI_COMMIT_REF_PROTECTED == "true"
```
lluiscampos, line 279: I think this rule correct: you don't need to have it from a protected branch in the caller pipeline (`mender-server` in this case).

alfrunes: The condition is for the dependencies only. It requires a protected branch (or modifications to both frontend and backend subdirectories) to build all images.

alfrunes:
> Could we have this in a template so that it can be included from other repos? Ideally I'd like to have it in `mendertesting`, but I am also okey with leaving it in `mender-server` as long as it is a separate file 😃

Hmm... There is a lot of mender-server specific configurations in the job. If I only split out the generic part, I think it would be (excluding Gitlab defaults):

```yaml
test:integration:
  variables:
    RUN_TESTS_FULL_INTEGRATION: "true"
  trigger:
    project: "Northern.tech/Mender/integration"
```

The rest is setting up integration to pull from the internal GitLab registry of this project and excluding enterprise tests (since this repository is only publishing the open source images).

lluiscampos:
> > Could we have this in a template so that it can be included from other repos? Ideally I'd like to have it in `mendertesting`, but I am also okey with leaving it in `mender-server` as long as it is a separate file 😃
> 
> Hmm... There is a lot of mender-server specific configurations in the job. If I only split out the generic part, I think it would be (excluding Gitlab defaults):
> 

I don't understand then why are we doing this, then.

The backend team decided that they didn't want such a trigger, we rejected the task, then I pointed out that we need the trigger from other places (mender client, addons, etc) and decided to reopen the task to make it generic somehow.

Now we are adding a trigger mender-serer -> integration (which I though you didn't want!) and still not solving the issue for the rest of the repos.

alfrunes:
> > > Could we have this in a template so that it can be included from other repos? Ideally I'd like to have it in `mendertesting`, but I am also okey with leaving it in `mender-server` as long as it is a separate file 😃
> > 
> > 
> > Hmm... There is a lot of mender-server specific configurations in the job. If I only split out the generic part, I think it would be (excluding Gitlab defaults):
> 
> I don't understand then why are we doing this, then.
> 
> The backend team decided that they didn't want such a trigger, we rejected the task, then I pointed out that we need the trigger from other places (mender client, addons, etc) and decided to reopen the task to make it generic somehow.
> 
> Now we are adding a trigger mender-serer -> integration (which I though you didn't want!) and still not solving the issue for the rest of the repos.

I'm not sure I understand. There's very little boilerplate to make the trigger for other repositories (which I hope this PR shows). Maybe I misunderstood (@kjaskiewiczz ?), but the ticket says:

> [...] but we also need a way to run full integration pipeline on mender-server PRs and this task is about making it possible.

So I thought that ticket was simply for creating a trigger from the mender-server repository.

EDIT/UPDATE: When it comes to other repositories, I imagine setting up a similar trigger as here. Create a trigger job with the respective image environment variable overwritten (leaving the rest untouched will default to main). The rules might be simpler as the mender-server needs to account for open source and enterprise tests as enterprise tests won't run with open source images and vice versa. On the other hand, we might need to expand the environment in integration repository to allow overwriting the image source for client/mender-gateway.

alfrunes:
Found another issue with the pytest filtering inside the integration repository: https://github.com/mendersoftware/integration/pull/2670

lluiscampos:
>but the ticket says:
> 
> > [...] but we also need a way to run full integration pipeline on mender-server PRs and this task is about making it possible.

We didn't update the ticket during/after the meeting. But I thought we were on the same page after the discussion on the meeting: it is not for me nor anyone out of the backend team to decide if `mender-server` should use a trigger or only nighties (I insist: I completely trust that you will do whatever best fits your dev flow!), but what we need is to have the trigger for the rest of the repos that use to have it via mender-qa pipeline.

> 
> So I thought that ticket was simply for creating a trigger from the mender-server repository.
> 
> EDIT/UPDATE: When it comes to other repositories, I imagine setting up a similar trigger as here. 

Yes, this PR shows how easy it is, but still doing it this way (with no template) means that we need to add it to many repos, and whenever there is a change (a variable rename, a variable addition, whatever) it will break or run with the incorrect versions. On top of my head, we need this at least `mender`, `mender-gateway`, `mender-cli`, `mender-artifact`, `mender-connect`.

I am sorry to be pedantic with this. The issue for the "non-server" release now is that the integration tests are completely out of the control of the release_tool (yey!) but then it requires a lot of human focus to test the right things (not "yey!") :smiley:

alfrunes:
Something like this https://github.com/mendersoftware/mender-gateway/pull/179 ?

lluiscampos:
> Something like this [mendersoftware/mender-gateway#179](https://github.com/mendersoftware/mender-gateway/pull/179) ?

Something like that, exactly! So do you want to have such triggers from all repos? And we'll we then maintain them separately?

(For `mender-gateway` specifically the story is way more complicated that that trigger, as the client image is built in yet a different place :sob: )

Also, what do you want to do on this PR? It looks good to me if the backend team has decided that the trigger is good to have from `mender-server` too.

alfrunes:
> > Something like this [mendersoftware/mender-gateway#179](https://github.com/mendersoftware/mender-gateway/pull/179) ?
> 
> Something like that, exactly! So do you want to have such triggers from all repos? And we'll we then maintain them separately?

I believe yes. The amount of shared ci yaml is pretty much the same as the include statement.

> (For `mender-gateway` specifically the story is way more complicated that that trigger, as the client image is built in yet a different place 😭 )

Maybe we should think about restructuring the way we're building/publishing the QEMU images? I suppose it would be a hot topic for the QA meeting :memo: :sauna_person:

> Also, what do you want to do on this PR? It looks good to me if the backend team has decided that the trigger is good to have from `mender-server` too.

For this PR, the goal was simply to have a way to manually trigger an integration pipeline from a protected branch (the ticket says PR, but we need the branch to be protected to build both frontend and backend). This will be handy for running integration tests on release tags, for instance.

lluiscampos (approved):

PR #188, opened 2024-11-06
chore: allow start case sentences in commits

oldgiova:
No need to be so strict about sentence case in commits subjects


mzedel (approved):

alfrunes (approved):

alfrunes:
Yay! My muscle memory appreciates this :pray:

PR #190, opened 2024-11-07
chore(deviceauth): moving tenantadm client to pkg

merlin-northern:
Changelog: None
Ticket: MEN-7715

### DIFF
--- a/backend/services/deviceauth/devauth/devauth.go
+++ b/backend/services/deviceauth/devauth/devauth.go
@@ -30,12 +30,12 @@ import (
 	"github.com/mendersoftware/mender-server/pkg/plan"
 	"github.com/mendersoftware/mender-server/pkg/ratelimits"
 	"github.com/mendersoftware/mender-server/pkg/requestid"
+	tenant2 "github.com/mendersoftware/mender-server/pkg/tenant"
 
 	"github.com/mendersoftware/mender-server/services/deviceauth/access"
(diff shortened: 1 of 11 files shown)


backend/services/deviceauth/devauth/devauth.go
```diff
@@ -30,12 +30,12 @@ import (
 	"github.com/mendersoftware/mender-server/pkg/plan"
 	"github.com/mendersoftware/mender-server/pkg/ratelimits"
 	"github.com/mendersoftware/mender-server/pkg/requestid"
+	tenant2 "github.com/mendersoftware/mender-server/pkg/tenant"
```
alfrunes, line 33: There's no name conflict for `tenant` as far as I can see, can we remove the `2` suffix or use a more descriptive import name like `tenantClient`?
```suggestion
	"github.com/mendersoftware/mender-server/pkg/tenant"
```

alfrunes (commented):
Nice initiative.
May I suggest a slightly different (opinionated) structure?
Do you think we can group the clients something like this?
`pkg/client/<service name>` so it becomes `pkg/client/tenantadm`.

mzedel:
@merlin-northern I think this now is superseded by the openapi client - right?

merlin-northern:
> @merlin-northern I think this now is superseded by the openapi client - right?

right! or: either way outdate heavily.

PR #191, opened 2024-11-08
MEN-7570 - SP RBAC + ensured linter runs on all TS files

mzedel:


frontend/src/js/components/settings/role-management/DeleteRoleDialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@mui/material';
+
+interface Props {
+  dismiss: () => void;
+  submit: () => void;
+  name: string;
+  open: boolean;
+}
+
+export const DeleteRoleDialog: FunctionComponent<Props> = ({ dismiss, open, submit, name }) => (
```
mineralsfree: Is it possible to use our `ConfirmModal` with some adjustments?

frontend/src/js/common-ui/forms/form.js
```diff
@@ -41,6 +41,12 @@ const getErrorMsg = (validateMethod, args) => {
         return `Must be at least ${args[0]} characters long`;
       }
       break;
+    case 'isAlphanumericLocator':
+      if (args[0] && validator.isWhitelisted(args[0], 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-')) {
```
mineralsfree: maybe 
```suggestion
      if (args[0] && validator.matches(args[0], /^[a-zA-Z0-9_-]+$/)) {
```
is a better way?

mineralsfree (dismissed):
few remarks, + several new components are missing prop types, otherwise is good! :+1:

frontend/src/js/components/settings/role-management/DeleteRoleDialog.tsx
mzedel: Good point! 🙊 if we don't tell anyone I think it should be fine 😄

mineralsfree (dismissed):

mineralsfree (dismissed):
Licenses file needs an update

mineralsfree (approved):

PR #193, opened 2024-11-11
chore(gui): added extra debug output to help track down auditlog failures

mzedel:
the output we've gotten so far is: `N̗_`  and I'm not sure we it's from 🙃

this should help tracking down these: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/8306707232


oldgiova (approved):

mineralsfree (approved):
:+1:

PR #194, opened 2024-11-13
MEN-7569 - feat: added tenant audit logs type support

mineralsfree:
Ticket: MEN-7569
Changelog: None


frontend/src/js/store/organizationSlice/constants.ts
```diff
@@ -60,6 +60,7 @@ export const AUDIT_LOGS_TYPES = [
   { title: 'Device', queryParameter: 'object_id', value: 'device' },
   { title: 'User', queryParameter: 'object_id', value: 'user' }
 ];
+export const SP_AUDIT_LOGS_TYPES = [...AUDIT_LOGS_TYPES, { title: 'Tenant', queryParameter: 'object_id', value: 'tenant' }];
```
mzedel: https://www.figma.com/design/xwdyJV8aKd5J3Rk0lBqzIS/Multi-tenant-access-UX?node-id=128-1501&node-type=instance&t=bx3dX27E0bjE9Y5l-0 shows only the `User` and `Tenant` in the list, this would also allow filtering for entries not relevant to the SP

mzedel (changes requested):
Please also add the basic event details rendering of the changes (I guess `eventdetails/devicedetails` or `eventdetails/userchange` could be good starting points)

frontend/src/js/components/auditlogs/__snapshots__/eventdetailsdrawer.test.js.snap
```diff
@@ -176,7 +176,11 @@ exports[`EventDetailsDrawer Component renders correctly 1`] = `
       <h3
         class="capitalized-start flexbox center-aligned"
       >
-        created user
+        <div
+          class="capitalized-start"
+        >
```
mzedel, line 181: The DrawerTitle should already capitalize the start - was it not?

frontend/src/js/components/auditlogs/auditlogslist.js
```diff
@@ -43,6 +43,12 @@ const UserChange = ({ item: { change = '-' } }) => {
     </div>
   );
 };
+const TenantChange = props => {
+  const {
+    item: { change = '-' }
+  } = props;
+  return <div>{change}</div>;
+};
```
mzedel: just as this is so close to the user change now (where the diff splitting seems to be no longer needed) - could they be unified to something like a `ChangeFallback` component?

mzedel (dismissed):
Fine as is, take the comments or leave them 😄

frontend/src/js/components/auditlogs/__snapshots__/eventdetailsdrawer.test.js.snap
mineralsfree: No it is not capitalizing, I will fix the DrawerTitle :)

frontend/src/js/common-ui/DrawerTitle.tsx
```diff
@@ -27,7 +27,7 @@ interface DrawerTitleProps {
 export const DrawerTitle: FunctionComponent<DrawerTitleProps> = ({ onClose, onLinkCopy, postTitle, preCloser, title }) => (
   <div className="flexbox margin-bottom-small space-between">
     <div className="flexbox center-aligned">
-      <h3 className="capitalized-start flexbox center-aligned">{title}</h3>
+      <h3 className="capitalized-start">{title}</h3>
```
mzedel: Unfortunately this won't work for device, deployment & release details 😬.
An "interesting" restriction to only allow this for `block` level elements... I can't think of a nice way to support both usages (`title` as string or component) so I'm leaning more towards adopting your previous approach in the use cases that don't already provide a "capitalized start title". What do you think?

mineralsfree: Works for me, on the other hand, having non-inline elements inside of h4 is not the best approach as well. IMO it's a minor issue and we can improve it along the way

mzedel (approved):

PR #195, opened 2024-11-13
MEN-7615 - device count retrieval + fixes to app init for SPs

mzedel:


mineralsfree (approved):
Looks good to me :+1:

PR #198, opened 2024-11-15
docs(deviceauth): migration to OpenAPI3

merlin-northern:
Changelog: None
Ticket: MEN-7714

### DIFF
--- a/backend/services/deviceauth/docs/devices_api.yml
+++ b/backend/services/deviceauth/docs/devices_api.yml
@@ -1,35 +1,18 @@
-swagger: '2.0'
+openapi: 3.0.0
 info:
-  version: '1'
-  title: Device authentication
+  version: '1.0'
+  title: Device Authentication
   description: |
     An API for device authentication handling. Intended for use by devices.
+servers:
+  - url: https://hosted.mender.io/api/devices/v1/authentication
 
-basePath: '/api/devices/v1/authentication'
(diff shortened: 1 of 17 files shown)


backend/services/deviceauth/docs/devices_api.yml
```diff
@@ -1,35 +1,18 @@
-swagger: '2.0'
+openapi: 3.0.0
```
tranchitella, line 1: How did you convert the specs? I suppose one of the (implicit) acceptance criteria was having an automated way to migrate from 2 to 3, so we can apply the same technique to the other repos.

merlin-northern: wooops I forgot about the research path. okey, this is done by hand and some scripting that is not even saved. I will do the research separately. sorry for that.

merlin-northern: no wait, I did some. but jira did not show me the whole ticket description! let me comment in the ticket.

merlin-northern: @tranchitella btw the ticket does not mention the research on the tools that can do the migration.

merlin-northern:
@kjaskiewiczz I need an early review on this one, there are lots of things still to refactor in those tests, but I had to draw a line somewhere.

kjaskiewiczz (approved):

PR #202, opened 2024-11-18
MEN-7733: Rate limits for devices APIs

alfrunes:
Added adaptive ratelimiting for devices APIs based on the max_devices
limit. Two new configuration options are exposed:

* `ratelimits.interval` - duration until the quota resets
* `ratelimits.quotas` - mapping plan names to quota weights. The final
    burst limit is computed as ratelimits.quotas[plan] * max_devices

The rate limiting counter is independent for each API path (independent
of method and version).

I also added caching for getting limits since the rate limiting depends on getting the device limit.

> [!TIP]
> To test the implementation you can use the following compose override
>
> ```yaml
> # mender-server/docker-compose.override.yml
> services:
>   redis:
>     image: redis:7.2
>     networks:
>       default:
>         aliases: [mender-redis]
>   deviceauth:
>     environment:
>       DEVICEAUTH_REDIS_CONNECTION_STRING: "redis://redis"
>       DEVICEAUTH_RATELIMITS_DEVICES_ENABLE: "true"
>       DEVICEAUTH_RATELIMITS_DEVICES_QUOTA_DEFAULT: "1.0"
>       DEVICEAUTH_RATELIMITS_DEVICES_INTERVAL: "1m"
> ```
> By manually setting the max devices for the "empty tenant" you can adjust the rate limits for open source:
> ```bash
> // By setting the limit to 5, a single device can perform at most 5 requests per minute.
> docker compose exec mongo mongosh --eval 'db.getSiblingDB("deviceauth").limits.insertOne({tenant_id: "", name: "max_devices", limit: 5})'
> ```

> [!NOTE]
> Open source does not have plans, so the RATELIMITS_QUOTAS configuration is ineffectual (but required). The default quota factor is set to 1.0.

### DIFF
--- a/backend/pkg/rate/limit_test.go
+++ b/backend/pkg/rate/limit_test.go
@@ -1,0 +1,47 @@
+package rate
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+func TestLimiter(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	limiter := NewLimiter(2, time.Hour)
+	resApproved, err := limiter.Reserve(ctx)
+	if err != nil {
+		t.Errorf("unexpected error reserving time slot #1: %s", err.Error())
+		t.FailNow()
+	}
+	_, err = limiter.Reserve(ctx)
+	if err != nil {
+		t.Errorf("unexpected error reserving time slot #2: %s", err.Error())
+		t.FailNow()
+	}
+	resDenied, err := limiter.Reserve(ctx)
+	if err != nil {
+		t.Errorf("unexpected error reserving time slot #3: %s", err.Error())
+		t.FailNow()
+	}
+	if !resApproved.OK() {
+		t.Error("expected first reservation to be available")
+	} else if resApproved.Delay() > 0 {
+		t.Error("an approved reservation should not have a delay")
+	} else if resApproved.Tokens() == 0 {
+		t.Error("expected more tokens to be available after first reservation")
+	}
+	if resDenied.OK() {
+		t.Error("reservation should not be available before 1h has passed")
+	} else {
+		if resDenied.Delay() == 0 {
+			t.Error("a denied reservation should have a non-zero delay")
+		}
+		if resDenied.Tokens() > 0 {
+			t.Errorf("a deneied reservation should not report free tokens, got: %d", resDenied.Tokens())
+		}
+	}
+}
--- a/backend/pkg/redis/ratelimit.go
+++ b/backend/pkg/redis/ratelimit.go
@@ -1,0 +1,129 @@
+package redis
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/redis/go-redis/v9"
+
+	"github.com/mendersoftware/mender-server/pkg/rate"
+)
+
+func NewFixedWindowRateLimiter(
+	client Client,
+	paramsFromContext RatelimitParamsFunc,
+) rate.Limiter {
+	return &fixedWindowRatelimiter{
+		client:     client,
+		paramsFunc: paramsFromContext,
+		nowFunc:    time.Now,
+	}
+}
+
+type RatelimitParams struct {
+	Burst     uint64
+	Interval  time.Duration
+	KeyPrefix string
+}
+
+type RatelimitParamsFunc func(context.Context) (*RatelimitParams, error)
+
+func FixedRatelimitParams(params RatelimitParams) RatelimitParamsFunc {
+	return func(ctx context.Context) (*RatelimitParams, error) {
+		return &params, nil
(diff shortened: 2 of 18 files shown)


backend/pkg/rate/limit_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	if !resApproved.OK() {
+		t.Error("expected first reservation to be available")
+	} else if resApproved.Delay() > 0 {
+		t.Error("an approved reservation should not have a delay")
+	} else if resApproved.Tokens() == 0 {
+		t.Error("expected more tokens to be available after first reservation")
+	}
+	if resDenied.OK() {
+		t.Error("reservation should not be available before 1s has passed")
```
kjaskiewiczz: 1s?

backend/pkg/redis/ratelimit.go
```diff
@@ -0,0 +1,125 @@
...
+	return fmt.Sprintf("%s:e:%d:c", prefix, epoch)
+}
+
+func (rl *fixedWindowRatelimiter) Reserve(ctx context.Context) (rate.Reservation, error) {
+	now := rl.nowFunc()
+	params, err := rl.paramsFunc(ctx)
+	if err != nil {
+		return nil, err
+	}
+	epoch := epoch(now, params.Interval)
```
kjaskiewiczz, line 84: In case there is no limit GetLimit will return 0,  RateLimitsFromContext will return -1 as limit, NewFixedWindowRateLimiter will return nil, nil and the code will crash here. Tested on my local setup.

backend/services/deviceauth/config/config.go
```diff
@@ -79,6 +81,13 @@ const (
 	// Has no effect if not running in multi-tenancy context.
 	SettingHaveAddons        = "have_addons"
 	SettingHaveAddonsDefault = false
+
+	// SettingRatelimits* configures adaptive rate limiting based on device limit.
```
kjaskiewiczz, line 85: these are not documented in the config.yaml

kjaskiewiczz (changes requested):

kjaskiewiczz:
I was also wondering if we need it here at all.

backend/pkg/redis/ratelimit.go
alfrunes: I covered the case where the params and error is nil as "no limit" returning a trivial reservation.

backend/services/deviceauth/config/config.go
alfrunes: I changed the configuration interface a little bit in order for the documentation to make sense in open source. I added a new parameter `quota_default` which allows configuring the fallback quota value in addition to renaming the previous `quotas` -> `quota_plan` which is only available in enterprise.

backend/pkg/rate/limit_test.go
alfrunes: Oops. Used to be 1s but changed it to 1h :)

kjaskiewiczz (dismissed):

kjaskiewiczz:
> // By setting the limit to 5, a single device can perform at most 5 requests per minute.
docker compose exec mongo mongosh --eval 'db.getSiblingDB("deviceauth").limits.insertOne({tenant_id: "", name: "max_devices", limit: 5})'

value:5 not limit:5

backend/services/deviceauth/config.yaml
```diff
@@ -147,8 +147,29 @@
...
+#     quota: 1.0
+#     quota_plan:
+#       os: 1.0
+#       professional: 1.5
+#       enterprise: 2.0
+# Using environment variables:
+# DEVICEAUTH_RATELIMITS_DEVICES_ENABLE: "false"
+# DEVICEAUTH_RATELIMITS_DEVICES_INTERVAL: "1m"
+# DEVICEAUTH_RATELIMITS_DEVICES_QUOTA_DEFAULT: "1.0"
+# DEVICEAUTH_RATELMIITS_DEVICES_QUOTA_PLAN: "professional=1.5 enterprise=2.0"
```
kjaskiewiczz: typo - RATELMIITS instead of RATELIMITS

kjaskiewiczz (dismissed):
looks good and works for me locally;
great job!

alfrunes:
> > // By setting the limit to 5, a single device can perform at most 5 requests per minute.
> > docker compose exec mongo mongosh --eval 'db.getSiblingDB("deviceauth").limits.insertOne({tenant_id: "", name: "max_devices", limit: 5})'
> 
> value:5 not limit:5

Ohh, no! Next time I should provide a snippet using the internal API instead :facepalm:
Sorry about the confusion.

alfrunes:
Resolved the conflict in `go.mod` by a rebase.

kjaskiewiczz (approved):

PR #204, opened 2024-11-18
chore(deployments): small improvement in the deployments docs

kjaskiewiczz:


backend/services/deployments/docs/management_api_v2.yml
```diff
@@ -62,13 +62,13 @@ paths:
         - name: id
           in: query
           description: |
-              Deployment identifier.
+              Deployment identifier. Can be repeated to query a set of deployments.
```
merlin-northern:
```suggestion
              Deployment identifier. You can provide it multiple times to query a set of deployments.
```

merlin-northern (dismissed):
the suggestion I left is according to an old rule "no passive voice in the docs" -- I have no idea if we still adhere to it.

tranchitella (dismissed):

backend/services/deployments/docs/management_api_v2.yml
kjaskiewiczz: done

mzedel (approved):

backend/services/deployments/docs/management_api_v2.yml
merlin-northern: 😘

PR #207, opened 2024-11-20
fix(gui): fixed an issue that prevented webhook creation as webhooks no longer can be edited, so their id was never adjusted

mzedel:


mineralsfree (approved):
LGTM :+1:

PR #208, opened 2024-11-20
feat(deployments): new endpoint for getting release by name

kjaskiewiczz:
Ticket: MEN-7575

### DIFF
--- a/backend/services/deployments/app/app_releases.go
+++ b/backend/services/deployments/app/app_releases.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Northern.tech AS
+// Copyright 2024 Northern.tech AS
 //
 //	Licensed under the Apache License, Version 2.0 (the "License");
 //	you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


backend/services/deployments/app/app_releases.go
```diff
@@ -141,3 +141,21 @@ func (d *Deployments) DeleteReleases(
...
+	switch err {
+	case nil:
+		return release, nil
+	case store.ErrNotFound:
+		return nil, ErrReleaseNotFound
+	default:
+		// Rewrite internal errors
+		log.FromContext(ctx).
+			Errorf("failed to get release from database: %s", err.Error())
+		return nil, ErrModelInternal
```
tranchitella, line 159: Why are we masking this error? It is the first time I see it.

tranchitella (approved):

backend/services/deployments/app/app_releases.go
kjaskiewiczz: It's the pattern we use in the app_releases, so I just applied it here without much thinking, tbh.
I wasn't the one to introduce it originally, maybe @merlin-northern knows something more?

merlin-northern: it is a good idea to do it. you never know what internal information the error will carry. I would keep it.

PR #211, opened 2024-11-25
MEN-7469-feat: updated upgrades and add-on page

mineralsfree:
Ticket: MEN-7469
Changelog: None


frontend/src/js/common-ui/dialogs/ConfirmAddon.tsx
```diff
@@ -0,0 +1,47 @@
+// Copyright 2024 Northern.tech AS
```
mzedel, line 1: not sure how much we can re-use these as `common-ui` components, but maybe we actually align with Alvaldi at some point 🤷 🤞

frontend/src/js/common-ui/dialogs/ConfirmUpgrade.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { onConfirm, onClose, newPlan, currentPlan, addOns } = props;
+  return (
+    <Dialog open={!!newPlan}>
+      <DialogTitle>Upgrade your plan to {newPlan.name}</DialogTitle>
+      <DialogContent>
+        <div className="margin-bottom-small">
+          You are currently subscribed to the <span className="bold">Mender {currentPlan.name}</span> plan.
+        </div>
+        <div className="margin-bottom-small">
+          Confirm to upgrade to <span className="bold">Mender {newPlan.name}</span>, billed at {newPlan.price}. <br />
```
mzedel:
```suggestion
          Confirm to upgrade to <b>Mender {newPlan.name}</b>, billed at {newPlan.price}. <br />
```

frontend/src/js/common-ui/dialogs/ConfirmUpgrade.tsx
```diff
@@ -0,0 +1,59 @@
...
+          {addOns.map(addon => {
+            return (
+              <>
+                The price of your <span className="bold">Mender {addon.name}</span> add-on package will change to {ADDONS[addon.name][newPlan.id].price}
```
mzedel: same as above

frontend/src/js/common-ui/forms/AddonSelect.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <InputLabel id="addons-selection-label">Select Addons</InputLabel>
+      <Select
+        multiple
+        value={selectedAddons}
+        onChange={onSelectedChange}
+        renderValue={selected => selected.map((role: AddonIds) => 'Mender ' + ADDONS[role].title).join(', ')}
+      >
+        {Object.values(ADDONS).map(addon => {
+          return (
+            <MenuItem style={{ width: 330 }} id={addon.id} key={addon.id} value={addon.id}>
```
mzedel:
```suggestion
            <MenuItem id={addon.id} key={addon.id} value={addon.id}>
```
to not mess with the selection

frontend/src/js/common-ui/forms/AddonSelect.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <FormControl id="addons-form" style={{ maxWidth: 330 }}>
+      <InputLabel id="addons-selection-label">Select Addons</InputLabel>
+      <Select
+        multiple
+        value={selectedAddons}
+        onChange={onSelectedChange}
+        renderValue={selected => selected.map((role: AddonIds) => 'Mender ' + ADDONS[role].title).join(', ')}
+      >
+        {Object.values(ADDONS).map(addon => {
+          return (
```
mzedel:
```suggestion
        {Object.values(ADDONS).map(addon => (
```
as long as there's no actual function body

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [message, setMessage] = useState('');
+  const [selectedAddons, setSelectedAddons] = useState<AddonIds[]>(addons.map(addon => addon.name));
+  return (
+    <Drawer anchor="right" open={true} PaperProps={{ style: { minWidth: '75vw' } }}>
+      <DrawerTitle onClose={onClose} title="Request for Mender Enterprise" />
+      <div className="flexbox column margin-bottom-large">
+        <div className="margin-top-large">
+          Get in touch with our team to request a quote for <span className="bold">Mender Enterprise.</span>{' '}
+        </div>
+        <FormControl style={{ marginBottom: 30, marginTop: 0, maxWidth: '550px' }}>
```
mzedel:
```suggestion
        </div>
        <FormControl className="margin-bottom margin-top-none" style={{ maxWidth: 550 }}>
```
the margin classes _should_ work, otherwise please choose a way to reference the size (px or not)

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const EnterpriseRequestExpanded = (props: EnterpriseRequestExpandedProps) => {
+  const { onSendRequest, onClose, addons } = props;
+  const [message, setMessage] = useState('');
+  const [selectedAddons, setSelectedAddons] = useState<AddonIds[]>(addons.map(addon => addon.name));
+  return (
+    <Drawer anchor="right" open={true} PaperProps={{ style: { minWidth: '75vw' } }}>
+      <DrawerTitle onClose={onClose} title="Request for Mender Enterprise" />
+      <div className="flexbox column margin-bottom-large">
+        <div className="margin-top-large">
+          Get in touch with our team to request a quote for <span className="bold">Mender Enterprise.</span>{' '}
```
mzedel:
```suggestion
          Get in touch with our team to request a quote for <b>Mender Enterprise.</b>{' '}
```

frontend/src/js/common-ui/forms/CountrySelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+      autoHighlight
+      renderInput={params => (
+        <TextField
+          {...params}
+          label="Country"
+          id={id || 'country'}
+          slotProps={{
+            htmlInput: {
+              ...params.inputProps,
+              autoComplete: 'new-password' // disable autocomplete and autofill
```
mzedel: 'new-password' for a country selection? This should rather be passed to the enclosing `<Form />` as an autocomplete disabling prop.

frontend/src/js/common-ui/forms/CountrySelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface CountrySelectProps {
+  id?: string;
+  onChange: (...event: any[]) => void;
+  [other: string]: any;
+}
+export const CountrySelect = (props: CountrySelectProps) => {
+  const { id, onChange, ...restProps } = props;
+  return (
+    <Autocomplete
+      getOptionLabel={option => option.label}
```
mzedel, line 35: using the `const regionNames = new Intl.DisplayNames(['en'], { type: 'region' });` browser knowledge, you could do 
```suggestion
      getOptionLabel={option => regionNames.of(option)}
```
and reduce maintaining the naming list/ translation

frontend/src/js/common-ui/forms/CountrySelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+      options={countries}
+      sx={{ width: 400 }}
```
mzedel: please align with the rest of the codebase

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+      country: { code },
+      state,
+      city,
+      line1,
+      postal_code
+    } = values;
+    const billing_profile = { email, name, address: { country: code, state, city, line1, postal_code } };
+    return dispatch(completeUpgrade({ tenantId: organization.id, plan: plan.id, billing_profile })).then(() => {
+      onCloseClick();
+      setTimeout(() => {
```
mzedel: This will have to either be tied to a ref and cleaned up on unmount or preferably be moved to the `completeUpgrade` thunk as it is likely dependent on the upgrade completion in any case.

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const dispatch = useDispatch();
+  const formSubmitRef = useRef<() => void | null>(null);
+  const { email } = useSelector(getCurrentUser);
+
+  const handleUpgrade = async () => {
+    if (formSubmitRef.current) {
+      formSubmitRef.current();
+      return;
+    }
+  };
```
mzedel: There is nothing being `await`ed in here, so the `async` isn't needed 👌 
And the return won't exit earlier than without it, so it's not needed either 👌

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+const { setSnackbar } = storeActions;
+
+interface PlanExpandedProps {
+  plan: Plan | null;
+  organization: any;
+  onCloseClick: () => void;
+}
+
+export const PlanExpanded = (props: PlanExpandedProps) => {
+  const { plan, onCloseClick, organization } = props;
```
mzedel: Since you have a selector in this component already you could also select the organization from here - in any case I would rename the plan to a `selectedPlan` as there is a plan in the organization object too.

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+      onCloseClick();
+      setTimeout(() => {
+        dispatch(getDeviceLimit());
+      }, TIMEOUTS.threeSeconds);
+    });
+  };
+  return (
+    <Drawer anchor="right" open={!!plan} PaperProps={{ style: { minWidth: '75vw' } }}>
+      {plan && (
+        <>
```
mzedel: Should this not be covered by the `open={!!plan}`? if not: would it be possible to `return null` before the `<Drawer />` to avoid having this condition nest the entire content?

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+              <TextInput hint="Zip or Postal code" required label="Zip or Postal code" id="postal_code" validations="isLength:4,trim" />
+              <ControlledCountrySelect required id={'country'} />
+            </div>
+          </Form>
+          {isValid && (
+            <>
+              <h4>Card Details</h4>
+              <CardSection organization={organization} onComplete={handleUpgrade} setSnackbar={message => dispatch(setSnackbar(message))} isSignUp={true} />
+            </>
+          )}
```
mzedel: in order to avoid the `onSubmit`/ `setIsValid` workaround, the form content should be moved to a component that then has access to the validity, dirtyFields and quite sure also the submit handler via `useFormContext`

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -0,0 +1,109 @@
...
+              <TextInput hint="Company name" label="Company name" id="name" required validations="isLength:2,trim" />
+            </div>
+            <h4>Billing address</h4>
+            <div className="flexbox column">
+              <TextInput hint="Address line 1" label="Address line 1" required id="line1" validations="isLength:3,trim" />
+              <TextInput hint="State" label="State" required id="state" validations="isLength:2,trim" />
+
+              <TextInput hint="City" label="City" required id="city" validations="isLength:2,trim" />
+              <TextInput hint="Zip or Postal code" required label="Zip or Postal code" id="postal_code" validations="isLength:4,trim" />
+              <ControlledCountrySelect required id={'country'} />
```
mzedel:
```suggestion
              <ControlledCountrySelect required id="country" />
```

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
 
 import { useStyles } from './planselection';
+import { addOnsToString, clearLocalStorageEntry, updateLocalStorage } from './upgrade.js';
 
-export const AddOnSelection = ({ addons = [], features, onChange, updatedPlan = PLANS.os.id }) => {
+export const AddButton = withStyles(Button, ({ palette }) => ({
+  root: {
+    backgroundColor: palette.background.default,
+    textTransform: 'none',
+    color: `${palette.primary.main}`,
```
mzedel:
```suggestion
    color: palette.primary.main,
```

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
-import InfoText from '@northern.tech/common-ui/infotext';
+import { ConfirmAddon } from '@northern.tech/common-ui/dialogs/ConfirmAddon';
 import { ADDONS, PLANS } from '@northern.tech/store/constants';
+import { requestPlanChange } from '@northern.tech/store/thunks';
 
 import { useStyles } from './planselection';
+import { addOnsToString, clearLocalStorageEntry, updateLocalStorage } from './upgrade.js';
 
-export const AddOnSelection = ({ addons = [], features, onChange, updatedPlan = PLANS.os.id }) => {
+export const AddButton = withStyles(Button, ({ palette }) => ({
```
mzedel: Both this and the `RemoveButton` are not buttons in the design, if you want to use a button then there is the `text` variant or you could use a regular `<button />` and style it

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+    textTransform: 'none',
+    color: `${palette.primary.main}`,
+    padding: '5px 7px'
+  }
+}));
+export const RemoveButton = withStyles(Button, ({ palette }) => ({
+  root: {
+    backgroundColor: palette.background.default,
+    textTransform: 'none',
+    color: '#921267 !important'
```
mzedel: this is likely handled in the palette (or a color very close to it) - if not, it should be handled through the palette.

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+import { Button } from '@mui/material';
+import { withStyles } from 'tss-react/mui';
 
-import InfoText from '@northern.tech/common-ui/infotext';
+import { ConfirmAddon } from '@northern.tech/common-ui/dialogs/ConfirmAddon';
 import { ADDONS, PLANS } from '@northern.tech/store/constants';
+import { requestPlanChange } from '@northern.tech/store/thunks';
 
 import { useStyles } from './planselection';
+import { addOnsToString, clearLocalStorageEntry, updateLocalStorage } from './upgrade.js';
```
mzedel:
```suggestion
import { addOnsToString, clearLocalStorageEntry, updateLocalStorage } from './upgrade';
```

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
-    () =>
-      Object.entries(ADDONS).reduce((accu, [addOnName, addOn]) => {
-        if (!addOn.needs.every(need => features[need])) {
+    () => {
+      const currentState = JSON.parse(localStorage.getItem(org.id + '_upgrades')) || {};
+      return Object.entries(ADDONS)
+        .reduce((accu, [addOnName, addOn]) => {
+          // if (!addOn.needs.every(need => features[need])) {
+          //   return accu;
+          // }
```
mzedel:
```suggestion

```

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
-              {/* eslint-disable-next-line react/jsx-no-target-blank */}
-              <a className="margin-left-small" href="https://mender.io/plans/features" target="_blank" rel="noopener">
-                Learn more
-              </a>
-            </div>
-          );
-        })}
+        <div className="flexbox column">
+          {relevantAddons.map(addOn => {
+            const isEligible = addOn.eligible.indexOf(currentPlan) > -1;
```
mzedel: I know it's preexisting, but this would probably be best done in the memoization too

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+                <a
+                  className={`${classes.link} flexbox center-aligned bold`}
+                  href={`https://mender.io/pricing/add-ons/${addOn.name}`}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  Mender {addOn.title}
+                  <LaunchIcon className="link-color margin-left-x-small" fontSize="small" />
+                </a>
+                {isUpgrade && currentPlanName !== 'Enterprise' ? (
```
mzedel: since we had the plan names change before this should likely check for the id:
```suggestion
                {isUpgrade && currentPlan.id !== PLANS.enterprise.id ? (
```

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+                    {isEligible ? (
+                      <>
+                        starting <span className="bold">{addOn[currentPlan].price}</span>
+                      </>
+                    ) : (
+                      <>not available on {currentPlanName} plan</>
+                    )}
+                  </div>
+                ) : (
+                  currentPlanName === 'Enterprise' && <div />
```
mzedel: as above

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+                )}
+                {addOn.isEnabled ? (
+                  <>
+                    <div className={classes.chip}>active</div>
+                    {addOn.pending && !addOn.pending.isAdd ? (
+                      <div className={`${classes.chip} margin-left-small`}>removal pending</div>
+                    ) : (
+                      <RemoveButton disabled={!isEligible} onClick={e => onAddOnClick(e, addOn.name, false)}>
```
mzedel:
```suggestion
                      <RemoveButton disabled={!isEligible} onClick={e => onAddOnClick(addOn.name, false)}>
```
as you're no longer using the `e`

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+                        Remove from plan
+                      </RemoveButton>
+                    )}
+                  </>
+                ) : (
+                  <>
+                    {addOn.pending && addOn.pending.isAdd ? (
+                      <div className={classes.chip}>pending</div>
```
mzedel: This and the above should use https://mui.com/material-ui/react-chip/

frontend/src/js/components/settings/addonselection.js
```diff
@@ -11,77 +11,160 @@
...
+                    ) : (
+                      <AddButton disabled={!isEligible} onClick={e => (isEligible ? onAddOnClick(e, addOn.name, true) : () => false)}>
+                        Add to plan
+                      </AddButton>
+                    )}
+                    <div style={{ width: '140px' }} />
```
mzedel: is this for padding only?

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
-import InfoText from '@northern.tech/common-ui/infotext';
 import { PLANS } from '@northern.tech/store/constants';
-import { isDarkMode } from '@northern.tech/store/utils';
+
+import { clearLocalStorageEntry, isUpgrade } from './upgrade.js';
 
 export const useStyles = makeStyles()(theme => ({
-  planNote: { marginBottom: -11, fontSize: 'smaller' },
+  planName: { margin: 0 },
+  planPrice: { fontSize: '22px' },
```
mzedel: this is likely a heading instead of a `fontSize` as it aligns with the `Upgrade now`

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
+  devicesCount: { marginLeft: '8px' },
+  planIcon: { color: theme.palette.primary.main, fontSize: '16px' },
+  planFeature: { marginBottom: '8px', fontSize: '12px' },
+  planFeatureText: { marginLeft: '8px' },
```
mzedel: all the margins here should either be `theme.spacing()` or using a `margin-...-x-small` class

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
-  price: { fontSize: '1rem' }
+  chip: {
+    padding: '7px 10px',
+    color: theme.palette.grey[800],
+    textTransform: 'uppercase',
+    background: theme.palette.tooltip.tierTipBackground,
+    borderRadius: theme.spacing(3),
+    fontSize: '11px',
+    fontWeight: 'bold'
+  },
```
mzedel: this should not be needed as the mui chip should be used

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
+                    {item.deviceCount}
+                    {item.id === 'enterprise' ? '' : '*'}
+                  </div>
+                </div>
+              </div>
+              <ul className="unstyled">
+                {item.features.map((feature, index) => (
+                  <li className={`flexbox ${classes.planFeature}`} key={`${item.id}-feature-${index}`}>
+                    <CheckIcon className={classes.planIcon} />
+                    <div className={classes.planFeatureText}>{feature}</div>{' '}
```
mzedel:
```suggestion
                    <div className={classes.planFeatureText}>{feature}</div>
```
I guess

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
-              ) : (
-                <>
-                  <div className={`link-color bold ${classes.price}`}>{item.price}</div>
-                  <div>{item.deviceCount}</div>
-                </>
-              )}
+              <p className={`${classes.planName}`}>Mender {item.name}</p>
+              <h4 className={`${classes.planPrice}`}>{item.price}</h4>
+              <div className={`margin-bottom`}>
+                <div className={`flexbox align-center`}>
```
mzedel:
```suggestion
              <div className="margin-bottom">
                <div className="flexbox align-center">
```

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
+                  <DeveloperBoardIcon fontSize="medium" className={`${classes.planIcon}`} />
+                  <div className={`${classes.devicesCount}`}>
```
mzedel:
```suggestion
                  <DeveloperBoardIcon fontSize="medium" className={classes.planIcon} />
                  <div className={classes.devicesCount}>
```

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
+              </ul>
             </div>
-            <ul className="unstyled">
-              {item.features.map((feature, index) => (
-                <li key={`${item.id}-feature-${index}`}>
-                  <InfoText variant="dense">{feature}</InfoText>
-                </li>
-              ))}
-            </ul>
+            <Button
```
mzedel: Considering that the `disabled` check and the content are not straight forward, this would probably best be extracted into a separate component

frontend/src/js/components/settings/planselection.js
```diff
@@ -11,82 +11,119 @@
...
+              onClick={() => onPlanSelect(item.id)}
+              variant="contained"
+              color={item.id === 'enterprise' ? 'secondary' : 'primary'}
+              disabled={
+                (item.id === currentPlan && !isTrial) ||
+                (isUpgradePending && item.id === isUpgradePending.plan) ||
+                (currentPlan === 'enterprise' && item.id !== 'enterprise') ||
+                (currentPlan === 'professional' && item.id === 'os')
+              }
+              className={`${classes.plantton}`}
```
mzedel:
```suggestion
              className={classes.plantton}
```
and the class referenced is probably not right

frontend/src/js/components/settings/upgrade.js
```diff
@@ -13,53 +13,53 @@
...
+  localStorage.setItem(orgId + '_upgrades', JSON.stringify(currentState));
+};
+export const clearLocalStorageEntry = (orgId, name) => {
+  const currentState = JSON.parse(localStorage.getItem(orgId + '_upgrades')) || {};
+  delete currentState[name];
+  localStorage.setItem(orgId + '_upgrades', JSON.stringify(currentState));
+};
+export const isUpgrade = orgId => {
+  const currentState = JSON.parse(localStorage.getItem(orgId + '_upgrades')) || {};
+  if (currentState['enterprise']) {
```
mzedel:
```suggestion
  if (currentState.enterprise) {
```
and below... or maybe get them in a list of plans to check and minimize duplication..

frontend/src/js/components/settings/upgrade.js
```diff
@@ -13,53 +13,53 @@
...
+    return { plan: 'enterprise' };
+  } else if (currentState['professional']) {
+    return { plan: 'professional' };
+  } else {
+    return false;
```
mzedel: there is probably a better return type than a `{ plan: string } | boolean` 😉

frontend/src/js/components/settings/upgrade.js
```diff
@@ -74,15 +74,14 @@ const upgradeNotes = {
...
   const [updatedPlan, setUpdatedPlan] = useState(PLANS.os.id);
-  const [upgraded, setUpgraded] = useState(false);
-  const navigate = useNavigate();
+  const [selectedPlan, setSelectedPlan] = useState(null);
+
   const dispatch = useDispatch();
   const features = useSelector(getFeatures);
   const org = useSelector(getOrganization);
-  const { addons: orgAddOns = [], plan: currentPlan = PLANS.os.id, trial: isTrial = true } = org;
+  const { addons: orgAddOns = [], plan: currentPlan = '', trial: isTrial = true } = org;
```
mzedel: the org should always have a plan, but if it doesn't the lowest plan should be assumed - is there a reason why you removed this for the empty string (here and elsewhere)?

frontend/src/js/components/settings/upgrade.js
```diff
@@ -101,89 +100,69 @@ export const Upgrade = () => {
...
+        <PlanExpanded plan={selectedPlan} organization={org} onCloseClick={() => setSelectedPlan(null)} />
+      ) : updatedPlan === PLANS.enterprise.id && selectedPlan ? (
+        <EnterpriseRequestExpanded addons={addOns} onClose={() => setSelectedPlan(null)} onSendRequest={onEnterpriseRequest} />
+      ) : (
+        selectedPlan &&
+        selectedPlan.id !== currentPlan && (
+          <ConfirmUpgrade
+            currentPlan={PLANS[currentPlan]}
+            onClose={() => setSelectedPlan(null)}
+            onConfirm={() => onSendRequest()}
```
mzedel:
```suggestion
            onConfirm={onSendRequest}
```

frontend/src/less/main.less
```diff
@@ -888,19 +888,27 @@ div:focus {
...
   &.muted,
   &.muted:hover {
     cursor: default;
     box-shadow: none;
   }
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  margin-right: 16px;
+  border-color: #E3E3E3;
```
mzedel: This is better defined in the components to ease theming.

frontend/src/less/main.less
```diff
@@ -888,19 +888,27 @@ div:focus {
...
 .planPanel {
-  padding: 10px 20px;
+  padding: 22px 16px;
   border-radius: 5px;
   border-width: 1px;
   border-style: solid;
   vertical-align: top;
-  width: 32%;
+  height: 367px;
+  width: 192px;
```
mzedel: these won't work for different screen sizes.

mzedel (changes requested):
I will have to go through it another time and haven't tested it locally yet, but I'll leave it with the current comments.
Apart from what I've noted in the code:
- Please try to structure the changes in different commits the next time.
- Please add some tests to check what the localstorage interactions look like, the drawer & dialog components.
- Please adjust the e2e tests to work with your changes.

frontend/src/js/common-ui/forms/CountrySelect.tsx
mineralsfree: This one is from MUI autocomplete example, this way it that doesn't mess with the autofill, as user needs to actually select the country

frontend/src/js/components/settings/addonselection.js
mineralsfree: by the way, is it still relevant?

frontend/src/js/components/settings/addonselection.js
mineralsfree: Doesn't look like heavy computation to me. I believe that makes sense to use memo whenever we encounter performance issues (e.g. fps drop) :thinking: What do you think?

frontend/src/js/common-ui/forms/CountrySelect.tsx
mzedel: For a `freeSolo` input I could maybe see the option (defaulting to random chars instead of something preselected), for a country selection with a limited selection set I think this can have autocompletion.

frontend/src/js/components/settings/addonselection.js
mzedel: doesn't look like it, all the add ons are now active on hM

frontend/src/js/components/settings/addonselection.js
mzedel: IIRC the memoization was need to have a stable interaction with the checkbox - but feel free to remove it 👌

frontend/src/js/components/settings/upgrade.test.js
```diff
@@ -64,4 +78,127 @@ describe('Upgrade Component', () => {
...
+    const upgradeButton = await screen.getAllByRole('button', { name: /subscribe/i })[0];
+    await user.click(upgradeButton);
+    await user.type(await screen.getByRole('textbox', { name: /address line 1/i }), 'Blindernveien');
+    await user.type(await screen.getByRole('textbox', { name: /state/i }), 'Oslo');
+    await user.type(await screen.getByRole('textbox', { name: /city/i }), 'Oslo');
+    await user.type(await screen.getByRole('textbox', { name: /zip or postal code/i }), '1234');
+    await user.type(await screen.getByRole('combobox', { name: /country/i }), 'Norw');
+    await user.keyboard('[ArrowUp]');
+    await user.keyboard('[Enter]');
+    //TODO: test stripe
```
mzedel: we can't really test stripe here, but you could add an `expect()` for the autocomplete maybe? or the absence of errors maybe?

frontend/src/js/components/settings/AddonSelection.tsx
```diff
@@ -0,0 +1,214 @@
...
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    borderTop: '1px solid',
+    borderColor: theme.palette.grey[300],
+    minHeight: '35px',
+    '&:first-of-type': {
+      border: 'none'
+    }
+  },
+  link: { width: '170px' },
```
mzedel: this will likely break depending on the screen size

frontend/src/js/components/settings/AddonSelection.tsx
```diff
@@ -0,0 +1,214 @@
...
+const useStyles = makeStyles()(theme => ({
+  chip: {
+    color: theme.palette.grey[800],
+    textTransform: 'uppercase',
+    background: theme.palette.tooltip.tierTipBackground,
+    borderRadius: theme.spacing(3),
+    fontSize: '11px',
+    fontWeight: 'bold',
+    height: '27px'
+  },
```
mzedel: this still isn't relying on the chip component

frontend/src/js/components/settings/AddonSelection.tsx
```diff
@@ -0,0 +1,214 @@
...
+  price: { weight: 400, flexGrow: 2 },
+  addButton: {
+    color: `${theme.palette.primary.main} !important`,
```
mzedel: As a rule of thumb also for the style definitions in here: if you have to rely on `!important` you're likely missing something. Here I would suggest:
```js
button: {
    textTransform: 'none',
    padding: '2px 5px',
    [`&.${buttonClasses.text}.addButton`]: {
          color: theme.palette.primary.main
    },
    [`&.${buttonClasses.text}.removeButton`]: {
          color: theme.palette.secondary.lighter
    }
}
```
and then for the buttons:
```jsx
<Button 
    className={`${classes.button} removeButton`} // or addButton
    color="secondary" // or primary
    size="small"
    ...
```

frontend/src/js/store/appSlice/constants.ts
```diff
@@ -47,37 +47,67 @@ export const locations = {
...
+  needs: string[];
+  eligible: AvailablePlans[];
+} & {
+  [key in Exclude<AvailablePlans, 'enterprise'>]: {
+    price: string;
+    deviceCount: string;
+  };
+};
+
+export type AddonIds = keyof typeof ADDONS;
```
mzedel: While referring to multiple an instance of the type is an `AddonId` right? with this the usage of `AddonIds[]` would IMO imply `[[],[],...]` (same with `AvailablePlans`) - what do you think?

frontend/src/js/components/settings/AddonSelection.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    ) : (
+                      <>not available on {currentPlanName} plan</>
+                    )}
+                  </div>
+                ) : (
+                  currentPlan === PLANS.enterprise.id && <div />
+                )}
+                {addOn.isEnabled ? (
+                  <>
+                    <Chip className={`${classes.chip} margin-right-small`} label="active" />
```
mzedel:
```
<Chip className={`${classes.chip} muted margin-right-small`} label="active" size="small" />
```
then the chip class would just be:
```
chip: {
    textTransform: 'uppercase'
},
```
and it would align with the rest of the UI

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+      boxShadow: 'none'
+    }
+  }
+}));
+
+const isPlanDisabled = (item: Plan, currentPlan: AvailablePlans, isTrial: boolean, isUpgradePending: Partial<{ plan: AvailablePlans }>) => {
+  return (
+    (item.id === currentPlan && !isTrial) ||
+    (isUpgradePending && item.id === isUpgradePending.plan) ||
+    (currentPlan === PLANS.professional.id && item.id === PLANS.os.id)
```
mzedel: you have the `canUpgrade` a few lines below - that should likely apply here...

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <div
+            key={item.id}
+            className={`planPanel ${classes.planPanel} ${updatedPlan === item.id ? 'active' : ''} ${isTrial || canUpgrade(item.id) ? '' : 'muted'}`}
+          >
+            <div>
+              <p className="margin-none">Mender {item.name}</p>
+              <h2>{item.price}</h2>
+              <div className="margin-bottom">
+                <div className="flexbox align-center">
+                  <DeveloperBoardIcon fontSize="medium" className={`${classes.planIcon}`} />
```
mzedel:
```jsx
<DeveloperBoardIcon fontSize="medium" className={classes.planIcon} />
```
once again

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    <div className="margin-left-x-small">{feature}</div>{' '}
+                  </li>
+                ))}
+              </ul>
+            </div>
+            <Button
+              onClick={() => onPlanSelect(item.id)}
+              variant="contained"
+              color={item.id === PLANS.enterprise.id ? 'secondary' : 'primary'}
+              disabled={isPlanDisabled(item, currentPlan, isTrial, isUpgradePending)}
```
mzedel, line 106: the discrepancy between the upgradability check in the `isPlanDisabled` leads to this appearing clickable despite the plan showing as disabled

mzedel (changes requested):

frontend/src/js/components/settings/AddonSelection.tsx
mineralsfree: I've checked, and it doesn't affect the layout in a bad way on a smaller screen sizes

frontend/src/js/store/appSlice/constants.ts
mineralsfree: Do you suggest to name it differently? AddonIds[] is just array of strings - `[ "configure", "monitor" ]`, it's not `[ ["configure"], ["monitor"] ]`.

frontend/src/js/components/settings/AddonSelection.tsx
mzedel: This example disagrees:
Screenshot 2024-12-05 at 10 02 18

frontend/src/js/store/appSlice/constants.ts
mzedel: Exactly, the `AddonIds` naming suggests it's a multiple/ collection/ list of `AddonId` - so an array of a multiple would then suggest a list of lists. So `AddonId[]` vs. `AddonIds[]`

frontend/src/js/components/settings/AddonSelection.tsx
mineralsfree: Removing `width: '170px'` doesn't change much in terms of line breaks on screen sizes `<1200px`. Plan panel starts to break the lines even earlier. What would you suggest?

mzedel: The screenshot was from my 1800px wide screen... one suggestion could be:
```
    display: 'grid',
    gridTemplateColumns: 'max-content 1fr max-content max-content'
```

frontend/src/js/components/settings/AddonSelection.tsx
```diff
@@ -0,0 +1,220 @@
...
+      <div className="flexbox column">
+        <p>
+          Enhance your Mender plan with optional add-on packages – offering additional features to easily manage your software and devices over-the-air. Below
+          you can request which add-ons you’d like to be included with your plan.
+        </p>
+        <div className={classes.planPanelAddon}>
+          {relevantAddons.map(addOn => {
+            const addonItemClasses = `${addOn.isEnabled ? 'active' : ''} ${addOn.isEligible ? '' : 'muted'} ${classes.planPanelAddonItem}`;
+            return (
+              <>
```
mzedel:
```suggestion
              <Fragment key={addOn.name}>
```
to not cause react errors

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,124 @@
...
+                  <DeveloperBoardIcon fontSize="medium" className={classes.planIcon} />
+                  <div className="margin-left-x-small">
+                    {item.deviceCount}
+                    {item.id === PLANS.enterprise.id ? '' : '*'}
+                  </div>
+                </div>
+              </div>
+              <ul className="unstyled">
+                {item.features.map((feature, index) => (
+                  <li className={`flexbox margin-bottom-x-small ${classes.planFeature}`} key={`${item.id}-feature-${index}`}>
```
mzedel:
```suggestion
                  <li className="flexbox margin-bottom-x-small slightly-smaller" key={`${item.id}-feature-${index}`}>
```
to align with the rest of the UI

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,124 @@
...
+const useStyles = makeStyles()(theme => ({
+  planIcon: { color: theme.palette.primary.main, fontSize: '16px', alignSelf: 'center' },
+  planFeature: { fontSize: '12px' },
+  planButton: { alignSelf: 'center' },
+  planPanel: {
+    padding: '22px 16px',
+    borderRadius: '5px',
+    verticalAlign: 'top',
+    width: '32%',
+    transition: 'all 0.2s ease',
```
mzedel: This and the `boxShadow` reset below seem to no longer being in use after the `.active` removal.

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,124 @@
...
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: 'space-between',
+    marginRight: theme.spacing(2),
+    border: `1px solid ${theme.palette.grey[600]}`,
+    fontSize: '14px',
+    fontWeight: 500,
+    '&.muted, &.muted:hover': {
+      cursor: 'default',
+      boxShadow: 'none'
```
mzedel: see the comment on the transition 👆

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,124 @@
...
+  planIcon: { color: theme.palette.primary.main, fontSize: '16px', alignSelf: 'center' },
+  planFeature: { fontSize: '12px' },
+  planButton: { alignSelf: 'center' },
+  planPanel: {
+    padding: '22px 16px',
+    borderRadius: '5px',
+    verticalAlign: 'top',
+    width: '32%',
+    transition: 'all 0.2s ease',
+    cursor: 'pointer',
```
mzedel: for the entire panel this seems to be implying that they respond to clicks - maybe you can add the `onPlanSelect` on the entire panel or otherwise I would remove the cursor here

frontend/src/js/components/settings/PlanSelection.tsx
```diff
@@ -0,0 +1,124 @@
...
+const isPlanDisabled = (item: Plan, currentPlan: AvailablePlans, isTrial: boolean, isUpgradePending: Partial<{ plan: AvailablePlans }>) => {
+  return (isUpgradePending.plan && !canUpgrade(item.id, isUpgradePending.plan)) || (!isTrial && !canUpgrade(item.id, currentPlan));
+};
+const isPlanMuted = (item: Plan, currentPlan: AvailablePlans, isTrial: boolean, isUpgradePending: Partial<{ plan: AvailablePlans }>) => {
+  return (
+    isPlanDisabled(item, currentPlan, isTrial, isUpgradePending) &&
+    (item.id !== currentPlan || isTrial) &&
+    !(isUpgradePending.plan && isUpgradePending.plan === item.id)
+  );
+};
```
mzedel: this seems to still have an issue, the following is on a basic plan, with enterprise requested:
Screenshot 2024-12-09 at 11 37 12

with the first panel appearing clickable + not muted

frontend/src/js/store/commonConstants.tsx
```diff
@@ -271,3 +271,259 @@ export const emptyRole: UiRoleDefinition = Object.freeze({
...
+  'VU',
+  'WF',
+  'WS',
+  'XK',
+  'YE',
+  'YT',
+  'ZA',
+  'ZM',
+  'ZW'
+].map(code => ({ code, label: regionNames.of(code) }));
```
mzedel: after trying this out: please sort these by label - it leads to a very confusing list otherwise 🙃

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
```diff
@@ -0,0 +1,63 @@
...
+        </FormControl>
+        <AddonSelect initialState={selectedAddons} onChange={setSelectedAddons} />
+        <div className="margin-top margin-bottom">
+          <Button onClick={onClose}>Cancel</Button>
+          <Button
+            className="margin-left-small"
+            color="secondary"
+            disabled={!message}
+            onClick={() => onSendRequest({ message, selectedAddons })}
+            style={{ alignSelf: 'flex-start' }}
```
mzedel: this won't apply here

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const EnterpriseRequestExpanded = (props: EnterpriseRequestExpandedProps) => {
+  const { onSendRequest, onClose, addons } = props;
+  const [message, setMessage] = useState('');
+  const [selectedAddons, setSelectedAddons] = useState<AddonId[]>(addons.map(addon => addon.name));
+  return (
+    <Drawer anchor="right" open={true} PaperProps={{ style: { minWidth: '75vw' } }}>
+      <DrawerTitle onClose={onClose} title="Request for Mender Enterprise" />
+      <div className="flexbox column margin-bottom-large">
+        <div className="margin-top-large">
+          Get in touch with our team to request a quote for <b>Mender Enterprise.</b>{' '}
```
mzedel:
```suggestion
          Get in touch with our team to request a quote for <b>Mender Enterprise.</b>
```

frontend/src/less/main.less
```diff
@@ -885,47 +889,6 @@ div:focus {
...
-.offerBox {
-  background-color: #337a87;
-  border-radius: 4px;
-  padding: 8px 12px;
-  color: @defaultBackground;
-  * {
-    color: inherit;
-    font-size: inherit;
-  }
-}
```
mzedel: This is still in use in the `OfferHeader` component - even though I'm not sure when we'll have another offer

mzedel (changes requested):
Some more things I discovered when testing locally + the `offerBox` class. Apart from that it would be nice if you could align the function definitions with the rest of the codebase (as in: no function body for straight returns).

frontend/src/js/components/settings/PlanSelection.tsx
mineralsfree: I'll remove `cursor: pointer` from the `planPanel`. otherwise first panel shouldn't be muted [according to the design](https://www.figma.com/design/0xHrBYtRf2LH9XNjFgjPI3/Billing-contact?node-id=3989-2475&node-type=frame&t=7M5a66Gkgu4pGLQT-0) as it's a current plan

mzedel (dismissed):
Very nice, thanks for sticking with it despite all the comments!

mzedel (approved):

PR #221, opened 2024-11-26
MEN-7773 - fix(gui): fixed an issue that prevented showing role details in user list or user details for SP users

mzedel:
Ticket: MEN-7773
Changelog: None


frontend/src/js/components/settings/user-management/userdefinition.js
```diff
@@ -116,11 +120,11 @@ export const UserDefinition = ({ currentUser, isEnterprise, onCancel, onSubmit,
 
   const { areas, groups } = useMemo(() => {
     const emptySelection = { areas: {}, groups: {}, releases: {} };
-    if (!(selectedRoles && roles)) {
+    if (!(selectedRoles && rolesById)) {
```
mineralsfree: rolesById seems to be always truthy ({}) even when empty

mineralsfree (changes requested):

frontend/src/js/components/settings/user-management/userdefinition.js
mzedel: indeed and `selectedRoles` too 😞 - should now be better 👌

mineralsfree (approved):

PR #225, opened 2024-11-26
chore: re-introduced staging e2e tests after monorepo migration

mzedel:
running after all merges to main as we can't check for gui changes only

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -354,7 +354,7 @@ test:prep:
     - mv mender-stress-test-client frontend/tests/e2e_tests/
     - cd frontend/tests/e2e_tests
     - curl -fsSL https://get.docker.com | sh
-    - docker pull mendersoftware/mender-client-docker-addons:mender-3.6.x
+    - docker pull mendersoftware/mender-client-docker-addons:mender-master
     - npm ci --cache .npm --prefer-offline
     - npx playwright install
   script:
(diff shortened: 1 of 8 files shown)


oldgiova (dismissed):
LGTM; a possible later improvement is to create a loop which is checking the version deployed on staging, querying for the header `x-mender-version`; if the value is the same as the pipeline_id, then the deployment is successful and we could move on and make the tests. 

For example in Alvaldi:
```
curl -L https://staging.app.alvaldi.com -sI | grep x-alvaldi-version | head -n1 | awk '{print $2}'
2.0.0-1561065228-staging+2
```

mzedel:
> LGTM; a possible later improvement is to create a loop which is checking the version deployed on staging, querying for the header `x-mender-version`;

that header doesn't seem to exist yet, but we can adjust once it does 👌

oldgiova (approved):
Great stuff! This is a nice improvement.

We just have to follow up pretty soon with the x-mender-version header: without it, we go blind with the actual staging version; it could be that for some reason the staging version tagged is rolled back by helm automatically, so the e2e would test the previous version, hidding possible problems.
This is not blocking, though, because also the subsequent production upgrade should suffer of the same helm rollback.

So my idea is to provide the x-mender-version header in the application, so instead of the 
```
  start_in: 10 minutes
```
we can add a ci loop which is polling the `x-mender-version` header, and the e2e starts only when the version appears, which means the staging is actually updated.

PR #227, opened 2024-11-27
MEN-7788: Use Virtual Device for OS updates in docker composition

lluiscampos:
For the upcoming release, we will use the same Virtual Device that we had in the Mender 3.x series. After the release, we will work on migrating to the Virtual Device for app updates consistently across docker composition, ui and docs. See more context in the parent JIRA epic related to this task.

 For the e2e tests, use an override image for the `client` service so that we use the new Virtual Device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,15 @@ To initialize the admin user, use the following snippet in step 3:
 MENDER_NAME=Admin
 MENDER_USERNAME=admin@docker.mender.io
 MENDER_PASSWORD=password123
-docker compose exec tenantadm tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME"
+TENANT_ID=$(docker compose exec tenantadm tenantadm create-org --username "$MENDER_USERNAME" --password "$MENDER_PASSWORD" --name "$MENDER_NAME)"
+```
+
+To add a virtual client with the composition, use the following snippet in step 5:
+
+```bash
+TENANT_TOKEN=$(docker compose exec tenantadm tenantadm get-tenant --id "$TENANT_ID" | jq -r .tenant_token)
+export TENANT_TOKEN
+docker compose run -d client
 ```
 
 Visit [https://localhost](https://localhost) and sign in using the credentials from the snippet above.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -359,25 +359,12 @@ services:
         aliases: [mender-nats]
 
   client:
-    image: mendersoftware/mender-client-docker-addons:mender-master
+    image: mendersoftware/mender-client-qemu:mender-master
     scale: 0
-    configs:
-      - source: client_json
-        target: /etc/mender/mender.conf
-    volumes:
-      - ./compose/certs/mender.crt:/var/lib/mender/mender.crt
-
-configs:
-  client_json:
-    content: |
-      {
-        "InventoryPollIntervalSeconds": 5,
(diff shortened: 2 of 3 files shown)


mzedel (commented):
Just to state the obvious: testing this locally it doesn't work as I'm on an arm Mac - so I don't think it's ok for me to approve this.

mzedel:
To put it into a different context: with the current state I can explore most of the functionality, with this change I can't even see let alone accept a device. also @estenberg

lluiscampos:
@mzedel I very well understand your concerns. And I think that it is great that you keep voicing this topic in behalf of all Mac ARM users out there.

This PR is a follow-up of yesterday discussion in Grooming meeting with everyone present where the agreement was to postpone the migration to the Virtual Device for app updates to after the release (not to abandon it). We can revert this commit the very same day that we branch out 4.0.x in this repo.

mzedel:
> This PR is a follow-up of yesterday discussion in Grooming meeting with everyone present where the agreement was to postpone the migration to the Virtual Device for app updates to after the release (not to abandon it). We can revert this commit the very same day that we branch out 4.0.x in this repo.

Completely understand, I just want to point it out at every step of the way as it was ignored for more than 1,5 years (this was likely present well before https://northerntech.atlassian.net/browse/MEN-6407) and now that we finally have it somewhat usable we again make the user experience worse - for every user that tries 4.0 or comes to this repository via the notices in the old service repos. Considering that usability concern have been ignored or pushed back against during grooming meetings long enough I hope this is a more noticeable place to bring these concerns up.
So as a user in the team that is affected by this change I can't approve of it as I will no longer be able to use the composition as is, as long as we have the qemu client - let alone an outside user trying this out.

estenberg:
We will get this @mzedel , it's just a very bad time to introduce it in the middle of the release process, and the fact that it was partially there was not intended (we can't release half-done replacement). I plan to take it right after the release.

kjaskiewiczz (dismissed):

lluiscampos:
@mzedel I have modified the e2e tests setup to use there the Virtual Device for app updates. You may want to look at that part :eyes: 

@kjaskiewiczz I need your review again because updating the PR dismisses existing approvals :+1:

mzedel (dismissed):
can you please put the image reference for the device into the e2e compose [here](https://github.com/mendersoftware/mender-server/blob/main/frontend/tests/e2e_tests/docker-compose.e2e-tests.yml#L56-L64) - last we had the qemu client, the tests didn't work with the qemu client and on the main e2e devs machines it doesn't work either + it's a smaller change 🙃

lluiscampos:
> can you please put the image reference for the device into the e2e compose [here](https://github.com/mendersoftware/mender-server/blob/main/frontend/tests/e2e_tests/docker-compose.e2e-tests.yml#L56-L64) -

I didn't realize that we already had an override for the client service there. Thanks!

> last we had the qemu client, the tests didn't work with the qemu client and on the main e2e devs machines it doesn't work either + it's a smaller change 🙃

Yes, my intention was already to run non-qemu client for e2e tests.

lluiscampos:
The two failing frontend acceptance tests, `[16/63] 02-layoutAssertions.spec.ts:38:3 › Layout assertions › can authorize a device` and `[26/63] 03-files.spec.ts:184:3 › Files › allows artifact downloads`, also failed in [latest build from `main`](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/8561835874), so this PR is not the culprit.

kjaskiewiczz (approved):

PR #233, opened 2024-11-29
MEN-7657 - device quick commands to trigger sending inventory & trigger deployment check

mzedel:


frontend/src/js/components/devices/troubleshoot/terminal-wrapper.js
```diff
@@ -183,6 +202,21 @@ const TroubleshootContent = ({ device, onDownload, setSocketClosed, setUploadPat
     return close;
   }, [close, sessionState]);
 
+  useEffect(() => {
+    setIsAwaitingCheckInUpdate(false);
+  }, [device.check_in_time]);
+
+  useEffect(() => {
+    setIsAwaitingInventoryUpdate(false);
+    inventoryTimer.current = setTimeout(() => setIsAwaitingInventoryUpdate(false), TIMEOUTS.refreshLong);
```
mineralsfree: Wouldn't local variable (`const timerId = setTimeout(() => setIsAwaitingInventoryUpdate(false), TIMEOUTS.refreshLong);`) and  `return () => clearTimeout(timerId)` do the job? instead of using ref and 2 effects?

mzedel: Ugh, rereading this it looks like I missed a) the timer for the checkin update and setting the timers up shouldn't happen in the effect, but the click handlers 🤦 

Regular local variables aren't stable across rerenders and regular state will trigger rerenders, thus the reliance on refs - and then the effect to clear any timeouts on unmount to prevent causing errors due to state changes on an unmounted component 👌

frontend/src/js/components/devices/troubleshoot/terminal-wrapper.js
```diff
@@ -183,6 +203,21 @@ const TroubleshootContent = ({ device, onDownload, setSocketClosed, setUploadPat
...
+  }, [device.check_in_time]);
+
+  useEffect(() => {
+    setIsAwaitingInventoryUpdate(false);
+  }, [device.updated_ts]);
+
+  useEffect(() => {
+    const currentTimers = timers.current;
+    return () => {
+      Object.values(currentTimers).map(clearTimeout);
```
mineralsfree: `forEach` would semantically fit better here

frontend/src/js/store/devicesSlice/thunks.test.tsx
```diff
@@ -1185,7 +1186,40 @@ describe('troubleshooting related actions', () => {
...
+      { type: actions.receivedDevice.type, payload: { attributes, id } },
+      { type: getDeviceById.fulfilled.type },
+      { type: triggerDeviceUpdate.fulfilled.type }
+    ];
+    // no await here to allow moving beyond the delayed device info update in the next line
+    store.dispatch(triggerDeviceUpdate({ id: defaultState.devices.byId.a1.id, type: 'inventoryUpdate' }));
+    await jest.advanceTimersByTimeAsync(TIMEOUTS.fiveSeconds);
+    const storeActions = store.getActions();
+    expect(storeActions.length).toEqual(expectedActions.length);
+    expectedActions.map((action, index) => expect(storeActions[index]).toMatchObject(action));
```
mineralsfree: `forEach` here and line 1221 as well

mzedel: Good point! adjusted it in all the tests & the rest of the codebase where applicable.

mineralsfree (approved):
:+1: looks good

PR #242, opened 2024-12-02
MEN-7814 - removed initial device limit for newly created tenants

mzedel:
Ticket: MEN-7814
Changelog: None


mineralsfree (approved):
device_limit must be a positive integer, so this value will not be accepted by backend and is invalid from form POV, but if there is no other way, it's ok

mzedel:
@mineralsfree that's the desired behaviour according to the ticket 👌

PR #245, opened 2024-12-03
ci: switch to stable releases

oldgiova:
We are approaching the next hM release, we have to switch the versioning strategy back to stable from the release candidate

Ticket: None
Changelog: None


oldgiova:
Not needed anymore: we switched to `saas` prereleases

PR #246, opened 2024-12-04
Chore/dependency updates

mzedel:
manually updated these, as dependabot seems to error out https://github.com/mendersoftware/mender-server/network/updates/21512785/jobs


kjaskiewiczz (approved):
looks good; did you try to re-trigger dependabot check?

mzedel:
> looks good; did you try to re-trigger dependabot check?

yes and it failed as well - although in this case I suspect it's caused by our internal packages in combination with a major version bump from a peer dependency of this project + the internal package... but I will have to package anyway between now and the next dependabot run, so hopefully that solves that!

PR #253, opened 2024-12-09
MEN-7466-feat: updated billing section in My Organization settings

mineralsfree:


frontend/src/js/common-ui/forms/CountrySelect.tsx
```diff
@@ -28,26 +29,36 @@ const useStyles = makeStyles()(() => ({
...
-    control={control}
-  />
-);
+export const ControlledCountrySelect = ({ control, id, required }) => {
+  const [defaultCountry, setDefaultCountry] = useState('');
+  const { getValues } = useFormContext();
+  useEffect(() => {
+    const values = getValues();
+    setDefaultCountry(values.country);
+  }, [getValues]);
```
mzedel:
```suggestion
  const { getValues } = useFormContext();
  const { country } = getValues();
  useEffect(() => {
    setDefaultCountry(country);
  }, [country]);
```

mzedel: As to the general approach: I'm not quite sure what you're trying to do here... It looks like you want to control the autocomplete via RHF, but I would assume this will reset the autocomplete on change (via the changing key), which results in the changed selection getting rendered, but more by accident

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -14,83 +14,115 @@
...
+      await dispatch(completeUpgrade({ tenantId: (organization as Organization).id, plan: (selectedPlan as Plan).id, billing_profile }));
+      dispatch(setSnackbar(successMessage((selectedPlan as Plan).name)));
+    }
     onCloseClick();
   };
+
   return (
     <Drawer anchor="right" open={true} PaperProps={{ style: { minWidth: '75vw' } }}>
-      <DrawerTitle title={<>Subscribe to {selectedPlan.name}</>} onClose={onCloseClick} />
+      <DrawerTitle title={<>{selectedPlan ? `Subscribe to ${selectedPlan.name}` : 'Edit billing details'}</>} onClose={onCloseClick} />
```
mzedel:
```suggestion
      <DrawerTitle title={<>{selectedPlan ? `Subscribe to Mender ${selectedPlan.name}` : 'Edit billing details'}</>} onClose={onCloseClick} />
```

frontend/src/js/components/settings/organization/organization.js
```diff
@@ -75,6 +83,10 @@ export const Organization = () => {
     dispatch(getUserOrganization());
   }, [dispatch]);
 
+  useEffect(() => {
+    dispatch(getUserBilling());
+  }, [dispatch]);
+
   useEffect(() => {
     dispatch(getSsoConfigs());
   }, [dispatch]);
```
mzedel:
```suggestion
  useEffect(() => {
    dispatch(getUserBilling());
    dispatch(getSsoConfigs());
  }, [dispatch]);
```

frontend/src/js/store/organizationSlice/index.ts
```diff
@@ -87,6 +87,9 @@ export const organizationSlice = createSlice({
     setOrganization: (state, action) => {
       state.organization = { ...state.organization, ...action.payload };
     },
+    setBillingProfile: (state, action) => {
+      state.organization = { ...state.organization, billing_profile: action.payload };
```
mzedel:
```suggestion
      state.organization.billing_profile = action.payload;
```
should be possible thanks to immer

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -240,6 +243,16 @@ export const editTenantDeviceLimit = createAsyncThunk(`${sliceName}/editDeviceLi
...
     });
 });
+export const editBillingProfile = createAsyncThunk(
+  `${sliceName}/editBillingProfileEmail`,
+  ({ billingProfile }: { billingProfile: BillingProfile }, { dispatch }) => {
+    return Api.patch(`${tenantadmApiUrlv2}/billing/profile`, { ...billingProfile })
+      .catch(err => commonErrorHandler(err, `Failed to change billing profile`, dispatch))
+      .then(() => {
+        return Promise.all([dispatch(setSnackbar('Billing Profile was changed successfully')), dispatch(getUserBilling())]);
+      });
```
mzedel:
```suggestion
      .then(() => Promise.all([dispatch(setSnackbar('Billing Profile was changed successfully')), dispatch(getUserBilling())]));
```

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -240,6 +243,16 @@ export const editTenantDeviceLimit = createAsyncThunk(`${sliceName}/editDeviceLi
       ]);
     });
 });
+export const editBillingProfile = createAsyncThunk(
+  `${sliceName}/editBillingProfileEmail`,
+  ({ billingProfile }: { billingProfile: BillingProfile }, { dispatch }) => {
+    return Api.patch(`${tenantadmApiUrlv2}/billing/profile`, { ...billingProfile })
```
mzedel:
```suggestion
    return Api.patch(`${tenantadmApiUrlv2}/billing/profile`, billingProfile)
```
I don't think this needs to be destructured to be sent

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -266,6 +279,11 @@ export const getUserOrganization = createAsyncThunk(`${sliceName}/getUserOrganiz
     return Promise.all(tasks);
   });
 });
+export const getUserBilling = createAsyncThunk(`${sliceName}/getUserBilling`, (_, { dispatch }) => {
+  return Api.get(`${tenantadmApiUrlv2}/billing/profile`).then(res => {
+    dispatch(actions.setBillingProfile(res.data));
+  });
```
mzedel:
```suggestion
  return Api.get(`${tenantadmApiUrlv2}/billing/profile`).then(res => dispatch(actions.setBillingProfile(res.data)));
```

frontend/src/less/main.less
```diff
@@ -300,7 +300,9 @@ each(@directions, .(@direction) {
     grid-template-columns: @defaultLeftBarWidth 1fr;
   }
 }
-
+.align-self-start {
+  align-self: flex-start;
```
mzedel:
```suggestion
  align-self: start;
```
to align with the one below?

frontend/src/js/components/settings/CardSection.tsx
```diff
@@ -47,26 +63,23 @@ const CardSection = ({ isSignUp, organization, onComplete, setSnackbar }) => {
...
   };
 
   const handleSubmit = async event => {
     event.preventDefault();
     setLoading(true);
-    return dispatch(startUpgrade(organization.id))
-      .unwrap()
-      .then(confirmCard)
-      .finally(() => setLoading(false));
+    return (beforeCardSubmit ? beforeCardSubmit() : dispatch(startUpgrade(organization.id)).unwrap()).then(confirmCard).finally(() => setLoading(false));
```
mzedel: maybe this could go into a `onSubmit` prop that would align the two cases (updating card & upgrading the tenant) to both work the same:
```
onSubmit={() => dispatch(startCardUpdate()).unwrap()}
```
and
```
onSubmit={() => dispatch(startUpgrade(organization.id)).unwrap()}
```

frontend/src/js/components/settings/PlanExpanded.tsx
```diff
@@ -14,83 +14,115 @@
...
+interface PlanProps extends PlanExpandedPropsBase {
+  isEdit: false;
+  plan: Plan;
 }
+
+const useStyles = makeStyles()(() => ({
+  submitButtonContainer: {
+    position: 'relative',
+    top: '172px',
+    height: 0
```
mzedel: I assume this is to position the buttons below the "Payment card" section while the payment details are outside of the `Form` - this will break easily + the positioning is probably not going to play nice with tests (`height: 0` might not be considered accessible for an element e.g.) - please adjust this to be part of the form or handle the form actions outside of the form

frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -172,7 +213,46 @@ export const Billing = () => {
...
+            currentBillingProfile={{ address: { ...billing.address }, name: billing.name, email: billing.email }}
+            card={card}
+          />
+        )}
+        <div className={classes.billingSection}>
+          <div className="flexbox center-aligned">
+            <div className={classes.emailTooltip}>
+              <b>Billing details</b>
+            </div>
+            {!isEnterprise && billing && billing.address && (
```
mzedel, line 216: `billing.address` seems to not work with the API response & the type definition shows the data duplication - OTOH maybe the backend response type has changed again 🤷‍♂️

frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -172,7 +213,46 @@ export const Billing = () => {
             </div>
           }
         />
-        {!organization.trial && !isEnterprise && <OrganizationPaymentSettings />}
+        {billing && changeBilling && (
+          <PlanExpanded
+            isEdit
+            onCloseClick={() => setChangeBilling(false)}
+            currentBillingProfile={{ address: { ...billing.address }, name: billing.name, email: billing.email }}
```
mzedel:
```suggestion
            currentBillingProfile={billing}
```

frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -172,7 +213,46 @@ export const Billing = () => {
...
+            <InfoText>
+              Enterprise plan payments are invoiced periodically to your organization. If you have any questions about your billing, <br /> please contact{' '}
+              <a href="mailto:support@mender.io" target="_blank" rel="noopener noreferrer">
+                support@mender.io
+              </a>
+            </InfoText>
+          ) : billing && billing.address ? (
+            <div className="flexbox">
+              <Address address={billing.address} email={billing.email} name={billing.name} />
+              {card && <CardDetails card={card} containerClass={classes.cardDetails} />}
```
mzedel: Sidenote: you could keep these two both free from a passed class with slightly different css.

mzedel (changes requested):
Trying this locally I couldn't use it without changes and the positioning was off for the parts that I got to test.
In this PR again it would have been better to structure your work in multiple commits and align your changes with the rest of the codebase.

frontend/src/js/components/settings/organization/Billing.tsx
mineralsfree: Yes, there there were issues with backend, https://github.com/mendersoftware/mender-server-enterprise/pull/214
It should work with image from this PR

frontend/src/js/store/organizationSlice/thunks.ts
mineralsfree: I prefer braces around function body, so that it's quite easy to debug / add more instructions to the body, but for the sake of consistency with the rest of the codebase, I think we should [enable this eslint rule ](https://eslint.org/docs/latest/rules/arrow-body-style). I will create a follow up ticket

frontend/src/js/common-ui/forms/CountrySelect.tsx
mineralsfree: In my implementation it only remounts the autocomplete once - to set the initial value, which in case of edit form - previously selected country. Maybe not the best workaround, but I didn't find a better way

frontend/src/js/store/organizationSlice/thunks.ts
mzedel: > I prefer braces around function body, so that it's quite easy to debug / add more instructions to the body, but for the sake of consistency with the rest of the codebase, I think we should [enable this eslint rule ](https://eslint.org/docs/latest/rules/arrow-body-style). I will create a follow up ticket

It is unfortunate that until now both internal and external contributions aligned with the existing codebase or took steps to ensure a change is consistent, yet we have to go through the same adjustments repeatedly (e.g. prop passing, arrow functions).
In the past I could have seen the easier debugging, yet today you can debug in arrow functions in most tools. However adding the body just for the sake of potential extensions is in most cases not going to be needed. It would be helpful to have the personal preferences put on the backseat and discussed in a PR to the linting rules or a ticket instead of the noise caused by the introduced inconsistencies.

frontend/src/js/store/organizationSlice/types.ts
```diff
@@ -97,10 +97,17 @@ export interface Addon {
   name: AvailableAddon;
   enabled: boolean;
 }
+export interface Address {
+  country: string;
+  state: string;
+  city: string;
+  line1: string;
+  postal_code: string;
+}
```
mzedel, line 106: this could use the generated `Address` type from the api, but I guess it's enough to note this for the future

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -269,6 +279,9 @@ export const getUserOrganization = createAsyncThunk(`${sliceName}/getUserOrganiz
     return Promise.all(tasks);
   });
 });
+export const getUserBilling = createAsyncThunk(`${sliceName}/getUserBilling`, (_, { dispatch }) =>
```
mzedel, line 282:
```suggestion
export const getUserBillingProfile = createAsyncThunk(`${sliceName}/getUserBilling`, (_, { dispatch }) =>
```
to be aligned with the edit naming..?

mzedel (approved):
sorry for the delay, I only now got around to test this locally - only some optional comments for the future, but I would like to get this in right away to not burden you with another rebase round for other PRs landing soon 😬

PR #257, opened 2024-12-10
MEN-7197-fix: switched to single release endpoint when selecting release

mineralsfree:
Ticket: MEN-7197
Changelog: None


mzedel (changes requested):
Instead of adding the new chunk, please adjust the existing `getRelease` thunk and check that it works as expected where it's used.

mzedel (approved):

PR #263, opened 2024-12-13
fix(gui): enable device configuration for non enterprise users

thall:
Currently it's not possible to see device configuration if you host
Mender self and have environment variable `HAVE_DEVICECONFIG=true`.

Changes the predicate to be the same as for `hasDeviceConnect`.

If I interpret the [add-ons/configure](https://mender.io/pricing/add-ons/configure) page correctly, it should be available in open-source version as well?


mzedel (approved):
Excellent, thank you!

thall:
I can't merge, will you do @mzedel?

PR #264, opened 2024-12-16
fix: stop user from having similar email and password

bahaa-ghazal:
Ticket: MEN-6462

### DIFF
--- a/backend/services/useradm/utils/similar_checker_test.go
+++ b/backend/services/useradm/utils/similar_checker_test.go
@@ -1,0 +1,81 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/mendersoftware/mender-server/services/useradm/model"
+)
+
+func TestSimilarChecker(t *testing.T) {
+	testCases := map[string]struct {
+		inUser   model.User
+		inUpdate model.UserUpdate
+
+		expected bool
+	}{
+		"true": {
+			inUser: model.User{
+				Email: "foo@bar.com",
+			},
+			inUpdate: model.UserUpdate{
+				Password: "foo@bar",
+			},
+			expected: true,
+		},
+		"true, new passowrd and email is similar": {
+			inUser: model.User{
+				Email: "foo@bar.com",
+			},
+			inUpdate: model.UserUpdate{
+				Password: "correcthorsebatterystaple",
+				Email:    "correcthorsebatterystaple@bar.com",
+			},
+			expected: true,
+		},
+		"true, new email similar to current password": {
+			inUser: model.User{
+				Email: "foo@bar.com",
+			},
+			inUpdate: model.UserUpdate{
+				Email:           "correcthorsebatterystaple@bar.com",
+				CurrentPassword: "correcthorsebatterystaple",
+			},
+			expected: true,
+		},
+		"false, passowrd diffrent from email": {
+			inUser: model.User{
+				Email: "foo@bar.com",
+			},
+			inUpdate: model.UserUpdate{
+				Password: "correcthorsebatterystaple",
+			},
+			expected: false,
+		},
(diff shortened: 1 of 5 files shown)


backend/services/useradm/utils/similar_chcker.go
```diff
@@ -0,0 +1,30 @@
...
+	lpass := strings.ToLower(pass)
+	return strings.Contains(lmail, lpass)
+}
+func CheckIfPassSimilarToEmail(user *model.User, userUpdate *model.UserUpdate) bool {
+	if len(userUpdate.Password) > 0 {
+		// check when changing the password
+		mail := string(user.Email)
+		if len(mail) == 0 {
+			mail = string(userUpdate.Email)
+		}
```
alfrunes: This logic should be the opposite. You want to compare with the email in the state after performing the update.
```suggestion
		mail := string(userUpdate.Email)
		if len(mail) == 0 {
			mail = string(user.Email)
		}
```
Even better, you could compare with both the previous and the updated (if any).
```suggestion
		pwd := userUpdate.Password
		// Check if similar to updated email
		if userUpdate.Email != "" && CheckIfPassSimilarToEmailRaw(userUpdate.Email, pwd) {
			return true
		}
		// Check if similar to (old) email
		return CheckIfPassSimilarToEmailRaw(user.Email, pwd)
```

backend/services/useradm/utils/similar_chcker.go
```diff
@@ -0,0 +1,30 @@
...
+		return CheckIfPassSimilarToEmailRaw(mail, string(userUpdate.Password))
+	} else if len(userUpdate.Email) > 0 {
+		// check when changing the email
+		return CheckIfPassSimilarToEmailRaw(string(userUpdate.Email), string(userUpdate.CurrentPassword))
+	}
```
alfrunes: You should only compare the password if it changed. Comparing the current password could possibly prevent you from ever changing the password if it happens to be bad.
```suggestion
	}
```

alfrunes (changes requested):
It would also be great if you could write some unit tests for `utils.CheckIfPassSimilarToEmailRaw`.

backend/services/useradm/utils/similar_chcker.go
```diff
@@ -0,0 +1,30 @@
...
+	if len(userUpdate.Password) > 0 {
+		// check when changing the password
+		mail := string(user.Email)
+		if len(mail) == 0 {
+			mail = string(userUpdate.Email)
+		}
+		return CheckIfPassSimilarToEmailRaw(mail, string(userUpdate.Password))
+	} else if len(userUpdate.Email) > 0 {
+		// check when changing the email
+		return CheckIfPassSimilarToEmailRaw(string(userUpdate.Email), string(userUpdate.CurrentPassword))
```
alfrunes: Side note: This line is too long.

backend/services/useradm/utils/similar_chcker.go
bahaa-ghazal: By removing the "else if" check it will be possible to change your email to the password or similar to the password. If the password and email is already similar it shouldn't be a problem to changing them

alfrunes: Very good observation!
So if the user changes the email, the email is similar to the current password that should also be prevented. However, we need to be careful to not compare the current password with the previous email, that would create the scenario I described above where a user cannot possibly change their password.
Please discard my suggestion above.

backend/services/useradm/utils/similar_checker_test.go
alfrunes, line 1: Great job! Thanks for covering so many test cases :rocket:

alfrunes (approved):
Well done! Congratulations with your first backend contribution :rocket:

PR #265, opened 2024-12-16
test: client compat migrated to backend integration tests and fixes

merlin-northern:
test: client compat migrated to backend integration tests and test separation

Client compatibility test is re-implemented as a backend integration test.

A general mechanism to run tests in a separate and completely independent
session is introduced; all that is needed is to name the test file with
`+separate` before .py extension, for instance moving `test_artifact.py`
to a separate pytest session can be achieved by changing the file name
to `test_artifact+separate.py`.

Ticket: QA-683


merlin-northern:
@kjaskiewiczz let's have a romantic chat about it. there are sone parts that are better spoken about. :)

backend/tests/integration/run
```diff
@@ -1,5 +1,50 @@
...
+        case "$compose_version" in
+            "2."*)  ;;
+            *)
+                echo "Docker Compose v2 is required; installed version: ${compose_version}"
+                rc=1
+                ;;
+        esac
+    fi
+    return $rc
+}
```
alfrunes: Can we simply replace `docker-compose` with `docker compose`?
This is adding unnecessary amount of complexity and the docker compositions will not work with out compose v2 (plugin) anyway.

backend/tests/Dockerfile.integration
```diff
@@ -2,6 +2,7 @@ FROM python:3.13-bullseye
 
 ARG K8S_VERSION=1.29
 
+RUN sed -i.backup -e 's/^deb /deb [trusted=yes] /' /etc/apt/sources.list
```
alfrunes, line 5: What is this for? Does this need a separate commit?

backend/tests/integration/README.md
alfrunes, line 1: Great that you're adding documentation!
Could you please move this into a separate commit?

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,613 @@
...
+    """Fixture setting up a clean (i.e. empty database). Yields
+    common.MongoClient connected to the DB.
+
+    Useful for tests with multiple testcases:
+    - protects the whole test func as usual
+    - but also allows calling MongoClient.cleanup() between cases
+    """
+    mongo_cleanup(mongo)
+    yield mongo
+    mongo_cleanup(mongo)
```
alfrunes: Isn't this just duplicating `clean_mongo`?

backend/tests/integration/tests/test_compat+separate.py
alfrunes, line 1: Most fixtures defined here seem to duplicate entire or partial fixtures defined elsewhere. Can we try to collect at least the common ones in the local `conftest.py` to avoid all this boilerplate?

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,613 @@
...
+            j = j - 1
+
+        for i in range(len(devices_ids)):
+            r = devauthm.with_auth(utoken).call(
+                "DELETE", deviceauth.URL_DEVICE, path_params={"id": devices_ids[i]}
+            )
+            assert r.status_code == 204
+
+        m = MongoClient("mender-mongo:27017")
+        m.cleanup()
```
alfrunes: This is handled by the fixture

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,613 @@
...
+
+def compare_aset(authset, api_authset):
+    assert authset.id == api_authset["id"]
+    assert authset.id_data == api_authset["identity_data"]
+    assert crypto.compare_keys(authset.pubkey, api_authset["pubkey"])
+    assert authset.status == api_authset["status"]
+
+
+@pytest.fixture(scope="function")
+def tenants_with_plans(clean_mongo):
```
alfrunes: This seems unused

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,613 @@
...
+    if per_page is None:
+        per_page = 20
+
+    lo = (page - 1) * per_page
+    hi = lo + per_page
+
+    return devs[lo:hi]
+
+
+def compare_aset(authset, api_authset):
```
alfrunes: unused?

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,613 @@
...
+            tenant_token=tenant_token,
+        )
+        dev.authsets.append(
+            Authset(aset.id, aset.did, dev.id_data, pub, priv, "pending")
+        )
+
+    return dev
+
+
+def filter_and_page_devs(devs, page=None, per_page=None, status=None):
```
alfrunes: unused?

backend/tests/integration/tests/test_compat+separate.py
alfrunes: On second passing, only the `user` fixture is used by `test_compat`, which means that of the non-test code, only `clean_mongo_client` (`clean_mongo`), `clean_migrated_mongo` and `user` is used. The rest seems to be unused.

backend/tests/integration/README.md
```diff
@@ -0,0 +1,41 @@
...
+Further more if you need to have the additional docker-compose files included in your separated run
+you have to define `COMPOSE_FILES_$yourtest_name`, see the example in the `backend/tests/integration/run`,
+for `test_compat+separate.py` the variable is:
+
+```sh
+COMPOSE_FILES_test_compat
+```
+
+where the `test_compat` corresponds to thetest file name without `.py` and `+separate`.
+
```
alfrunes: This is a lot of instructions for adding a test, and it is also adding complexity to the entire structure.
How about doing something like [this](https://docs.pytest.org/en/7.1.x/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs)?
Or, add a pytest argument that select only that test and skip all others and use that in the docker compose override?

merlin-northern: I will do anything to stay away from the pytest mechanisms as they are unportable and really, really obscure and make things _extremely_ confusing. so unless you block it, I will fight for this. all you do is you add "+separate" to the filename, how can this be complicated?

merlin-northern: going "the pytest way" defies the purpose here: with the present patch we have a general mechanism to run any test separately.  all you need to do is add as many test_*+separate.py files and each one of them and out of the box will run completely isolated.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: thank you see above.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: thank you see above.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: thank you see above.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: thank you see above.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: thank you see above.

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: general comment for all the "unused" comments from you: I am sorry it was a left over in the file, I cleaned up everything. thank you for keeping an eye.

backend/tests/integration/README.md
merlin-northern: sure. moved.

backend/tests/Dockerfile.integration
merlin-northern: without it the tests dont run for me on any env including a fresh new debian machine with nothing but docker installed, since it needs signature verified and in order to have it needs to install something from a repo that is not verified, classical. moved to separate commit.

backend/tests/integration/run
merlin-northern: sure. I added a variable in case there is another docker compose in the future (or past :))

backend/tests/integration/run
```diff
@@ -272,6 +277,28 @@ for suite in "${TEST_SUITES[@]}"; do
...
+            script_failed=$run_tests_retcode
+        fi
+
+        if [ $run_tests_retcode -ne 0 ]; then
+            tmppath=$(TMPDIR=./tests mktemp logs.XXXXXX)
+            echo "-- tests failed, dumping logs to $tmppath"
+            compose_cmd logs >> "$tmppath" 2>&1
+        fi
+    done < <(find "${MENDER_SERVER_PATH}/backend/tests/integration/tests" -name '*+separate.py')
+
```
alfrunes: This makes it impossible to parallelize backend integration tests into separate pytest jobs.
Can we:
 * put this in a separate suite (`-s` argument)?
 * put this test in a separate test module and extend the docker-compose.backend-tests-compat.yml to mount this test module under `/tests`?

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,200 @@
...
+            assert r.status_code == 200
+            time.sleep(1)
+            j = j - 1
+            accepted = r.json()
+            if len(accepted) >= len(expected_client_versions):
+                break
+        assert len(accepted) == len(expected_client_versions)
+
+        invm = ApiClient(inventory.URL_MGMT)
+        seen_versions = {}
```
alfrunes:
```suggestion
        seen_versions = set()
```

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,200 @@
...
+                    a
+                    for a in inv_dev["attributes"]
+                    if a["name"] == "mender_client_version"
+                ]
+                if len(mender_client_version_attributes) > 0:
+                    break
+                j = j - 1
+                time.sleep(1)
+            mender_client_version = mender_client_version_attributes[0]["value"]
+            seen_versions[mender_client_version] = True
```
alfrunes:
```suggestion
            seen_versions.add(mender_client_version)
```

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,200 @@
...
+            device_id_to_mender_version[device["id"]] = mender_client_version
+            devices_ids.append(device["id"])
+
+        artifact_file = "tests/data/date.mender"
+        for i in range(len(expected_client_versions)):
+            assert seen_versions[expected_client_versions[i]]
```
alfrunes: This assertion will never fail because it throws a `KeyError` instead.
```suggestion
            assert expected_client_versions[i] in seen_versions
```

backend/tests/integration/tests/test_compat+separate.py
```diff
@@ -0,0 +1,200 @@
...
+            "2.3.2",
+            "2.4.2",
+            "3.5.0",
+            "2.6.1",
+            "3.2.1",
+            "2.0.1",
+            "2.5.4",
+            "2.2.1",
+            "2.1.3",
+            "3.3.2",
```
alfrunes: There are versions here that are not present in the composition: for instance: 4.0.6

alfrunes:
I did this to be able to run the test it in isolation: https://github.com/merlin-northern/mender-server/compare/compat_test_in_backend_integration...alfrunes:mender-server:compat_test_in_backend_integration

merlin-northern:
> I did this to be able to run the test it in isolation: [merlin-northern/mender-server@compat_test_in_backend_integration...alfrunes:mender-server:compat_test_in_backend_integration](https://github.com/merlin-northern/mender-server/compare/compat_test_in_backend_integration...alfrunes:mender-server:compat_test_in_backend_integration)

I see. and if I have a new test I want to run in separation, what do I do?

backend/tests/integration/tests/test_compat+separate.py
merlin-northern: yeah this is tricky, this one is hidded in the docker client tag 3.7

backend/tests/integration/run
merlin-northern: I do not understand. this makes it possible to run any test you put in the test_name+separate.py in a separate session. why do we have to complicate it?

alfrunes:
Sorry, after cleaning up my PR to your fork, I accidentally broke the other test suites. https://github.com/merlin-northern/mender-server/pull/2

merlin-northern:
> Sorry, after cleaning up my PR to your fork, I accidentally broke the other test suites. [merlin-northern#2](https://github.com/merlin-northern/mender-server/pull/2)

do not worry

merlin-northern:
@alfrunes https://github.com/mendersoftware/mender-server-enterprise/pull/255 is the enterprise counter part.

alfrunes (approved):

PR #267, opened 2024-12-17
fix: removed error responder middleware after monorepo migration

mzedel:
[prior PR here](https://github.com/mendersoftware/integration/pull/2642)


oldgiova (approved):
Note to self: to remove port 8080 from the Helm chart.
I'm not so familiar with the json-error-responder, so take my approval with a pinch of salt.

PR #269, opened 2024-12-17
ci: only gui updates registry and repository

oldgiova:
The other services don't have to update the registry and the repository. This to solve the issue with other projects that are relying on the same services other than the gui.

Ticket: ALV-294

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -714,8 +714,10 @@ release:mender-docs-changelog:
           esac
         fi
         THIS_KEY=".${CONTAINER_KEY}.image.tag" THIS_VALUE="${THIS_TAG}" yq -i 'eval(strenv(THIS_KEY)) = strenv(THIS_VALUE)' ${CHART_DIR}/values-${SYNC_ENVIRONMENT}.yaml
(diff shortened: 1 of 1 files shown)


mzedel (commented):
should this not extend to `deviceconnect` and `deviceconfig` as they are both OS?

oldgiova:
> should this not extend to `deviceconnect` and `deviceconfig` as they are both OS?

Right now the automation is working for the `gui` container only (as requested for the e2e tests). Sorry, I didn't got that this was needed for that two repos too.
Deviceconnect in particular is used also for Alvaldi, so in this case we should tune the script even more, to now break Alvaldi 
helm version bump.
Looping in also @alfrunes because we were discussing about this solution

alfrunes:
There was a decision made back in the beginning of October about releasing the backend component to hosted Mender from the closed source monorepo, so I believe this is following what we agreed on back then.

alfrunes (approved):

mzedel:
> There was a decision made back in the beginning of October about releasing the backend component to hosted Mender from the closed source monorepo, so I believe this is following what we agreed on back then.

Ah right, during the discussion to ignore the OS users made by the backend majority, explicitly ignoring the OS users 👌. Good that we have this clarified in the appropriate place then.

PR #271, opened 2024-12-17
MEN-7838 - text input appearance

mzedel:


chiachenglu:
Nice, thank you! @mzedel Is there also a way for me to see the visual changes too? like clicking around to see how it appears in different places in UI, or I can only see after it is merged and see it on [staging.mender.io](https://staging.hosted.mender.io/ui/)?

aleksandrychev (dismissed):

mzedel:
No unfortunately we don't have preview deployments, but you can run it locally using the image built in CI. I'm sure @alfrunes can help you set this up 😉

chiachenglu:
Ok, I ran it locally using the image built. I noticed some inconsistencies on the page. For example, some places failed to update to the new variant or did update to the latest variant but did not update its different states. For example, the default state looks correct, but the focus state still looks like the standard variant. Attached are some of the noticeable inconsistencies I noticed. I am guessing these are built in different classes or something which is why it doesn't get to be updated. Nevertheless, let's try to update as much as possible before we publish the changes to make sure we have a consistent look in the most of the pages and features.  @mzedel 

image (1)
image (3)
image (2)
image (4)
image (5)
image (6)

chiachenglu:
@michaelntech I am considering to also update the [select property](https://mui.com/material-ui/react-text-field/#select) from standard to outlined variant to get the consistent input field for Mender GUI, What do you think? 
Screenshot 2024-12-18 at 11 18 16

michaelntech:
> @michaelntech I am considering to also update the [select property](https://mui.com/material-ui/react-text-field/#select) from standard to outlined variant to get the consistent input field for Mender GUI, What do you think?

Yes, I agree with using outlined everywhere

chiachenglu:
Awesome, @mzedel kindly update the select property for us as well.

chiachenglu (changes requested):
Please adjusted the inconsistent places (see above comments).

mzedel:
> Please adjusted the inconsistent places (see above comments). 

@chiachenglu unfortunately I don't think I will get to it before the end of the sprint 😕
I did get to renew the changes to the codebase changes that rolled in in the meantime. 
There I stumbled across some more inconsistencies when at least rebasing the existing changes, mainly height differences in the release section - should these be aligned to the taller appearance?
And some of the spots you pointed out (email address, not sure what was the other) are disabled/ not editable/ need to be turned editable first, should they also be shown in the bordered style already when disabled?

chiachenglu:
That is no problem. I left the comments so we don't forget to update them. 
Although I am not sure why there are height differences, I guess they are from different sizes of the text input component perhaps (medium and small)? In general, I would expect the text field component to behave and look consistently in its different states (default, hover, focus, disable, and error), whether it is editable or not. So yes, outlined style in disable state too! 👍🏻 

You can view the text field component and how it looks in different states with different properties here: https://www.figma.com/design/e2WPEGusQtx3JnKtOO2san/%F0%9F%8E%B8(UI)-Mender-Design-System-(WIP)?node-id=6570-46740&t=0s73ZMr7eAkjqaen-1

chiachenglu:
Thanks for the fixes, overall looks good! Only found following things that needs to be final adjusted, and to the rest I think we can create the tickets to adjust further (paddings to other elements and sizes for example). 

Screenshot 2025-01-21 at 09 42 47

Screenshot 2025-01-21 at 09 37 11

Screenshot 2025-01-21 at 09 33 15

Screenshot 2025-01-21 at 10 22 33

Screenshot 2025-01-21 at 10 20 32

Screenshot 2025-01-21 at 10 46 28

chiachenglu (dismissed):
Please fix the above points before closing the PR.

mzedel:
There seems to be a steady scope creep happening with every review round 🙃...
 
> Screenshot 2025-01-21 at 09 42 47

Here the theme default is 14px for the outlined input - I'll revert to that instead, if you actually want the divergence, please include it in one of the follow up tasks.a
> Screenshot 2025-01-21 at 09 37 11

> Screenshot 2025-01-21 at 10 22 33

> Screenshot 2025-01-21 at 10 46 28

The alignment here is pre-existing, like the device status select, for future review cycles these things should go along with other follow up adjustments.

chiachenglu:
I agreed some of that we could move to follow-up tasks. Can you share your plan or what needs to be followed up next time? Maybe in the comments. This would help me to review more effectively actually because coming from fresh eyes, I would just blindly expect the UI to look aligned or perfect (at least on the main pages and features) after adaption to the new style. As to the task, did you make any changes or adjustments from the last comments? I tried to fetch the latest changes but I did not see any differences from my local setup. Maybe I miss something? @mzedel

frontend/src/js/components/settings/role-management/PermissionsItems.tsx
```diff
@@ -88,7 +88,7 @@ const ScopeSelect: FunctionComponent<IScopedPermissionSelect> = ({ disabled, per
   const { key, placeholder } = permissionsArea;
   return disabled ? (
     // empty label as a shortcut to align the layout with the select path
```
mineralsfree: remove comment as well?

mzedel:
> I agreed some of that we could move to follow-up tasks. Can you share your plan or what needs to be followed up next time? Maybe in the comments. This would help me to review more effectively actually because coming from fresh eyes, I would just blindly expect the UI to look aligned or perfect (at least on the main pages and features) after adaption to the new style. As to the task, did you make any changes or adjustments from the last comments? I tried to fetch the latest changes but I did not see any differences from my local setup. Maybe I miss something? @mzedel

As to the expectations for what to note in the review: the PR exists in the context of the previous design requirements, so e.g. the role definition/ integrations indentation/ device status selection padding were put there deliberately in the previous designs and thus kept as such. Here I would expect this to be checked against the current state or past designs if unsure and if it's predating the current task and doesn't cause visibility issues (as described in the ticket) => follow up. The user id/ name/ password display e.g. are read only in their current state (user id is always read-only) and as such now diverge from the rest of the UI for read-only values (e.g. device attributes).
With the task at hand in mind, where the description is for now focused on text fields, the review expanded this to other components during the review rounds.

The alignment expectations seem to be a good example as to why "look aligned" or "perfect" is a too subjective matter without a point of reference (as previous design vs. design expectation here seems to diverge - independent of the implementation). I hope that helps to clarify.

As to the content of the PR: Should all be adjusted as requested (except the left padding being 14 instead of 12px to align with the MUI setting there) 👌. It might be required to do a hard reset (e.g. via `git reset mzedel/men-7838 --hard`) - for the future I would be happy to have some more people request feature deployments capabilities to make these PRs easier to review 😄.

mineralsfree (dismissed):
Looks good!

frontend/src/js/components/settings/role-management/PermissionsItems.tsx
mzedel: Argh, thanks for catching that 🙈 !

mineralsfree (approved):
:+1:

PR #273, opened 2024-12-19
fix(gui): removed effect that caused unwanted state modifications

mineralsfree:
Ticket: MEN-7466
Changelog: None


mzedel (dismissed):
The state modifications are indeed wanted to reflect the filter changes in the URL, please create a follow up task to fix this, thanks!

mineralsfree:
> The state modifications are indeed wanted to reflect the filter changes in the URL, please create a follow up task to fix this, thanks!

Changes in the URL ( like startDate and user id)  seem to be reflected still. I can create one just in case though

kjaskiewiczz (approved):

PR #274, opened 2024-12-19
fix: implement signal handler for `server` commands

bahaa-ghazal:
Changelog: Title
Ticket: QA-782

### DIFF
--- a/backend/services/deployments/server.go
+++ b/backend/services/deployments/server.go
@@ -19,10 +19,13 @@ import (
 	"encoding/base64"
 	"net/http"
 	"net/url"
+	"os"
+	"os/signal"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 
 	"github.com/mendersoftware/mender-server/pkg/config"
 	"github.com/mendersoftware/mender-server/pkg/log"
@@ -165,6 +168,7 @@ func SetupObjectStorage(ctx context.Context) (objManager storage.ObjectStorage, 
 
 func RunServer(ctx context.Context) error {
 	c := config.Config
+	l := log.New(log.Ctx{})
 	dbClient, err := mstore.NewMongoClient(ctx, c)
 	if err != nil {
 		return err
@@ -221,5 +225,31 @@ func RunServer(ctx context.Context) error {
 		return http.ListenAndServeTLS(listen, cert, key, handler)
 	}
 
-	return http.ListenAndServe(listen, handler)
+	srv := &http.Server{
+		Addr:    listen,
+		Handler: handler,
+	}
+
+	errChan := make(chan error, 1)
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			errChan <- err
+		}
+	}()
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, unix.SIGINT, unix.SIGTERM)
+	select {
+	case sig := <-quit:
+		l.Infof("received signal %s: terminating", sig)
+	case err = <-errChan:
+		l.Errorf("server terminated unexpectedly: %s", err.Error())
+		return err
+	}
+	l.Info("server shutdown")
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(ctxWithTimeout); err != nil {
+		l.Error("error when shutting down the server ", err)
+	}
+	return nil
 }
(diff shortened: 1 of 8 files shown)


backend/services/deployments/server.go
```diff
@@ -221,5 +225,26 @@ func RunServer(ctx context.Context) error {
...
+	}()
+
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, unix.SIGINT, unix.SIGTERM)
+	<-quit
+
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(ctxWithTimeout); err != nil {
+		l.Fatal("error when shutting down the server ", err)
```
alfrunes: Can we log the error to the `Error` stream and pass the error to the parent function?
```suggestion
		l.Errorf("error when shutting down the server: %s", err.Error())
		return err
```

backend/services/deployments/server.go
```diff
@@ -221,5 +225,26 @@ func RunServer(ctx context.Context) error {
...
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			l.Fatalf("listen: %s\n", err)
+		}
+	}()
+
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, unix.SIGINT, unix.SIGTERM)
+	<-quit
```
alfrunes: If the go routine created here terminates unexpectedly, this process will continue to run doing nothing until a signal is received. We need some way of notifying the "parent" go routine that the "child" is stopping. You could for example achieve this by creating an additional `chan`nel to pass the error from the child and use a [select](https://go.dev/tour/concurrency/5) statement to listen on both `chan`nels.
```suggestion
	errChan := make(chan error, 1)
	go func() {
		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
			errChan <- err
		}
	}()

	quit := make(chan os.Signal, 1)
	signal.Notify(quit, unix.SIGINT, unix.SIGTERM)
	select {
	case sig := <-quit:
		l.Infof("received signal %s: terminating", sig)
	case err = <-errChan:
		l.Errorf("server terminated unexpectedly: %s", err.Error())
		return err
	}
```

alfrunes (changes requested):
This is looking pretty good, your PR is capturing the full coverage report. I only have a couple minor change requests. I didn't want to repeat myself, but I see the change is the same for all 4 files changed so the comments also apply there.

backend/services/deployments/server.go
```diff
@@ -221,5 +225,25 @@ func RunServer(ctx context.Context) error {
...
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			errChan <- err
+		}
+	}()
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, unix.SIGINT, unix.SIGTERM)
+	select {
+	case sig := <-quit:
+		l.Infof("received signal %s: terminating", sig)
```
alfrunes, line 243: I think you accidentally removed the call to `Server.Shutdown`. Can you please bring it back and add it to this `case`?

bahaa-ghazal: I see, that was a stupid  mistake.  I just added `Server.Shutdown` before returning. Thanks for the review!

backend/services/deployments/server.go
```diff
@@ -221,5 +225,31 @@ func RunServer(ctx context.Context) error {
...
+	case err = <-errChan:
+		l.Errorf("server terminated unexpectedly: %s", err.Error())
+		return err
+	}
+	l.Info("server shutdown")
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(ctxWithTimeout); err != nil {
+		l.Fatal("error when shutting down the server ", err)
```
alfrunes: Can you change the log level to `Error` please? :upside_down_face:

alfrunes (commented):
Just one small thing before I click the green button :monocle_face:

alfrunes (approved):

oldgiova:
Just a nitpick suggestion for the next time: by using a `fix:` commit message, this change will be published in the public changelog: https://github.com/mendersoftware/mender-server/pull/309/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR20

Maybe a different commit message for the next time? Like `chore:` or `misc:`, if you don't want this note to go public ;).

But if you really wanted this to be public, please ignore my message

mzedel:
or use `Changelog: None` - since we have the comfort to opt out of the changelog, while using a fitting scope 😉

PR #275, opened 2024-12-19
fix(gui): addressed timestamp filter issue in audit logs

mineralsfree:
Ticket: ME-339
Changelog: None


mzedel (approved):
https://github.com/mendersoftware/mender-server/pull/262 contains related work, but for now this should be ok 👌, maybe add the end of-/ start of day adjustments as seen in nt-gui

PR #285, opened 2024-12-23
ci: Use hard reset instead of checkout for resetting generated changelog

alfrunes:


.gitlab-ci.yml
```diff
@@ -586,7 +586,7 @@ changelog:
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - gh pr checkout --force $RELEASE_PLEASE_PR
     - wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
-    - git checkout ${CI_COMMIT_REF_NAME} -- CHANGELOG.md
+    - git reset --hard -- CHANGELOG.md
```
oldgiova: Does this work on maintenance branches too?

alfrunes: Turns out you cannot do hard reset with pathspec argument, so I changed it to checkout HEAD to do the hard reset.

alfrunes:
This PR won't work since `release-please` is not updating the local git index. Closing...

PR #286, opened 2024-12-23
MEN-7833-feat(gui): switched to pragmatic-drag-and-drop from deprecated react-beautiful-dnd

mineralsfree:
Ticket: MEN-7833
Changelog: None


mzedel (changes requested):
very close 👍, but only the drag indicator is supposed to be draggable and have this reflected in the mouse cursor

mzedel (commented):
With the recent change it's further away from the current implementation and as such the focus on the ordering gets lost a bit IMO - but I'll leave this to @michaelntech to decide:
 
https://github.com/user-attachments/assets/ed01d233-5c62-45bd-b8de-0887a378948d



I would guess the indicator would have to at least be in a brand color.

michaelntech:
> With the recent change it's further away from the current implementation and as such the focus on the ordering gets lost a bit IMO - but I'll leave this to @michaelntech to decide:
> 
> I would guess the indicator would have to at least be in a brand color.

Yes I would suggest #1D8FAF for the horizontal indicator line.
I do miss the "reordering" animation of the current implementation as you mention. MUI has no draggable component but Material Design's dragging elements looks like this: https://m3.material.io/foundations/interaction/gestures#d5f3383f-88f3-4727-9d4c-fb361066be35

Sorry to come late with the suggestion but is _react-dnd_ an option? Seem close to the _react-beautiful-dnd_ utility but is not deprecated: https://react-dnd.github.io/react-dnd/examples/sortable/simple. If not, the proposed PR is fine for now with the color amendment.

mzedel:
speaking of alternatives, there also is: https://github.com/hello-pangea/dnd - supposedly a drop in replacement to beautiful-dnd 🤷

PR #290, opened 2024-12-31
MEN-7851-chore(gui) moved to rspack instead of webpack

mineralsfree:


mzedel (dismissed):
Will the webpack plugins (TsconfigPaths, Html, MiniCss, Copy) still be used by rspack for the respective tasks or can we also remove them?

mzedel (approved):

PR #300, opened 2025-01-02
fix: Deployment device count should not exceed max devices

alfrunes:
Added a condition to skip deployments when the device count reaches max devices.

Changelog: Title
Ticket: MEN-7847

### DIFF
--- a/backend/services/deployments/app/app.go
+++ b/backend/services/deployments/app/app.go
@@ -1404,17 +1404,27 @@ func (d *Deployments) getNewDeploymentForDevice(ctx context.Context,
 	//get deployments newer then last device deployment
 	//iterate over deployments and check if the device is part of the deployment or not
 	var deploy *model.Deployment
-	deploy, err = d.db.FindNewerActiveDeployment(ctx, lastDeployment, deviceID)
-	if err != nil {
-		return nil, nil, errors.Wrap(err, "Failed to search for newer active deployments")
-	}
-	if deploy != nil {
-		deviceDeployment, err := d.createDeviceDeploymentWithStatus(ctx,
-			deviceID, deploy, model.DeviceDeploymentStatusPending)
+	for lastDeployment != nil {
+		deploy, err = d.db.FindNewerActiveDeployment(ctx, lastDeployment, deviceID)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrap(err, "Failed to search for newer active deployments")
 		}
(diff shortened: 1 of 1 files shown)


backend/services/deployments/app/app.go
```diff
@@ -1404,17 +1404,26 @@ func (d *Deployments) getNewDeploymentForDevice(ctx context.Context,
...
+	for deploy == nil {
+		deploy, err = d.db.FindNewerActiveDeployment(ctx, lastDeployment, deviceID)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrap(err, "Failed to search for newer active deployments")
+		}
+		if deploy != nil {
+			if deploy.MaxDevices > 0 &&
+				deploy.DeviceCount != nil &&
+				*deploy.DeviceCount < deploy.MaxDevices {
```
kjaskiewiczz: is that the right condition? I would expect something opposite (*deploy.DeviceCount >= deploy.MaxDevices)

backend/services/deployments/app/app.go
```diff
@@ -1404,17 +1404,26 @@ func (d *Deployments) getNewDeploymentForDevice(ctx context.Context,
...
+			}
+			deviceDeployment, err := d.createDeviceDeploymentWithStatus(ctx,
+				deviceID, deploy, model.DeviceDeploymentStatusPending)
+			if err != nil {
+				return nil, nil, err
+			}
+			return deploy, deviceDeployment, nil
 		}
-		return deploy, deviceDeployment, nil
+		break
```
kjaskiewiczz: do we need a break here?

kjaskiewiczz (changes requested):

backend/services/deployments/app/app.go
alfrunes: You're absolutely right!

backend/services/deployments/app/app.go
alfrunes: I can change it to return instead.

kjaskiewiczz: what I mean is that if we are here, I'm pretty sure deploy == nil and the loop will finish anyway, isn't that the case?

alfrunes: Ohh, right! I forgot about the condition in the for loop :see_no_evil:

kjaskiewiczz (dismissed):
looks good

kjaskiewiczz (approved):
looks good

PR #306, opened 2025-01-06
fix(gui): fixed an issue that would prevent showing deployment reports for phased deployments

mzedel:
FYI @vpodzime


vpodzime (approved):
Looks good to my very limited understanding of the code. :)

PR #308, opened 2025-01-07
MEN-7850-chore(gui) renamed component files to PascalCase and TypeScript extension

mineralsfree:


mzedel (changes requested):
Thank you! Only one small issue: the `OAuth2Providers` in the `Oauth2Providers` file 😉

mzedel (approved):

PR #326, opened 2025-01-08
Stable release 4.0.0

oldgiova:
Release-As: 4.0.0

Note: merge this only when ready to release

Requirements:
* [ ] All the relevant PRs are merged
* [ ] Manual test on on-prem Opensource
* [ ] Manual test on on-prem Enterprise


alfrunes (commented):
Could you also update the release-please config since it's configured as prerelease?
There's also #245

oldgiova:
> Could you also update the release-please config since it's configured as prerelease? There's also #245

I could, but still this is subjected to review during the Mender Planning next week, for how to make releases on hosted Mender... Does it make sense to wait for it before merging this down?

Additionally, this commit is required even if we switch now to stable release only in release-please: it would read anyway the previous release as rc, and it would continue to work on the next rc (already verified)

kjaskiewiczz (approved):

alfrunes (approved):

PR #328, opened 2025-01-09
MEN-7858 - feat(gui): added feedback on file size limits to artifact upload dialog

mzedel:
Ticket: MEN-7858
Changelog: Title

### DIFF
--- a/frontend/src/js/common-ui/InputErrorNotification.tsx
+++ b/frontend/src/js/common-ui/InputErrorNotification.tsx
@@ -17,7 +17,7 @@ import { TIMEOUTS } from '@northern.tech/store/commonConstants';
 
 export const InputErrorNotification = ({ className, content }: { className: string; content: string }) => {
   const [isVisible, setIsVisible] = useState(false);
-  const timer = useRef<NodeJS.Timeout | undefined>();
+  const timer = useRef<ReturnType<typeof setTimeout> | undefined>();
 
   useEffect(() => {
     if (!content) {
--- a/frontend/src/js/common-ui/InputErrorNotification.tsx
+++ b/frontend/src/js/common-ui/InputErrorNotification.tsx
@@ -1,0 +1,32 @@
+// Copyright 2025 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+import React, { useEffect, useRef, useState } from 'react';
+
+import { TIMEOUTS } from '@northern.tech/store/commonConstants';
+
+export const InputErrorNotification = ({ className, content }: { className: string; content: string }) => {
+  const [isVisible, setIsVisible] = useState(false);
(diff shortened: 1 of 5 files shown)


frontend/src/js/components/releases/dialogs/AddArtifact.tsx
```diff
@@ -67,29 +90,48 @@ const shortenFileName = name => {
   return name;
 };
 
-export const ArtifactUpload = ({ setSnackbar, updateCreation }) => {
+const singleFileLimit = 536870912; //512MiB
+const menderFileLimit = 10737418240; //10GiB
+const reFilename = new RegExp(/^[a-z0-9.,_-]+$/i);
```
mineralsfree: is it only lowercase letters that are allowed in filename?

frontend/src/js/common-ui/InputErrorNotification.tsx
```diff
@@ -0,0 +1,36 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+import React, { useEffect, useRef, useState } from 'react';
+
+import { TIMEOUTS } from '@northern.tech/store/commonConstants';
+
+export const InputErrorNotification = ({ className, content }: { className: string; content: string }) => {
+  const [isVisible, setIsVisible] = useState(false);
+  const timer = useRef<NodeJS.Timeout | undefined>();
```
mineralsfree: `ReturnType<typeof setTimeout>` might be a better alternative

frontend/src/js/common-ui/InputErrorNotification.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    if (!content) {
+      return;
+    }
+    setIsVisible(true);
+    console.log('set visible');
```
mineralsfree: please, remove console.logs

frontend/src/js/components/releases/dialogs/AddArtifact.tsx
```diff
@@ -67,29 +90,48 @@ const shortenFileName = name => {
   return name;
 };
 
-export const ArtifactUpload = ({ setSnackbar, updateCreation }) => {
+const singleFileLimit = 536870912; //512MiB
```
mineralsfree: maybe it's better to move those consts in const file and write them as `512 * (1024 ** 2)` and `10 * (1024 ** 3)`?

mineralsfree (changes requested):

frontend/src/js/components/releases/dialogs/AddArtifact.tsx
mzedel: ha nice find, that's been wrong since introduction then... I can't imagine uppercase not being supported in the backend/ client and [the mender artifact regex supports it](https://github.com/mendersoftware/mender-artifact/blob/master/awriter/writer.go#L619)

frontend/src/js/common-ui/InputErrorNotification.tsx
mzedel: whoops!

frontend/src/js/components/releases/dialogs/AddArtifact.tsx
mzedel: Since this is very isolated I wouldn't want to move it anywhere else until it gets reused in a different context.

frontend/src/js/components/releases/dialogs/AddArtifact.tsx
mzedel: ah, nope it was case independent already, but still good to have it aligned with Mender Artifact 👍

mineralsfree (approved):
:+1:

PR #330, opened 2025-01-09
MEN-7833-chore(gui): switched to supported library for dnd functionality

mineralsfree:
Ticket: MEN-7833
Changelog: None


mineralsfree:
#286 should be rejected with this being merged

mzedel (approved):

PR #333, opened 2025-01-13
test: Remove `mock-workflows` from deployments acceptance tests

alfrunes:


backend/services/deployments/tests/tests/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def _handle_request(self):
+        self.send_response(200)
+        self.end_headers()
+
+    def __getattribute__(self, name: str, /) -> Any:
+        if name.startswith("do_"):
+            return self._handle_request
+        return super().__getattribute__(name)
```
merlin-northern: could you leave some comment about this method, I have no idea why it is needed here.

merlin-northern (dismissed):

backend/services/deployments/tests/tests/conftest.py
alfrunes: I added a docstring for the method.

alfrunes:
I added a docstring and removed the dependabot tracking for the service, and also discovering that we have some dangling `.github` directories from the old repos still left over that I cleaned up :broom:

mzedel (approved):

PR #348, opened 2025-01-18
Split CHANGELOG files and saas prerelease tag

oldgiova:
1. Set prerelease tags as `saas`
2. Split CHANGELOG files between CHANGELOG.md (Open source), CHANGELOG-saas.md (hosted Mender), CHANGELOG-enterprise (Enterprise)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -582,6 +582,10 @@ changelog:
     # GITHUB_TOKEN for Github cli authentication
     - export GITHUB_TOKEN=${GITHUB_CLI_TOKEN}
     - cp CHANGELOG.md CHANGELOG.md.${CI_COMMIT_REF_NAME} # Saving the changelog for later
+    - cp CHANGELOG.md CHANGELOG.md.${CI_COMMIT_REF_NAME} || true # Saving the changelog for later
+    - cp CHANGELOG-saas.md CHANGELOG-saas.md.${CI_COMMIT_REF_NAME} || true # Saving the changelog for later
+    - cp CHANGELOG-enterprise.md CHANGELOG-enterprise.md.${CI_COMMIT_REF_NAME} || true # Saving the changelog for later
+    - export CHANGELOG_SUFFIX=""
   script:
     - release-please release-pr
       --token=${GITHUB_BOT_TOKEN_REPO_FULL}
@@ -595,14 +599,23 @@ changelog:
     - test -z "$RELEASE_PLEASE_PR" && echo "No release-please PR found" && exit 0
     - gh pr checkout --force $RELEASE_PLEASE_PR
     - wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
-    - mv CHANGELOG.md.${CI_COMMIT_REF_NAME} CHANGELOG.md
-    - git cliff
-      --unreleased
-      --prepend CHANGELOG.md
-      --github-repo ${GITHUB_REPO_URL}
-      --use-branch-tags
-      --tag $(jq -r '.["."]' .release-please-manifest.json)
-    - git add CHANGELOG.md
+    - RELEASE_VERSION="$(jq -r '.["."]' .release-please-manifest.json)"
+    - |
+      case $RELEASE_VERSION in
+        *saas*)
+          if [[ "$CI_PROJECT_NAME" == "mender-server-enterprise" ]]; then
+            ./.gitlab/generate_changelog.sh "${RELEASE_VERSION}" "-saas" "${GITHUB_REPO_URL}" "${CI_COMMIT_REF_NAME}"
+          fi
+          ;;
+        *)
+          if [[ "$CI_PROJECT_NAME" == "mender-server-enterprise" ]]; then
+            ./.gitlab/generate_changelog.sh "${RELEASE_VERSION}" "-enterprise" "${GITHUB_REPO_URL}" "${CI_COMMIT_REF_NAME}"
+            ./.gitlab/generate_changelog.sh "${RELEASE_VERSION}" "-saas" "${GITHUB_REPO_URL}" "${CI_COMMIT_REF_NAME}"
+          else
+            ./.gitlab/generate_changelog.sh "${RELEASE_VERSION}" "" "${GITHUB_REPO_URL}" "${CI_COMMIT_REF_NAME}"
+          fi
+          ;;
+      esac
     - git commit --amend -s --no-edit
     - git push github-${CI_JOB_ID} --force
     # Update the PR body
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -582,6 +582,10 @@ changelog:
     # GITHUB_TOKEN for Github cli authentication
     - export GITHUB_TOKEN=${GITHUB_CLI_TOKEN}
     - cp CHANGELOG.md CHANGELOG.md.${CI_COMMIT_REF_NAME} # Saving the changelog for later
+    - THIS_REPO=$(echo ${GITHUB_REPO_URL} | cut -d '/' -f2)
```
alfrunes: There is a gitlab variable called `$CI_PROJECT_NAME` storing exactly this information :)

.gitlab-ci.yml
```diff
@@ -647,17 +668,39 @@ release:mender-docs-changelog:
...
+    - gh pr create --title "Update CHANGELOG${CHANGELOG_SUFFIX}.md for $CI_PROJECT_NAME" --body "Automated change to the CHANGELOG${CHANGELOG_SUFFIX}.md file" --base master --head changelog-${CI_JOB_ID}
+
+release:mender-docs-changelog:saas:
+  extends: release:mender-docs-changelog
+  variables:
+    CHANGELOG_REMOTE_FILE: "12.Hosted-Mender/docs.md"
+  rules:
+    - if: '$CI_PROJECT_NAME == "mender-server"'
+      when: never
+    - if: $CI_COMMIT_TAG
```
alfrunes:
```suggestion
    - if: $CI_COMMIT_TAG =~ "^v\d+\.\d+\.\d+-saas[\.0-9a-z+]+$"
```

.gitlab-ci.yml
```diff
@@ -647,17 +668,39 @@ release:mender-docs-changelog:
     - git config --global user.name "${GITHUB_USER_NAME}"
     # GITHUB_TOKEN for Github cli authentication
     - export GITHUB_TOKEN=${GITHUB_CLI_TOKEN}
+    - THIS_REPO=$(echo ${GITHUB_REPO_URL} | cut -d '/' -f2)
```
alfrunes: Replace with `CI_PROJECT_NAME`?

.gitlab-ci.yml
```diff
@@ -595,14 +599,31 @@ changelog:
...
-      --github-repo ${GITHUB_REPO_URL}
-      --use-branch-tags
-      --tag $(jq -r '.["."]' .release-please-manifest.json)
-    - git add CHANGELOG.md
+    - if [[ "${THIS_REPO}" == "mender-server-enterprise" ]]; then
+        if [[ $(grep --count "saas" .release-please-manifest.json) -eq 1 ]]; then
+          echo "INFO - running CHANGELOG for saas only";
+          CHANGELOG_SUFFIX="-saas";
+          mv CHANGELOG${CHANGELOG_SUFFIX}.md.${CI_COMMIT_REF_NAME} CHANGELOG${CHANGELOG_SUFFIX}.md;
+          git cliff --unreleased --prepend CHANGELOG${CHANGELOG_SUFFIX}.md --github-repo ${GITHUB_REPO_URL} --use-branch-tags --tag $(jq -r '.["."]' .release-please-manifest.json);
```
alfrunes: Can we reuse this `jq` expression for the version? I think it's better than using the grep expression of the entire manifest as this would conflict if we want to move other components to this repo.

.gitlab-ci.yml
```diff
@@ -595,14 +599,31 @@ changelog:
...
+          git cliff --unreleased --prepend CHANGELOG${CHANGELOG_SUFFIX}.md --github-repo ${GITHUB_REPO_URL} --use-branch-tags --tag $(jq -r '.["."]' .release-please-manifest.json) --ignore-tags saas;
+          git add CHANGELOG${CHANGELOG_SUFFIX}.md;
+        fi;
+      else
+        echo "INFO - running CHANGELOG for non-enterprise";
+        export CHANGELOG_SUFFIX="";
+        mv CHANGELOG${CHANGELOG_SUFFIX}.md.${CI_COMMIT_REF_NAME} CHANGELOG${CHANGELOG_SUFFIX}.md;
+        git cliff --unreleased --prepend CHANGELOG${CHANGELOG_SUFFIX}.md --github-repo ${GITHUB_REPO_URL} --use-branch-tags --tag $(jq -r '.["."]' .release-please-manifest.json) --ignore-tags saas;
+        git add CHANGELOG${CHANGELOG_SUFFIX}.md;
+      fi;
```
alfrunes: There's also a lot of duplication in the different path. How about refactoring it to something like this?
```suggestion
    - RELEASE_VERSION="$(jq -r '.["."]' .release-please-manifest.json)"
    - case $RELEASE_VERSION in
    - *saas*) CHANGELOG_SUFFIX="-saas";;
    - *) test $CI_PROJECT_NAME = "mender-server-enterprise" && CHANGELOG_SUFFIX="-enterprise";;
    - esac
    - CHANGELOG_FILE="CHANGELOG${CHANGELOG_SUFFIX:-}.md"
    - echo "Generating changelog file ${CHANGELOG_FILE} for release ${RELEASE_VERSION}"
    - mv "${CHANGELOG_FILE}.${CI_COMMIT_REF_NAME}" "$CHANGELOG_FILE";
    -  git cliff --unreleased --prepend CHANGELOG${CHANGELOG_SUFFIX}.md --github-repo ${GITHUB_REPO_URL} --use-branch-tags --tag "$RELEASE_VERSION";
    - git add "$CHANGELOG_FILE"
```

.gitlab-ci.yml
oldgiova: On this point: what we want to do when we publish the Stable on-prem tags? For example: v4.1.0, which contains the same code as v4.1.0-saas.rc1 for example. We want to publish it to hosted Mender also or not?

.gitlab-ci.yml
oldgiova: I think I don't got this.. Isn't the expression already using the `jq` command?

.gitlab-ci.yml
alfrunes: You're right! Maybe we should add a restriction that has a version tag prefix?
Just in case we would create other tags for whatever reason.

.gitlab-ci.yml
alfrunes: I meant this line here: https://github.com/mendersoftware/mender-server/blob/7b5b82936bb30034d0ad203ab9c724f926868453/.gitlab-ci.yml#L603

.gitlab-ci.yml
oldgiova: That's why I love when you review my ugly work! Yes, I love this one. 
But, please note that to make CHANGELOG-enterprise works properly, you have to set the `--ignore-tags saas` argument. I'll check it later

.gitlab-ci.yml
oldgiova: Yes, I'm going to reuse your suggested change and include `saas` as optional.

.gitlab-ci.yml
oldgiova: Additionally, we want to deploy run the `saas` changelog on every change (also for stable tags), while running the enterprise changelog only on stable tags, so the case `*)` cannot work this way

oldgiova: That's the best I could do

alfrunes: If I understand it correctly we have 4 cases:
 * `mender-server-enterprise/v4.1.0-saas.123`: Updates only `CHANGELOG-saas.md`
 * `mender-server-enterprise/v4.1.0`: Would update both `CHANGELOG-enterprise.md` and `CHANGELOG-saas.md`
 * `mender-server/v4.1.0`: Would update `CHANGELOG.md`
 * `mender-server/v4.1.0-saas.123`: Do not update anything. 
Then, would something like this work?

```suggestion
    - if test $CI_PROJECT_NAME = "mender-server-enterprise"; then
    - case $RELEASE_VERSION in
    - *saas*) CHANGELOG_FILES="CHANGELOG-saas.md";;
    - *) CHANGELOG_FILES="CHANGELOG-enterprise.md CHANGELOG-saas.md"
    - ;;
    - esac
    - else
    - case $RELEASE_VERSION in
    - *saas*) CHANGELOG_FILES="";; # Do nothing
    - *) CHANGELOG_FILES="CHANGELOG.md";;
    - esac
    - fi
    - for CHANGELOG_FILE in $CHANGELOG_FILES; do
    - echo "Generating changelog file ${CHANGELOG_FILE} for release ${RELEASE_VERSION}"
    - mv "${CHANGELOG_FILE}.${CI_COMMIT_REF_NAME}" "$CHANGELOG_FILE";
    - git cliff --unreleased --prepend CHANGELOG${CHANGELOG_SUFFIX}.md --github-repo ${GITHUB_REPO_URL} --use-branch-tags --tag "$RELEASE_VERSION";
    - git add "$CHANGELOG_FILE"
    - done
```
Or maybe it would make sense to put this in it's own script? The formatting is starting to look a bit unreadable :man_shrugging:

oldgiova: Yeah, let's write an helper script

oldgiova: Done, let's try it now

alfrunes (approved):

PR #349, opened 2025-01-20
docs: Remove unnecessary step override for Mender Server Enterprise

lluiscampos:
In `mender-server-enterprise`, the variable is already set up correctly for the user in the `.env` file, so this extra step is not necessary.


mzedel (changes requested):
As a consumer of the enterprise setup I will have access to the images, but not the repository - so the steps here are relevant for Enterprise users, whereas the ones in enterprise likely not so much...

lluiscampos:
> As a consumer of the enterprise setup I will have access to the images, but not the repository - so the steps here are relevant for Enterprise users, whereas the ones in enterprise likely not so much...

Good point! Thank you.

PR #365, opened 2025-01-23
ci: avoid running same e2e from multiple repos

oldgiova:
When two CICD jobs have the same resource_group, only one at a time takes place; this avoids running the same e2e tests from both the mender-server and mender-server-enterprise

Ticket: None
Changelog: None


mzedel (approved):
🙏 thank you - I wasn't aware of this functionality!

PR #375, opened 2025-01-24
Fix/webkit some more

mzedel:
webkit job, down to 1 failure https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/8959767143 - which is still unexpected, but I would wait for https://github.com/mendersoftware/mender-server/pull/375/commits/c1d181eb1ed4d0b882e78dc959e3be2b560d5beb to land on staging and see if that makes a difference there 🤞


kjaskiewiczz (approved):
thanks!

PR #376, opened 2025-01-24
fix: Internally signed AWS URLs manipulating global storage settings

alfrunes:


kjaskiewiczz (approved):
looks good;
I also checked locally and it's working fine (I was able to update and download artifact)

PR #379, opened 2025-01-24
ME-426 - fix(gui): fixed an issue that could prevent listing devices with their custom identity in a deployment report

mzedel:
Ticket: ME-426
Changelog: Title

### DIFF
--- a/frontend/src/js/common-ui/DeviceIdentity.tsx
+++ b/frontend/src/js/common-ui/DeviceIdentity.tsx
@@ -64,7 +64,7 @@ export const getDeviceIdentityText = ({ device = {}, idAttribute }) => {
   const { attribute, scope } = idAttribute;
   // special handling for tags purely to handle the untagged devices case
   if (attribute === 'name' && scope === 'tags') {
-    return tags[idAttribute] ?? `${id.substring(0, 6)}...`;
+    return tags[attribute] ?? `${id.substring(0, 6)}...`;
   }
   return defaultTextRender({ column: { attribute: { name: attribute, scope } }, device });
 };
--- a/frontend/src/js/common-ui/DeviceIdentity.tsx
+++ b/frontend/src/js/common-ui/DeviceIdentity.tsx
@@ -94,8 +94,8 @@ export const DeviceIdentityDisplay = props => {
   const { classes } = useStyles();
 
   let Component = attributeComponentMap.default;
-  if (isEditable && attribute === 'name' && scope === 'tags') {
-    Component = attributeComponentMap[attribute] ?? attributeComponentMap.default;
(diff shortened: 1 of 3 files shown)


frontend/src/js/common-ui/DeviceIdentity.tsx
```diff
@@ -64,7 +64,7 @@ export const getDeviceIdentityText = ({ device = {}, idAttribute }) => {
   const { attribute, scope } = idAttribute;
   // special handling for tags purely to handle the untagged devices case
   if (attribute === 'name' && scope === 'tags') {
-    return tags[idAttribute] ?? `${id.substring(0, 6)}...`;
+    return tags[attribute] ?? `${id.substring(0, 6)}...`;
```
mineralsfree, line 67: Maybe it's better to fallback to `-` so it's less misleading?

frontend/src/js/components/deployments/deployment-report/DeviceList.tsx
```diff
@@ -72,8 +72,8 @@ const deviceListColumns = [
     title: 'id',
     renderTitle: ({ idAttribute }) => idAttribute.attribute,
     render: ({ device, idAttribute }) => (
-      <Link style={{ fontWeight: 'initial', opacity: idAttribute.attribute === 'name' ? 0.6 : 1 }} to={`/devices?id=${device.id}`}>
-        <DeviceIdentityDisplay device={device} isEditable={false} />
+      <Link style={{ fontWeight: 'initial' }} to={`/devices?id=${device.id}`}>
+        {getDeviceIdentityText({ device, idAttribute })}
```
mineralsfree: was DeviceIdentityDisplay the issue?

mineralsfree (dismissed):

frontend/src/js/components/deployments/deployment-report/DeviceList.tsx
mzedel: Not necessarily 🙈 - the combination of this also having to work with the name tag and not being editable - there's an exemption in the DeviceIdentityDisplay for the name tag - and I just took the shortcut of showing the rendered content instead of fixing the DeviceIdentityDisplay (which is used in several other places)... I'll fix that instead!

frontend/src/js/common-ui/DeviceIdentity.tsx
mzedel: that's according to the design (to have at least a reference to the device id in a device list if there's no name configured), unfortunately the old design with the definition for this is no longer accessible, but https://northerntech.atlassian.net/browse/MEN-4157?focusedCommentId=59511 still has _some_ reference to this 👌

mineralsfree (approved):

PR #386, opened 2025-01-28
MEN-8010 - fix(gui): ensured target directory is nonempty on artifact generation

mzedel:
Ticket: MEN-8010
Changelog: Title


mineralsfree (approved):
Good  enough IMO

PR #389, opened 2025-01-29
Fix changelog generation

oldgiova:
Reference: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/8979426766#L175


alfrunes (approved):
Ouch! Those quotation marks should not be there :face_with_peeking_eye:  `ls(1)` is also fine.

PR #398, opened 2025-01-30
ci: Add license check for third-party services

alfrunes:


kjaskiewiczz (commented):
Do I understand correctly that the hack for license not being present in the parent directory is now covered by check with unknown in disallowed types?

alfrunes:
> Do I understand correctly that the hack for license not being present in the parent directory is now covered by check with unknown in disallowed types?

No, the hack no longer applies because of the ignore argument.

kjaskiewiczz (approved):
nice!

PR #403, opened 2025-01-31
test: webhook: test for device-auth events.

merlin-northern:
Ticket: QA-865

### DIFF
--- a/backend/tests/integration/docker/docker-compose.backend-tests.yml
+++ b/backend/tests/integration/docker/docker-compose.backend-tests.yml
@@ -60,6 +60,8 @@ services:
     user: ${UID:-0}:${GID:-0}
     volumes:
       - ${GOCOVERDIR:-cover}:/cover
+  mock-httpd:
+    image: mockserver/mockserver:5.15.0
 
 volumes:
   cover: {}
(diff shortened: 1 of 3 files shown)


kjaskiewiczz:
If I'm not wrong, the test covers the case with webhook configured to receive both, deviceauth and inventory events. That's fine, but to test filtering of events by webhook scope I'd add tests with webhook configured to receive inventory events only (maybe we already have this one?) and another webhook with deviceauth events only.
@merlin-northern what do you think?
EDIT: sorry, the comment should be here - https://github.com/mendersoftware/mender-server-enterprise/pull/309

kjaskiewiczz (approved):
looks good

merlin-northern:
> If I'm not wrong, the test covers the case with webhook configured to receive both, deviceauth and inventory events. That's fine, but to test filtering of events by webhook scope I'd add tests with webhook configured to receive inventory events only (maybe we already have this one?) and another webhook with deviceauth events only. @merlin-northern what do you think? EDIT: sorry, the comment should be here - [mendersoftware/mender-server-enterprise#309](https://github.com/mendersoftware/mender-server-enterprise/pull/309)

do not worry. thanks.

PR #411, opened 2025-02-03
ME-434 - fix(gui): fixed an issue that would cause the ui to crash when creating phased deployments

mzedel:
phase information is tracked in the store/ covered by immer, so mutating state caused issues


mineralsfree (dismissed):

mineralsfree (changes requested):
Please, fix the unit test. One of the commit messages header is too long

mineralsfree (approved):

PR #412, opened 2025-02-03
MEN-7728-fix(gui): prevented role creation dialog from closing in case of error

mineralsfree:
Ticket: MEN-7728
Changelog: None

### DIFF
--- a/frontend/src/js/store/usersSlice/thunks.ts
+++ b/frontend/src/js/store/usersSlice/thunks.ts
@@ -610,7 +610,14 @@ const roleActions = {
   }
 };
 
-const roleActionErrorHandler = (err, type, dispatch) => commonErrorHandler(err, `There was an error ${roleActions[type].errorMessage} the role.`, dispatch);
+const roleActionErrorHandler = (err, type, dispatch, meta) => {
+  const { permissionSetsCreated, name } = meta;
+  let errorContext = `There was an error ${roleActions[type].errorMessage} the role.`;
+  if (permissionSetsCreated) {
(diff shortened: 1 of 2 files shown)


mzedel (changes requested):
Looks good so far, but doesn't quite solve the issue yet. The task wants to have the number of permission sets in the error message if there's too many created and preferably also an indication of how many permission sets the current selection will result in.
I don't think the latter will be 100% useful as the limit can be set per tenant and the UI has no way of knowing what the limit is, but considering the 15 sets default limit, a notification once that number is reached should be possible.
The error message part has to be done in here - the notification part might otherwise come back in a separate task sooner or later.

frontend/src/js/store/usersSlice/thunks.ts
```diff
@@ -610,7 +610,14 @@ const roleActions = {
   }
 };
 
-const roleActionErrorHandler = (err, type, dispatch) => commonErrorHandler(err, `There was an error ${roleActions[type].errorMessage} the role.`, dispatch);
+const roleActionErrorHandler = (err, type, dispatch, meta) => {
+  const { permissionSetsCreated, name } = meta;
+  let errorContext = `There was an error ${roleActions[type].errorMessage} the role.`;
+  if (permissionSetsCreated) {
+    errorContext += ` Tried to create role ${name} with ${permissionSetsCreated} permission sets.`;
```
mzedel:
```suggestion
    errorContext += ` Tried to ${type} role ${name} with ${permissionSetsCreated} permission sets.`;
```

mzedel (changes requested):
Just a tiny bit to make it work with the editing too 👌

mzedel (approved):

PR #420, opened 2025-02-07
ci(gui): adjusted the way pending devices are seen in UI after dependency updates

mzedel:
looks like the stress-test-runner wasn't at fault, but more likely the playwright update that changed the style in browser locators work 🙈


kjaskiewiczz (approved):
thanks!

PR #422, opened 2025-02-07
ci: Add gosec to backend static checks

alfrunes:


backend/.golangci.yml
```diff
@@ -48,3 +49,135 @@ linters-settings:
     parameters-are-used: true
     local-variables-are-used: true
     generated-is-used: false
+
+  gosec:
```
lluiscampos: @alfrunes This is very hard to review - could you highlight which are the deviations from the default configuration for `gosec`? Even better if we add it here in a comment, for future maintenance.

alfrunes: I updated the commit message. It is more or less the defaults, but I removed the rules we already satisfy from the exclusion list. I compacted the configuration, but left the comments explaining what each config does.

lluiscampos (approved):

PR #441, opened 2025-02-13
test: fix failing test

oldgiova:
Template updated with the snapshot expectation

Ticket: QA-918


oldgiova:
Reference:
* failing job: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9124067214
* fix: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9125139495

mzedel (commented):
passing job with current snapshot: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9112658377 - there's not that much to gain from the alignment until the ids get calculated stable, we can remove the ids from the snapshot (as we have in some other tests), but I'd rather wait for https://github.com/mendersoftware/mender-server/pull/431 to land and see how that behaves...

oldgiova:
> passing job with current snapshot: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9112658377 - there's not that much to gain from the alignment until the ids get calculated stable, we can remove the ids from the snapshot (as we have in some other tests), but I'd rather wait for #431 to land and see how that behaves...

Let's close this then

PR #444, opened 2025-02-14
Revert "revert: Revert "docs(deviceauth): migration to OpenAPI3""

merlin-northern:
This reverts commit 93ab08ab6051aec3508bb550a4455d30ba2a9b56.


kjaskiewiczz (approved):
I didn't go through it, since I did that already in the past.
If something will go wrong, we can always revert :D

PR #457, opened 2025-02-18
MEN-8109 - fix(gui): fixed an issue that prevented creating a deployment from a single release

mzedel:


frontend/src/js/store/releasesSlice/index.ts
```diff
@@ -15,9 +15,38 @@
 import { DEVICE_LIST_DEFAULTS, SORTING_OPTIONS } from '@northern.tech/store/constants';
 import { createSlice } from '@reduxjs/toolkit';
 
+import { Release } from '../api/types/Release';
+
 export const sliceName = 'releases';
 
-export const initialState = {
+interface SortOptions {
+  direction?: string;
```
mineralsfree: I think it should be narrowed to possible directions

frontend/src/js/store/releasesSlice/index.ts
```diff
@@ -15,9 +15,38 @@
...
+    perPage: number;
+    releaseIds: string[];
+    searchedIds: string[];
+    searchTerm: string;
+    searchTotal: number;
+    selectedTags: [];
+    selection: number[];
+    sort?: SortOptions;
+    total: 0;
+    type: string;
```
mineralsfree, line 39: should it also be narrowed?

frontend/src/js/store/releasesSlice/index.ts
```diff
@@ -15,9 +15,38 @@
...
+  byId: Record<string, Release>;
+  releasesList: {
+    isLoading?: boolean;
+    page: number;
+    perPage: number;
+    releaseIds: string[];
+    searchedIds: string[];
+    searchTerm: string;
+    searchTotal: number;
+    selectedTags: [];
```
mineralsfree: I think it should be any[], not an empty array

mineralsfree (changes requested):

frontend/src/js/store/releasesSlice/index.ts
mzedel: right... I reused one of the existing type definitions for the same in the codebase, but in https://github.com/mendersoftware/mender-server/pull/460/commits/246e5ba5d8608db264a317a8a390ca6a6f12c32e I made this dependent on the directions... so once that PR is in, I'll rebase this and use the shared type in hear - would that be ok?

later I would put this in the `types` package we talked about (to decouple this from the store)

mineralsfree (approved):

PR #462, opened 2025-02-19
docs: Migrate deployments specs to OpenAPI 3.0

alfrunes:
The openapi specs were generated by using `openapi-generator`.
```
openapi-generator generate -g openapi-yaml ...
```

### DIFF
--- a/backend/docs/api/internal_v1_deployments.yaml
+++ b/backend/docs/api/internal_v1_deployments.yaml
@@ -1,0 +1,1124 @@
+openapi: 3.0.1
+info:
+  description: |
+    Internal API of deployments service
+  title: Deployments Internal API
+  version: "1"
+servers:
+- url: http://mender-deployments:8080/api/internal/v1/deployments
+paths:
+  /health:
+    get:
+      operationId: Check Health
+      responses:
+        "204":
+          content: {}
+          description: |
+            Service is healthy and all dependencies are up and running.
+        "500":
+          content:
+            application/json:
+              example:
+                error: internal error
+                request_id: ffd712be-d697-4cb7-814b-88ff1e2eb5f6
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: |
+            Unexpected internal error
+        "503":
+          content:
+            application/json:
+              example:
+                error: "error reaching MongoDB: context deadline exceeded"
+                request_id: ffd712be-d697-4cb7-814b-88ff1e2eb5f6
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: |
+            Service unhealthy / not ready to accept traffic. At least one dependency is not running.
+      summary: Check the health of the service
+      tags:
+      - Internal API
+  /alive:
+    get:
+      operationId: Check Liveliness
+      responses:
+        "204":
+          content: {}
+          description: Service is up and running.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: |
+            Internal API error
+      summary: |
+        Trivial endpoint that unconditionally returns an empty 200 response whenever the API handler is running correctly.
+      tags:
+      - Internal API
+  /tenants/{id}/storage/settings:
+    get:
+      description: |
+        Returns an object with per tenant storage layer specific settings.
+      operationId: Get Storage Settings
+      parameters:
+      - description: Tenant ID
+        in: path
+        name: id
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StorageSettings'
+          description: Successful response with all available settings.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Internal error.
+      summary: Get storage setting for a given tenant
+      tags:
+      - Internal API
+    put:
+      description: Set the storage layer settings for a given tenant.
+      operationId: Set Storage Settings
+      parameters:
+      - description: Tenant ID
+        in: path
+        name: id
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/StorageSettings'
+        description: |-
+          Settings to set.
+          If set to null or an empty object, the tenant will use the default settings.
+        required: false
+      responses:
+        "204":
+          content: {}
+          description: Settings updated.
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: The request body is malformed.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Internal server error.
+      summary: Set storage settings for a given tenant
+      tags:
+      - Internal API
+  /tenants/{id}/limits/storage:
+    get:
+      description: |
+        Get storage limit and current storage usage for given tenant.
+        If the limit value is 0 it means storage space is unlimited
+      operationId: Get Storage Usage
+      parameters:
+      - description: Tenant ID
+        in: path
+        name: id
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StorageUsage'
+          description: Successful response.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Internal Server Error.
+      summary: Get storage limit and current storage usage for given tenant
+    put:
+      description: |
+        Set storage limit for given tenant.
+        If the limit value is 0 it means storage space is unlimited
+      operationId: Set Storage Limit
+      parameters:
+      - description: Tenant ID
+        in: path
+        name: id
+        required: true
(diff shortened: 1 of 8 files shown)


backend/docs/devices_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,396 @@
...
+                    expire: 2016-03-11T13:03:17.063493443Z
+                  device_types_compatible:
+                    - rspi
+                    - rspi2
+                    - rspi0
+              schema:
+                $ref: "#/components/schemas/DeploymentInstructions"
+          description: Successful response.
+        "204":
+          content: {}
```
merlin-northern: is this one on purpose explicitly mentioned as empty?

backend/docs/devices_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,396 @@
...
+            - already-installed
+          type: string
+        substate:
+          description: Additional state information
+          type: string
+      required:
+        - status
+      type: object
+    DeploymentInstructions:
+      example:
```
merlin-northern: why do we start with an example? we never started in that way, is this part of some practices?

backend/docs/devices_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,396 @@
...
+      properties:
+        messages:
+          description: Array of log entries of a deployment
+          items:
+            $ref: "#/components/schemas/DeploymentLog_messages_inner"
+          type: array
+      required:
+        - messages
+      type: object
+    DeploymentInstructions_artifact_source:
```
merlin-northern: was this your decision to introduce those, I mean: to separate every object into a new schema definition? also: I will not comment about it more, I do not think we were using this name notation before, was it a decision to switch to it?

backend/docs/internal_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,1130 @@
...
+  version: "1"
+servers:
+  - url: http://mender-deployments:8080/api/internal/v1/deployments
+paths:
+  /health:
+    get:
+      operationId: Check Health
+      responses:
+        "204":
+          content: {}
```
merlin-northern: I will not repeat the comments to keep it clear.

backend/docs/internal_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,1130 @@
...
+              example:
+                error: "error reaching MongoDB: context deadline exceeded"
+                request_id: ffd712be-d697-4cb7-814b-88ff1e2eb5f6
+              schema:
+                $ref: "#/components/schemas/Error"
+          description: |
+            Service unhealthy / not ready to accept traffic. At least one dependency is not running.
+      summary: Check the health of the service
+      tags:
+        - Internal API
```
merlin-northern: we are moving tags and summary to the end. while I am not adamant here, I got used to having it at the top. all the comments I make about the order of fields are based on what we already have in device auth open api3 -- I want to keep it consistent; we either change here or create a PR for the deviceauth.

backend/docs/internal_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,1130 @@
...
+          description: Service is up and running.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+          description: |
+            Internal API error
+      summary: |
+        Trivial endpoint that unconditionally returns an empty 200 response whenever the API handler is running correctly.
```
merlin-northern: I know how it is and was, but either this is not true or the responses are not matching.

backend/docs/internal_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,1130 @@
...
+        "503":
+          content:
+            application/json:
+              example:
+                error: "error reaching MongoDB: context deadline exceeded"
+                request_id: ffd712be-d697-4cb7-814b-88ff1e2eb5f6
+              schema:
+                $ref: "#/components/schemas/Error"
+          description: |
+            Service unhealthy / not ready to accept traffic. At least one dependency is not running.
```
merlin-northern: same story as in the other comments: the description we usually put first.

backend/docs/internal_v1_deployments.oai.yaml
```diff
@@ -0,0 +1,1130 @@
...
+      operationId: Get Storage Settings
+      parameters:
+        - description: Tenant ID
+          in: path
+          name: id
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
```
merlin-northern: why are we using double quotes? werent it the single ones preferred?

backend/docs/devices_v1_deployments.oai.yaml
merlin-northern, line 1: general comment for all the files: I believe we should add the default set of tests via the generator for *.oai.yaml (as we have them now for deviceauth) with this PR. we are adding new files, and the automatically generated tests will check many things that maybe difficult to catch when we modify those later.

backend/docs/devices_v1_deployments.oai.yaml
alfrunes: It's just an artifact of the generated output. I have not touched any of these files apart from removing some internally generated fields.

backend/docs/devices_v1_deployments.oai.yaml
alfrunes: That was not part of the scope. But the specs themselves are generated by openapi-generator os they should map one to one with the old swagger specs.

merlin-northern: ok no problema. so you want to do it with the test task?

merlin-northern: I have another general comment, why the inter-fix here is "oai"?

alfrunes: It was meant to avoid conflict with other potential backend internal documentation. Now that you mention it, I seem to have confused the short name for OpenAPI Specification (OAS) with another not-so-open business :face_with_peeking_eye: 
I moved all the specs to the directory `backend/docs/api` to avoid confusion :shrug:

backend/docs/internal_v1_deployments.oai.yaml
alfrunes: This was my yamlls formatter. I regenerated the file without applying additional formatting.

Side-note; the generated output is actually a little inconsistent with the use of single-quotes and double-quotes - it tends to prefer double-quotes, but for responses it uses single quotes.

backend/docs/api/internal_v1_deployments.yaml
```diff
@@ -0,0 +1,1124 @@
...
+          description: Service is up and running.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: |
+            Internal API error
+      summary: |
+        Trivial endpoint that unconditionally returns an empty 200 response whenever the API handler is running correctly.
```
merlin-northern: Screenshot 2025-02-25 at 10 56 36

backend/services/deployments/docs/devices_api.yml
```diff
@@ -1,3 +1,4 @@
+# DEPRECATED: Superseded by OpenAPI v3 specs in ../../../docs/*.oai.yaml
```
merlin-northern:
```bash
#ls ../../../docs/*.oai.yaml
ls: cannot access '../../../docs/*.oai.yaml': No such file or directory
```

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -0,0 +1,1286 @@
...
+      - description: Deployment status filter.
+        in: query
+        name: status
+        schema:
+          enum:
+          - inprogress
+          - finished
+          - pending
+          type: string
+      - description: |
```
merlin-northern, line 41: it seems to be that we use `|` randomly, here we have it, but on the line 50 below with ` - description: Results page number`  we dont.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -0,0 +1,1286 @@
...
+                    artifact_name: test
+                    rootfs-image.checksum: 32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99
+                    rootfs-image.version: test
+                  artifact_depends:
+                    device_type:
+                    - test
+                  clears_artifact_provides:
+                  - rootfs-image.*
+                  size: 36891648
+                  modified: 2016-03-11T13:03:17.063493443Z
```
merlin-northern, line 327: could you double check that this is formatted correctly, I maybe seeing things.
```yaml
                  clears_artifact_provides:
                  - rootfs-image.*
                  size: 36891648
                  modified: 2016-03-11T13:03:17.063493443Z
```
looks strange/scary.

merlin-northern (dismissed):
left some extra comments, approving not to block.

backend/docs/api/management_v2_deployments.yaml
alfrunes: `clears_artifact_provides` is an array of strings it's correct, although I would prefer an indentation, the generator didn't produce it and I don't want to create unnecessary conflicts with enterprise.

merlin-northern (approved):

PR #463, opened 2025-02-20
Commitlint mandatory

oldgiova:
In some cases git cliff is ignoring commits when they don't respect the conventional commit guideline, for example when the body don't start with a blank line.
This is causing issue with the changelog generator (git cliff).

We have two way to prevent this to happen:
1. we don't filter out unconventional commit; see https://github.com/mendersoftware/mendertesting/blob/master/utils/cliff.toml#L89 thus this will lead to unformatted changelog entries
2. we set the commitlinting and the conventional commit as mandatory (this PR) - thus this would cause development friction, also among the contributor


mzedel (dismissed):

alfrunes (dismissed):

oldgiova:
Had to force-push a conflict resolution

mzedel (approved):

PR #467, opened 2025-02-20
test: snooze staging-deployments other 5 min

oldgiova:
Push forward the staging-deployments tests another 5 minutes, to allow for a complete staging.hosted.mender.io installation before doing tests.

Ticket: QA-920


mzedel (approved):
🤞

PR #471, opened 2025-02-24
MEN-8140 - fix(gui): fixed pagination for pending deployments

mzedel:


mineralsfree (approved):
We definitely need ts :)

PR #472, opened 2025-02-24
MEN-8078 - release tags editing behaviour for read-only users

mzedel:


frontend/src/js/store/devicesSlice/thunks.tsx
```diff
@@ -967,13 +967,22 @@ export const setDeviceTags = createAsyncThunk(`${sliceName}/setDeviceTags`, ({ d
   // to prevent tag set failures, retrieve the device & use the freshest etag we can get
   Promise.resolve(dispatch(getDeviceById(deviceId))).then(device => {
     const headers = device.etag ? { 'If-Match': device.etag } : {};
-    return GeneralApi.put(
-      `${inventoryApiUrl}/devices/${deviceId}/tags`,
-      Object.entries(tags).map(([name, value]) => ({ name, value })),
-      { headers }
-    )
+    const { isNameChange, tagList } = Object.entries(tags).reduce(
```
mineralsfree: I think readability can be improved with this: 
```
const tagList = Object.entries(tags).map(([name, value]) => ({ name, value }));
const isNameChange = tagList.some(({ name }) => name === 'name');
```

mzedel: Ah, my falling in love with reducers strike again 🙈 - will adjust!

mineralsfree (approved):

PR #180, opened 2024-11-05
fix(gui): fixed an error that prevented upgrading account to paid plan

mzedel:
Ticket: None
Changelog: None

PR #473, opened 2025-02-24
Refac/rtk wrapping

mzedel:


mineralsfree:
I think it is also a good time to start using `useAppSelector` :thinking:    (https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks)

mzedel:
> I think it is also a good time to start using `useAppSelector` 🤔 (https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks)

Yeah, I had that as an internal debate while going through this 😅...
For the selectors I thought we could get most of the typing done in the selector definitions/ in the package already, then they won't be tied to the specific definitions of each product, but instead only tied to the store layout the selectors care about (so e.g. when adopting the packages in Alvaldi we will have to redefine `useAppDispatch` as this is tied to the Mender store layout). What do you think?

frontend/src/js/common-ui/dialogs/ConfirmDismissHelpTips.tsx
```diff
@@ -24,7 +24,7 @@ const { setShowDismissOnboardingTipsDialog } = storeActions;
 
 export const ConfirmDismissHelptips = () => {
   const dispatch = useAppDispatch();
-  const onClose = () => dispatch(setOnboardingCanceled());
+  const onClose = () => dispatch(setOnboardingCanceled()).unwrap();
```
mineralsfree, line 27: There is no "follow up" on this call

frontend/src/js/common-ui/dialogs/PhysicalDeviceOnboarding.tsx
```diff
@@ -208,7 +208,7 @@ export const PhysicalDeviceOnboarding = ({ progress }) => {
   const ComponentToShow = steps[progress];
   return (
     <ComponentToShow
-      advanceOnboarding={step => dispatch(advanceOnboarding(step))}
+      advanceOnboarding={step => dispatch(advanceOnboarding(step)).unwrap()}
```
mineralsfree, line 211: There is no "follow up" on this call

mineralsfree (changes requested):
Many of the added .unwrap() calls in thunk dispatches are not followed up on, processed, or caught (I’m only mentioning two of them in the review). That changes the way things work and could potentially affect behavior. I’d rather add them as we go, in cases where we actually need something from the thunk or need to handle an error.

frontend/src/js/common-ui/dialogs/PhysicalDeviceOnboarding.tsx
mzedel: How would either this component know that the child component doesn't need it, how would the child component know that it has to unwrap the advancement?

frontend/src/js/common-ui/dialogs/ConfirmDismissHelpTips.tsx
mzedel: The component doesn't know what this is handling and the `BaseDialog` this is passed into won't know either - if that relies on the resolution of the call, it won't be aware of RTK and the wrapping state.

mzedel:
> Many of the added .unwrap() calls in thunk dispatches are not followed up on, processed, or caught (I’m only mentioning two of them in the review).

IMO that is a misalignment of the isolation of a component/ who is responsible of handling the redux calls...

> That changes the way things work and could potentially affect behavior. I’d rather add them as we go, in cases where we actually need something from the thunk or need to handle an error.

Yes, but unlike in other cases we have had multiple issue caused by thunks not being unwrapped as the assumption for the codebase used to be to receive non-wrapped return values.

@mineralsfree do you have any opinions on:
> For the selectors I thought we could get most of the typing done in the selector definitions/ in the package already, then they won't be tied to the specific definitions of each product, but instead only tied to the store layout the selectors care about (so e.g. when adopting the packages in Alvaldi we will have to redefine useAppDispatch as this is tied to the Mender store layout). What do you think?

mineralsfree:
> 
> IMO that is a misalignment of the isolation of a component/ who is responsible of handling the redux calls...
> 
I think one of the issues we’re facing is inconsistency in error handling, and I’m not sure that unwrapping thunk calls really helps with that. It also seems like we mostly rely on selectors rather than using the result of the thunk dispatch directly.

We could proceed with unwrapping everywhere and returning results, even if they're left hanging,  but to me that feels a bit off. Personally, I’d only unwrap and return values when they’re actually used or handled.

Merging this PR as-is might also introduce more subtle bugs or regressions down the line, especially if we continue to mix  patterns inconsistently

That said, I’m aware you have a better overall view of how this is used across the app, so I’m not insisting and just sharing my take
 
> @mineralsfree do you have any opinions on:
> 
> > For the selectors I thought we could get most of the typing done in the selector definitions/ in the package already, then they won't be tied to the specific definitions of each product, but instead only tied to the store layout the selectors care about (so e.g. when adopting the packages in Alvaldi we will have to redefine useAppDispatch as this is tied to the Mender store layout). What do you think?

I think we can use slice-isolated selectors that will have type of the slice and then, consumer app will create a store with the slices it needs. We could export a  function for selectors creating for every slice that would accept a selector to select a slice of the store. 
e.g.
`const  { getDeploymentsSelectionState, ... } = createDeploymentSelectors((state: RootState) => state.deployments)`
which can be reexported and used in components

mzedel:
> I think one of the issues we’re facing is inconsistency in error handling, and I’m not sure that unwrapping thunk calls really helps with that. It also seems like we mostly rely on selectors rather than using the result of the thunk dispatch directly.

Maybe for context:

https://github.com/mendersoftware/mender-server/pull/472/commits/9b0c7226f187a77a4831460dd48ea8bc2cba9ef2 https://github.com/mendersoftware/mender-server/pull/180/commits/967a07aba1d12069582a0457f584a1b3be3fc3f4
https://github.com/mendersoftware/mender-server/commit/c3cffda732ecb31c883ab66c83632ae204a033aa

were some of the instances that triggered this... since the unwrapping has to be handled by the caller, the error case here is also the responsibility of the components/ won't be caught in the store. And this could introduce issues - but OTOH it is meant to address the already unintended issues introduced by the RTK switch and the requirement for the unwrapping (e.g. https://github.com/mendersoftware/mender-server/pull/473/files#diff-0c43228820b70f552e9223206cc5d716dfa3a0c6bb5704ac1f7e6a22cb88f09dR979 will ATM never get the `etag` from the response, other follow ups will always run as the thunk execution was successful, but an error caught in the thunk might not be unwrapped)...
Does that clarify the motivation?

>  It also seems like we mostly rely on selectors rather than using the result of the thunk dispatch directly.

Unfortunately that's not quite how it is so far - if we could rely more on the RTK handling/ RTK query we would get the error tracking more easily, but since the UI data is assembled from the various services/ in bigger thunks we would have to introduce the thunk status handling ourselves to then rely on the selectors - which again requires quite a bit more changes.

---

> I think we can use slice-isolated selectors that will have type of the slice and then, consumer app will create a store with the slices it needs. We could export a function for selectors creating for every slice that would accept a selector to select a slice of the store. e.g. `const { getDeploymentsSelectionState, ... } = createDeploymentSelectors((state: RootState) => state.deployments)` which can be reexported and used in components

Yes - that's roughly what I meant (the "regular" selectors are luckily already only concerned with their own state layout, so we can rely on their own understanding directly). Unfortunately that won't be enough as we have the `commonSelectors`/ selectors spanning multiple slices (and some of these are quite crucial) 😕
But for these we could indeed produce selectors based on the app specific RootState and handle the discrepancies in store layout within the selectors 👍.

frontend/src/js/components/devices/device-details/DeviceTags.tsx
```diff
@@ -70,6 +70,7 @@ export const DeviceTags = ({ device, setSnackbar, userCapabilities }) => {
     Tracking.event({ category: 'devices', action: 'modify_tags' });
     setIsEditDisabled(true);
     return dispatch(setDeviceTags({ deviceId: device.id, tags: changedTags }))
+      .unwrap()
```
mzedel, line 73: @mineralsfree next one: https://northerntech.atlassian.net/browse/ME-529

PR #476, opened 2025-02-24
MEN-7632 - chore(gui): switched virtual device onboarding to docker based client

mzedel:
due to the absence of the qemu device type, checks for a virtual device now rely on the kernel coming from a buildkit system - not sure if that's enough, but I couldn't find an alternative attribute

### DIFF
--- a/frontend/src/js/store/onboardingSlice/thunks.ts
+++ b/frontend/src/js/store/onboardingSlice/thunks.ts
@@ -70,7 +70,7 @@ const deductOnboardingState = ({ devicesById, devicesByStatus, onboardingState, 
     ),
     showTips: onboardingState.showTips != null ? onboardingState.showTips : true,
     deviceType,
-    approach: onboardingState.approach || (deviceType.some(type => type.startsWith('qemu')) ? 'virtual' : 'physical'),
+    approach: onboardingState.approach || (deviceType.some(type => type.startsWith('qemu') || type === 'generic-x86_64') ? 'virtual' : 'physical'),
     progress
   };
 };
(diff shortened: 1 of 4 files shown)


frontend/src/js/store/onboardingSlice/thunks.ts
```diff
@@ -70,7 +72,7 @@ const deductOnboardingState = ({ devicesById, devicesByStatus, onboardingState,
     ),
     showTips: onboardingState.showTips != null ? onboardingState.showTips : true,
     deviceType,
-    approach: onboardingState.approach || (deviceType.some(type => type.startsWith('qemu')) ? 'virtual' : 'physical'),
```
lluiscampos, line 73: Can't we rely on the device type `generic-x86_64`? If the issue is that the name is not appropriate (too generic :sweat_smile: ) we can change it to something explicit like `virtual-device-x86_64` or whatever you suggest.

So far this has been our internal image, and now that we are about to expose it to the users we have the opportunity to improve UX.

mzedel: my assumption was with `generic-x86_64` there would be a high chance there's other users as well that land on this as a fallback - if that's not the case I'll happily use that!

lluiscampos: > my assumption was with generic-x86_64 there would be a high chance there's other users as well that land on this as a fallback

It is a grounded concern. I really don't know (nor that I know why we chose that device_type in the first place).

From another point of view, these tooltips are shown on onboarding, so we don't really expect users other devices around so it will be very rare to have a collision with a user defined `generic-x86_64` device type.

I would keep it simple and use the device type.

lluiscampos (dismissed):
I don't have an strong opinion. Pre-approving and trusting your judgment :+1:

lluiscampos (approved):

PR #477, opened 2025-02-24
MEN-7989,MEN-8135 - fix(gui): fixed time filter in past deployments

mineralsfree:
Ticket: MEN-7989,  MEN-8135

### DIFF
--- a/frontend/tests/e2e_tests/integration/04-deployments.spec.ts
+++ b/frontend/tests/e2e_tests/integration/04-deployments.spec.ts
@@ -11,13 +11,24 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+import { Page } from '@playwright/test';
 import dayjs from 'dayjs';
+import advancedFormat from 'dayjs/plugin/advancedFormat.js';
 import isBetween from 'dayjs/plugin/isBetween.js';
 
 import test, { expect } from '../fixtures/fixtures';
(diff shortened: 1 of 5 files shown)


mzedel (dismissed):

frontend/tests/e2e_tests/integration/04-deployments.spec.ts
```diff
@@ -29,8 +31,22 @@ test.describe('Deployments', () => {
...
-  test('allows shortcut deployments', async ({ loggedInPage: page }) => {
+  test('allows shortcut deployments', async ({ baseUrl, loggedInPage: page }) => {
+    // check filters on deployment page
+    const checkTimeFilter = async (name: string, isSetToday?: boolean) => {
+      const input = await page.getByRole('textbox', { name });
+      if (isSetToday) {
+        await expect(input).toHaveValue(dayjs().format('MMMM Do'));
+      }
+      await expect(input.locator('..')).not.toHaveClass(/Mui-error/);
+    };
```
mzedel: This should ideally go into the commands, or at least outside of the test - then the page would be passed in too

mzedel (changes requested):
Checking the time filters after the deployment is good, but the navigating to the deployments page should go to a separate test.

frontend/tests/e2e_tests/integration/04-deployments.spec.ts
mineralsfree: I'll keep it in the file, as it probably won't be reused elsewhere

mzedel (approved):

PR #478, opened 2025-02-25
MEN-8154 - fix(gui): fixed an issue with editing credit card when no billing add…

mineralsfree:
…ress was provided

Ticket: MEN-8154

### DIFF
--- a/frontend/src/js/store/organizationSlice/types.ts
+++ b/frontend/src/js/store/organizationSlice/types.ts
@@ -19,7 +19,7 @@ import { Tenant } from '../api/types/Tenant';
 import { SORTING_OPTIONS } from '../commonConstants';
 
 //TODO: improve types
-interface Card {
+export interface Card {
   brand: string;
   expiration: {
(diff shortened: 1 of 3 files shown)


frontend/src/js/store/organizationSlice/types.ts
```diff
@@ -95,7 +95,7 @@ export interface Addon {
 }
 
 export interface BillingProfile {
```
mzedel, line 97: seeing this, it looks like `BillingProfileUpdate` would be the fitting type from the API, while the `BillingProfile` type there requires the address ☹️

mzedel (approved):

frontend/src/js/store/organizationSlice/types.ts
mineralsfree: I think `BillingProfileUpdate` meant to be used for editing billing profile, while `BillingProfile` was supposed to come from the backend. The reality is quite different anyway :smile:

PR #481, opened 2025-02-26
MEN-8118 - fix(gui): fixed location not being changed on device status filter change

mineralsfree:
Ticket: MEN-8118

### DIFF
--- a/frontend/src/js/components/devices/DeviceGroups.tsx
+++ b/frontend/src/js/components/devices/DeviceGroups.tsx
@@ -11,7 +11,7 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
-import React, { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/devices/DeviceGroups.tsx
```diff
@@ -246,6 +246,11 @@ export const DeviceGroups = () => {
 
   const toggleMakeGatewayClick = () => setShowMakeGateway(toggle);
 
+  const changeLocation = (newLocation: string) => {
+    isInitialized.current = false;
+    setLocationParams({ pageState: { ...deviceListState, state: newLocation }, filters, selectedGroup });
+  };
```
mzedel: This I assume will lead to the `onDeviceStateSelectionChange` callback being reset on every render of `DeviceGroups` (iirc the `useCallback` is needed there to have a properly tracked `refreshTrigger` value). Please wrap the `changeLocation` in a `useCallback` too.

mzedel (changes requested):

mzedel (approved):

PR #483, opened 2025-02-26
feat(iot-manager): getting events by integration by ID

merlin-northern:
Allows to:
* get device events from iot-manager by integration ID
* creating new integration creates a new ID
* maintains backward compatibility for GET /event
* with UI calling the new endpoint this fixes the showing of events from old webhooks in the integraiton details

Changelog: Do not show events from previous webhooks in the integration details by making possible for the UI to get the events by integration ID, at the same time create new ID on ech integration creation.
Ticket: MEN-7801

### DIFF
--- a/backend/services/iot-manager/api/http/management.go
+++ b/backend/services/iot-manager/api/http/management.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Northern.tech AS
+// Copyright 2025 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -233,6 +233,18 @@ func (h *ManagementHandler) GetEvents(c *gin.Context) {
 			err,
 		)
 		return
+	}
+	integrationID := c.Request.URL.Query().Get(paramQueryIntegrationID)
+	if len(integrationID) > 0 {
+		if err := uuid.Validate(integrationID); err == nil {
+			filter.IntegrationID = &integrationID
+		} else {
+			rest.RenderError(c,
+				http.StatusBadRequest,
+				ErrInvalidIntegrationID,
+			)
+			return
+		}
 	}
 
 	events, err := h.app.GetEvents(ctx, *filter)
--- a/backend/services/iot-manager/api/http/management_state.go
+++ b/backend/services/iot-manager/api/http/management_state.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	paramDeviceID      = "id"
-	paramIntegrationID = "integrationId"
+	paramDeviceID           = "id"
+	paramIntegrationID      = "integrationId"
+	paramQueryIntegrationID = "integration_id"
 )
 
 var (
--- a/backend/services/iot-manager/store/mongo/datastore_mongo.go
+++ b/backend/services/iot-manager/store/mongo/datastore_mongo.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Northern.tech AS
+// Copyright 2025 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -45,6 +45,8 @@ const (
 	KeyProvider       = "provider"
 	KeyTenantID       = "tenant_id"
 	KeyCredentials    = "credentials"
+	KeyStatus         = "status"
+	KeyIntegrationID  = "integration_id"
 
 	ConnectTimeoutSeconds = 10
 	defaultAutomigrate    = false
@@ -294,15 +296,8 @@ func (db *DataStoreMongo) CreateIntegration(
 	ctx context.Context,
 	integration model.Integration,
 ) (*model.Integration, error) {
-	var tenantID string
-	if id := identity.FromContext(ctx); id != nil {
-		tenantID = id.Tenant
-	}
 	collIntegrations := db.Collection(CollNameIntegrations)
-
-	// Force a single integration per tenant by utilizing unique '_id' index
-	integration.ID = uuid.NewSHA1(uuid.NameSpaceOID, []byte(tenantID))
-
+	integration.ID = uuid.New()
 	_, err := collIntegrations.
 		InsertOne(ctx, mstore.WithTenantID(ctx, integration))
 	if err != nil {
--- a/backend/services/iot-manager/store/mongo/migrations.go
+++ b/backend/services/iot-manager/store/mongo/migrations.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Northern.tech AS
+// Copyright 2025 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -26,7 +26,7 @@ import (
 
 const (
 	// DbVersion is the current schema version
-	DbVersion = "1.1.1"
+	DbVersion = "1.2.0"
 
 	// DbName is the database name
 	DbName = "iot_manager"
@@ -63,6 +63,10 @@ func Migrate(ctx context.Context,
 			db:     db,
 		},
 		&migration_1_1_1{
+			client: client,
+			db:     db,
+		},
+		&migration_1_2_0{
 			client: client,
 			db:     db,
 		},
--- a/backend/services/iot-manager/api/http/management_test.go
+++ b/backend/services/iot-manager/api/http/management_test.go
@@ -1010,6 +1010,7 @@ func (neverExpireContext) Deadline() (time.Time, bool) {
 
 func TestGetEvents(t *testing.T) {
 	t.Parallel()
+	integrationId := uuid.New().String()
 	testCases := []struct {
 		Name string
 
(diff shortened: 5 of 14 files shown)


merlin-northern:
@alfrunes could you give me an early review? it is untested atm.

backend/services/iot-manager/store/mongo/datastore_mongo.go
```diff
@@ -294,14 +296,15 @@ func (db *DataStoreMongo) CreateIntegration(
...
 	integration model.Integration,
 ) (*model.Integration, error) {
-	var tenantID string
-	if id := identity.FromContext(ctx); id != nil {
-		tenantID = id.Tenant
-	}
+	//var tenantID string
+	//if id := identity.FromContext(ctx); id != nil {
+	//	tenantID = id.Tenant
+	//}
```
alfrunes: ?

alfrunes (commented):
I would drop the part with the integration ID and let them be predictable as this leads to another potentially problematic migration.

There should've been two commits here. One with a `feat` changelog entry for the new endpoint, and another (`fix`?) for changing the ID format. But I would prefer not doing the latter unless we relax the implicit uniqueness constraint.

backend/services/iot-manager/store/mongo/datastore_mongo.go
merlin-northern: fixed.

merlin-northern:
> I would drop the part with the integration ID and let them be predictable as this leads to another potentially problematic migration.
> 
> There should've been two commits here. One with a `feat` changelog entry for the new endpoint, and another (`fix`?) for changing the ID format. But I would prefer not doing the latter unless we relax the implicit uniqueness constraint.

thanks I proceeded as we agreed during MSTM.

merlin-northern:
> > I would drop the part with the integration ID and let them be predictable as this leads to another potentially problematic migration.
> > There should've been two commits here. One with a `feat` changelog entry for the new endpoint, and another (`fix`?) for changing the ID format. But I would prefer not doing the latter unless we relax the implicit uniqueness constraint.
> 
> thanks I proceeded as we agreed during MSTM.

and fixed the commits and other things you mentioned.

backend/services/iot-manager/api/http/management.go
```diff
@@ -234,6 +234,12 @@ func (h *ManagementHandler) GetEvents(c *gin.Context) {
 		)
 		return
 	}
+	integrationID := c.Param(paramIntegrationID)
+	if len(integrationID) > 0 {
+		if err := uuid.Validate(integrationID); err == nil {
+			filter.IntegrationID = &integrationID
+		}
```
alfrunes: The behavior here is a bit unintuitive.
If you pass an integration ID that is not UUID it will return all events, if you pass a random uuid you get no results.
If this is a proper REST resource, I believe the correct thing to do is to also check for the existence of the integration with a given ID and return 404 if it doesn't exist.
Another, probably easier way, would adding `integration_id` as a query parameter to the original endpoint and simply filter by whatever you get (an invalid ID would return empty results).

merlin-northern: sure. I went with the BadRequest return when the id is bad. I also went for bonus points and added API docs :>

alfrunes: I'd still say returning 400 is incorrect behavior for a REST resource that doesn't exist.
It would be more suitable if it was a query parameter.

kjaskiewiczz (changes requested):
I'm pretty sure it's not working
```
iot_manager> db.integrations.find()
[
  {
    _id: UUID('ae7e022f-6cb3-4297-9037-c5e8acc72c63'),
    provider: 'webhook',
    credentials: {
      type: 'http',
      http: {
        url: 'https://webhook.site/48b652c2-c7c8-4391-aa57-cb31b0d7bfbe'
      }
    },
    tenant_id: ''
  },
  {
    _id: UUID('7c58f78b-c8c1-4b87-95ee-70dd57ef457e'),
    provider: 'webhook',
    credentials: {
      type: 'http',
      http: {
        url: 'https://webhook.site/48b652c2-c7c8-4391-aa57-cb31b0d7bfbf'
      }
    },
    tenant_id: ''
  }
]
```
`level=warning msg="migration to version 1.1.2 skipped, target version 1.1.1 is lower"`
because you didn't update https://github.com/mendersoftware/mender-server/blob/3f043588de5ef451b8d52ed529207143c2dc258f/backend/services/iot-manager/store/mongo/migrations.go#L29

> Another, probably easier way, would adding integration_id as a query parameter to the original endpoint and simply filter by whatever you get (an invalid ID would return empty results).

+1

> I would drop the part with the integration ID and let them be predictable as this leads to another potentially problematic migration.

I'm pretty sure I was not fully aware of what we were discussing last time during MSTM, sorry...
@alfrunes the migration doesn't look very complex, but I agree it will complicate the release - we need to wait for the index to be created, so we'll need a downtime (this is also generic issue for our frequent releases).

Still, if we will keep integration Id generated from tenant Id, the whole thing will not work as we are trying to make it work (if I'm not wrong) - if I'll create integration, generate some events, remove integration and create new one, get /events/{integration_id} (or /events?integration_id={integration_id}) will return events from the previous integration(s), because new integration Id will be the same as the old one (in fact, same as all integrations for given tenant).

Can we hide change to Id generation behind feature flag?
We could deploy software which generates integration Id in the old way (because new way will be disabled with feature flag). In this case the migration which will create index can run in the background. Once index will be created we could enable new way of generating integration Ids by toggling feature flag. Once all done, we could prepare PR cleaning up feature flag.

What do you think @alfrunes @merlin-northern?
Am I making it too complicated?

backend/services/iot-manager/api/http/management.go
merlin-northern: okey, so since it is you 😘 I modified, and now it is integration id in the query, I also extended backend integration test. the only thing:

> I'd still say returning 400 is incorrect behavior for a REST resource that doesn't exist.

I am sorry but the correct behaviour when integration id defined in the API docs as uuid format is given in a bad format is BadRequest. unless you think otherwise?

> I believe the correct thing to do is to also check for the existence of the integration with a given ID and return 404 if it doesn't exist.

I am sorry but you want to pass arbitrary string to lower layers, app and db, without validation. this is insecure, and I refuse to do it. if the integration id is not in uuid format it is a BadRequest. the NotFound will be returned automatically if the integration is not found.

merlin-northern:
>  DbVersion = "1.1.1" 

thanks for catching the DbVersion. I have no idea how on earth is worked, but perhaps I should revisit the test.

> Another, probably easier way, would adding integration_id as a query parameter to the original endpoint and simply filter by whatever you get (an invalid ID would return empty results).

I am sorry, how can you +1 to the "filter by whatever you get"? this is by definition insecure coding. we have to validate everything we get before passing further.

merlin-northern:
> Can we hide change to Id generation behind feature flag?

sure. do you think it makes sense to introduce feature flags in open source? I can do it in the enterprise. makes sense?

kjaskiewiczz:
> > Another, probably easier way, would adding integration_id as a query parameter to the original endpoint and simply filter by whatever you get (an invalid ID would return empty results).
> 
> I am sorry, how can you +1 to the "filter by whatever you get"? this is by definition insecure coding. we have to validate everything we get before passing further.

I guess I'm just stupid. I was referring to the first part (query param instead of path param). I know you were presenting it during MSTM, but it's actually difficult to follow/judge the solution without looking at it for some time and/or having alternative approaches presented during the meeting.
About validating integration id - I agree we should validate input data, but
> I am sorry but the correct behaviour when integration id defined in the API docs as uuid format is given in a bad format is BadRequest. unless you think otherwise?

you didn't mention integration id is an uuid in the docs (you introduced it after my last review).

kjaskiewiczz:
> sure. do you think it makes sense to introduce feature flags in open source? I can do it in the enterprise. makes sense?

if the whole idea make sense to @alfrunes then we can have it in the <s>OS</s> Enterprise only, I think

merlin-northern:
> > > Another, probably easier way, would adding integration_id as a query parameter to the original endpoint and simply filter by whatever you get (an invalid ID would return empty results).
> > 
> > 
> > I am sorry, how can you +1 to the "filter by whatever you get"? this is by definition insecure coding. we have to validate everything we get before passing further.
> 
> I guess I'm just stupid. I was referring to the first part (query param instead of path param). I know you were presenting it during MSTM, but it's actually difficult to follow/judge the solution without looking at it for some time and/or having alternative approaches presented during the meeting. About validating integration id - I agree we should validate input data, but
> 
> > I am sorry but the correct behaviour when integration id defined in the API docs as uuid format is given in a bad format is BadRequest. unless you think otherwise?
> 
> you didn't mention integration id is an uuid in the docs (you introduced it after my last review).

do not worry. lets get this done.

alfrunes:
If there's a feature flag, I believe it should be present both places to avoid diversions and confusing the configuration between the two.
However, I don't fully see the point of having the feature flag. My initial assumption about the migration system not being backwards compatible was wrong. It is backwards, but not forwards compatible - which also seems to be correct in this case. To not over-complicate things, I suggest we move on with the implementation as is unless you disagree.

merlin-northern:
> If there's a feature flag, I believe it should be present both places to avoid diversions and confusing the configuration between the two. However, I don't fully see the point of having the feature flag. My initial assumption about the migration system not being backwards compatible was wrong. It is backwards, but not forwards compatible - which also seems to be correct in this case. To not over-complicate things, I suggest we move on with the implementation as is unless you disagree.

if we are voting, I would like to move on as well. @kjaskiewiczz would you be willing to get the feature flag when some other occasion presents itself?

kjaskiewiczz:
@merlin-northern yes, let's continue without feature flag

backend/services/iot-manager/store/mongo/migrations.go
```diff
@@ -26,7 +26,7 @@ import (
 
 const (
 	// DbVersion is the current schema version
-	DbVersion = "1.1.1"
+	DbVersion = "1.1.2"
```
kjaskiewiczz: this should be in the second commit

backend/services/iot-manager/api/http/management_state.go
```diff
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	paramDeviceID      = "id"
-	paramIntegrationID = "integrationId"
+	paramDeviceID           = "id"
+	paramIntegrationID      = "integrationId"
```
alfrunes, line 32: This parameter is unused now, right?

backend/services/iot-manager/store/mongo/migrations.go
alfrunes: One small request, can we bump the minor version since this migration is required for the service to remain backwards compatible?

kjaskiewiczz (dismissed):

alfrunes (dismissed):

backend/services/iot-manager/api/http/management_state.go
alfrunes: Sorry, I was confused by the indentation change.

backend/services/iot-manager/store/mongo/migrations.go
merlin-northern: sure. 😘

merlin-northern: sure. it meant not only rebase with conflicts, but also discarding of your review 😘

backend/services/iot-manager/api/http/management_state.go
merlin-northern: do not despair! it is going to be alright. 😘

alfrunes (approved):

kjaskiewiczz (approved):

PR #484, opened 2025-02-26
MEN-7958 - feat(gui): added sentry integration

mzedel:


oldgiova (commented):
LGTM, just an observation: with this change the Frontend is locking with the product Sentry. I guess there is not a agnostic sdk, like OpenTelemetry for this integration, right?

frontend/src/js/components/App.tsx
```diff
@@ -97,6 +109,19 @@ const useStyles = makeStyles()(() => ({
...
+  const Sentry = await import(/* webpackChunkName: "@sentry/react" */ '@sentry/react');
+  Sentry.init({
+    dsn: location,
+    integrations: [browserTracingIntegration(), replayIntegration({ networkDetailAllowUrls: [window.location.origin] })],
+    release: `mender-frontend@${commit}`,
+    tracesSampleRate, // defaults to capturing 100% of the transactions
+    tracePropagationTargets: ['localhost', /^https:\/\/(\w*\.)*hosted\.mender\.io/, 'https://docker.mender.io'],
+    replaysSessionSampleRate, // defaults to 0.1 in the environment settings, to be adjusted externally
+    replaysOnErrorSampleRate: 1.0 // always change the sample rate to 100% when sampling sessions where errors occur
+  });
```
mineralsfree, line 122: we might also want to set `normalizeDepth`, as default value (3) will truncate redux state we have. https://docs.sentry.io/platforms/javascript/guides/react/configuration/options/#normalizeDepth
https://docs.sentry.io/platforms/javascript/guides/react/features/redux/#normalization-depth

frontend/src/js/store/store.ts
```diff
@@ -73,11 +74,35 @@ const rejectionLoggerMiddleware = () => next => action => {
   return next(action);
 };
 
+const sentryReduxEnhancer = createReduxEnhancer({
+  actionTransformer: action => {
```
mineralsfree, line 81: `users/loginUser/pending` might expose user email and password, so we want to ignore it as well

frontend/src/js/store/store.ts
```diff
@@ -73,11 +74,35 @@ const rejectionLoggerMiddleware = () => next => action => {
...
+    if (action.type === userActions.successfullyLoggedIn.type) {
+      return null;
+    }
+    return action;
+  },
+  // Transform the state to remove sensitive information
+  stateTransformer: (state: RootState) => {
+    const transformedState = {
```
mineralsfree, line 89: Maybe we should exclude Organization (Tenant) token, just to be on a safe side ?

mzedel:
> LGTM, just an observation: with this change the Frontend is locking with the product Sentry. I guess there is not a agnostic sdk, like OpenTelemetry for this integration, right?

There is an OpenTelemetry SDK to integrate directly with OpenTelemetry, but the browser support is still [considered experimental](https://opentelemetry.io/docs/languages/js/getting-started/browser/). OTOH the Sentry SDK is used as basis by many of the alternative solutions, so the change should be rather easy. We could add OpenTelemetry instrumentation to the Sentry instrumentation here to maybe move to an OpenTelemetry setup eventually - but IMO with the browser SDK not being quite ready we can also look into going there once the SDK is ready.

frontend/src/js/store/store.ts
```diff
@@ -73,11 +74,35 @@ const rejectionLoggerMiddleware = () => next => action => {
   return next(action);
```
mineralsfree, line 75: I wonder if we should add sentry error capture here? Or will it catch thunk rejections automagically? 
https://docs.sentry.io/platforms/javascript/guides/react/usage/#capturing-errors

mineralsfree (commented):
Few concerns, looking forward to try it out :+1:

oldgiova:
> There is an OpenTelemetry SDK to integrate directly with OpenTelemetry, but the browser support is still [considered experimental](https://opentelemetry.io/docs/languages/js/getting-started/browser/). OTOH the Sentry SDK is used as basis by many of the alternative solutions, so the change should be rather easy. We could add OpenTelemetry instrumentation to the Sentry instrumentation here to maybe move to an OpenTelemetry setup eventually - but IMO with the browser SDK not being quite ready we can also look into going there once the SDK is ready.

Then no concerns from me. I'm going to approve it because the non-frontend stack LGTM

oldgiova (dismissed):

frontend/src/js/components/App.tsx
mzedel: We will still get the actions log and honestly that's a flood of information already, so I would try it with the truncated state for now.
Because of the created noise (compared to error & session data) I made this to be opt in only for now, so that if we come across an error we can't deduct from the reported info already, we can enable it for a while to track the changes.

frontend/src/js/store/store.ts
mzedel: 👌 + the related ones - I would still expect more adjustments to be needed over time...

frontend/src/js/store/store.ts
mzedel: done, but I expect us to adjust both the traced data + the redux info down the line (as we will likely have to adjust the response body processing e.g.)

frontend/src/js/store/store.ts
mzedel: We will get them caught through the error logging already, but we can also be more explicit (and also get rid of our final class component) down the line.

oldgiova (approved):

PR #485, opened 2025-02-26
fix: Stop deviceconnect's `/connect` endpoints from sending ping

bahaa-ghazal:
Changelog: Title
Ticket: MEN-8061

### DIFF
--- a/backend/services/deviceconnect/api/http/device.go
+++ b/backend/services/deviceconnect/api/http/device.go
@@ -39,9 +39,6 @@ import (
 )
 
 var (
-	// Time allowed to read the next pong message from the peer.
-	pongWait = time.Minute
-
 	// Seconds allowed to write a message to the peer.
 	writeWait = time.Second * 10
 )
@@ -234,22 +231,12 @@ func (h DeviceController) connectWSWriter(
 	}()
 
 	// handle the ping-pong connection health check
-	err = conn.SetReadDeadline(time.Now().Add(pongWait))
 	if err != nil {
 		l.Error(err)
 		return err
 	}
 
-	pingPeriod := (pongWait * 9) / 10
-	ticker := time.NewTicker(pingPeriod)
-	defer ticker.Stop()
-	conn.SetPongHandler(func(string) error {
-		ticker.Reset(pingPeriod)
-		return conn.SetReadDeadline(time.Now().Add(pongWait))
-	})
 	conn.SetPingHandler(func(msg string) error {
-		ticker.Reset(pingPeriod)
-		err := conn.SetReadDeadline(time.Now().Add(pongWait))
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


backend/services/deviceconnect/api/http/device.go
```diff
@@ -240,15 +240,7 @@ func (h DeviceController) connectWSWriter(
...
-	pingPeriod := (pongWait * 9) / 10
-	ticker := time.NewTicker(pingPeriod)
-	defer ticker.Stop()
-	conn.SetPongHandler(func(string) error {
-		ticker.Reset(pingPeriod)
-		return conn.SetReadDeadline(time.Now().Add(pongWait))
-	})
 	conn.SetPingHandler(func(msg string) error {
-		ticker.Reset(pingPeriod)
 		err := conn.SetReadDeadline(time.Now().Add(pongWait))
```
alfrunes: Since the deadline is never set anywhere else, this would lead to the breaking the connection after a minute unless the client sends another ping.
```suggestion
```

backend/services/deviceconnect/api/http/management.go
```diff
@@ -346,15 +334,7 @@ func (h ManagementController) websocketWriter(
...
-	pingPeriod := (pongWait * 9) / 10
-	ticker := time.NewTicker(pingPeriod)
-	defer ticker.Stop()
-	conn.SetPongHandler(func(string) error {
-		ticker.Reset(pingPeriod)
-		return conn.SetReadDeadline(time.Now().Add(pongWait))
-	})
 	conn.SetPingHandler(func(msg string) error {
-		ticker.Reset(pingPeriod)
 		err := conn.SetReadDeadline(time.Now().Add(pongWait))
```
alfrunes: Same here. Could you remove this one too?
```suggestion
```

alfrunes (changes requested):
The `SetReadDeadline` calls on the connection now depends on the client sending ping requests.
Could you please clean up all the calls to `SetReadDeadline` (and clean up the `pongWait` constant)?
There's no need for setting the connection deadline since we rely on TCP keep-alives for connection liveliness and the proxy should block indefinitely on the websocket connection regardless.

backend/services/deviceconnect/api/http/device.go
alfrunes: There's another one on line 237:
https://github.com/mendersoftware/mender-server/blob/ceb2a79f4b41bc014f5f91094f44f682b1cbc986/backend/services/deviceconnect/api/http/device.go#L237-L241

backend/services/deviceconnect/api/http/management.go
alfrunes: There's another call on line 331:
https://github.com/mendersoftware/mender-server/blob/ceb2a79f4b41bc014f5f91094f44f682b1cbc986/backend/services/deviceconnect/api/http/management.go#L331-L335

alfrunes (approved):

alfrunes:
@bahaa-ghazal Could you rebase your PR on `main`? There is a broken dependency in Gitlab that is preventing CI from running. The issue was just fixed in `main`.

bahaa-ghazal:
@alfrunes the ci is collapsing 😱

PR #498, opened 2025-03-03
MEN-8108 - fix(gui): fixed permission extension issue in roles form

mineralsfree:
Ticket: MEN-8108

### DIFF
--- a/frontend/src/js/components/settings/role-management/PermissionsItems.test.ts
+++ b/frontend/src/js/components/settings/role-management/PermissionsItems.test.ts
@@ -1,0 +1,63 @@
+// Copyright 2025 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/settings/role-management/PermissionsItems.test.ts
```diff
@@ -0,0 +1,63 @@
...
+const items = [
+  {
+    'title': 'All releases',
+    'notFound': false
+  },
+  {
+    'title': '123',
+    'notFound': false
+  }
+];
```
mzedel: All of these do not align with the object format for the rest of the codebase - it would be helpful if you could make an effort to aim for consistency without having to introduce lint rules for every aspect of the code structure. Also in reference to [this](https://github.com/mendersoftware/mender-server/pull/503#discussion_r1980959455).

mzedel (changes requested):
...

mzedel (dismissed):

mzedel (approved):

PR #501, opened 2025-03-04
ci: trigger backend unit tests on backend dependency changes

kjaskiewiczz:
Changelog: None
Ticket: QA-968

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -202,7 +202,7 @@ test:backend:unit:
       artifacts: false
   rules:
     - changes:
-        paths: ["backend/**/*.go"]
+        paths: ["backend/**/*.go", "backend/go.mod"]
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -202,7 +202,7 @@ test:backend:unit:
       artifacts: false
   rules:
     - changes:
-        paths: ["backend/**/*.go"]
+        paths: ["backend/**/*", "backend/go.mod"]
```
alfrunes: Shouldn't we still filter by source files?
Now it will trigger unit tests even if you change acceptance test or other miscellaneous files too.
```suggestion
        paths: ["backend/**/*.go", "backend/go.mod"]
```

kjaskiewiczz: yes, we should

alfrunes (approved):

PR #503, opened 2025-03-04
MEN-8106 - test(gui): fixed skipped united tests

mineralsfree:


frontend/src/js/common-ui/XTerm.tsx
```diff
@@ -24,7 +24,7 @@ const fitAddon = new FitAddon();
 const addons = [fitAddon, searchAddon];
 const defaultOptions = { allowProposedApi: true, scrollback: 5000 };
 
-export const Xterm = ({ className, customKeyEventHandler, options = {}, onResize, style, triggerResize, xtermRef, ...remainingProps }) => {
```
mzedel: For one this should be in a separate commit, but more importantly it is not aligned with the lib https://xtermjs.org - it would be very helpful if you could separate your personal preferences from the required code changes.

mzedel (changes requested):
Looks good for the most part, but another unfortunate change in the mix...

frontend/src/js/common-ui/XTerm.tsx
mineralsfree: It's not my personal preference, it's based on pascal case naming convention for react components. I don't see why it's necessarily should match the library name and casing, as it is our own component, but I will revert this change. :ok_man:    
P.S. it's `XTerm` in comments few lines down :smiley:

mzedel (approved):

PR #504, opened 2025-03-04
MEN-7925 - fix(gui): default to eu in eu.hosted domain during signup

mineralsfree:
Ticket: MEN-7925

### DIFF
--- a/frontend/src/js/components/login/Signup.tsx
+++ b/frontend/src/js/components/login/Signup.tsx
@@ -67,6 +67,10 @@ const useStyles = makeStyles()(theme => ({
   },
   logo: { marginLeft: '5vw', marginTop: 45, maxHeight: 50 }
 }));
+const getCurrentLocation = (location: Location): string => {
+  const currentLocation = Object.values(locations).find(value => [`staging.${value.location}`, value.location].includes(location.hostname));
+  return currentLocation ? currentLocation.key : locations.us.key;
+};
 
 export const Signup = () => {
   const [step, setStep] = useState(1);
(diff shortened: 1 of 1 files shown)


frontend/src/js/components/login/Signup.tsx
```diff
@@ -67,6 +67,10 @@ const useStyles = makeStyles()(theme => ({
   },
   logo: { marginLeft: '5vw', marginTop: 45, maxHeight: 50 }
 }));
+const getCurrentLocation = (location: Location) => {
+  const currentLocationKey = Object.keys(locations).find(key => [`staging.${locations[key].location}`, locations[key].location].includes(location.hostname));
```
mzedel:
```suggestion
  const { key: currentLocationKey } = Object.values(locations).find(value => [`staging.${value.location}`, value.location].includes(location.hostname));
```

since you use the key to access the `locations` value or

```suggestion
  const [currentLocationKey] = Object.entries(locations).find(([key, value]) => [`staging.${value.location}`, value.location].includes(location.hostname));
```

if you want to rely on the key instead of the key attribute

mzedel (changes requested):

frontend/src/js/components/login/Signup.tsx
mineralsfree: Both proposed solutions are type unsafe (would cause TypeError if `find` returns `undefined`), but your comment helped me find an issue with my solution, thanks

mzedel: Then use a variant that is both type safe + a proper use of the object properties instead:
```suggestion
  const currentLocationKey = 
Object.entries(locations).reduce(
      (accu, [key, value]) => ([`staging.${value.location}`, value.location].includes(window.location.hostname) ? key : accu),
      locations.us.key
    )
```
If you iterate on keys, you shouldn't rely on the values.

mzedel (changes requested):

mzedel (approved):

PR #506, opened 2025-03-05
fix(iot-manager): do not store events if there are no integrations.

merlin-northern:
Changelog: Title
Ticket: MEN-7868

### DIFF
--- a/backend/services/iot-manager/app/app.go
+++ b/backend/services/iot-manager/app/app.go
@@ -323,7 +323,9 @@ func (a *app) setDeviceStatus(ctx context.Context, deviceID string, status model
 		}
 		event.DeliveryStatus = append(event.DeliveryStatus, deliver)
 	}
-	err = a.store.SaveEvent(ctx, event)
+	if len(integrations) > 0 {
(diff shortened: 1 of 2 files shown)


merlin-northern:
I am adding tests but I want to know how many of the existing tests will notice this.

backend/services/iot-manager/app/app.go
```diff
@@ -323,7 +323,9 @@ func (a *app) setDeviceStatus(ctx context.Context, deviceID string, status model
 		}
 		event.DeliveryStatus = append(event.DeliveryStatus, deliver)
 	}
-	err = a.store.SaveEvent(ctx, event)
+	if len(integrations) > 0 {
+		err = a.store.SaveEvent(ctx, event)
+	}
```
alfrunes, line 328: We could also bail out even after calling GetIntegrations.

alfrunes (approved):
LGTM

backend/services/iot-manager/app/app.go
merlin-northern: right. but I had to draw a line somewhere I usually want to write everything de novo.

PR #508, opened 2025-03-05
Ci/job rules

mzedel:
As follow up from https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/1700789853 - which had no backend related changes (yet was api validated), but frontend changes that weren't lint checked...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -173,6 +173,13 @@ test:backend:static:
 test:backend:validate-open-api:
   stage: test
   needs: []
+  rules:
+    - changes:
+        paths: ["backend/**/docs/*.yml", "backend/docs/api/*.yaml"]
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -173,6 +173,13 @@ test:backend:static:
 test:backend:validate-open-api:
   stage: test
   needs: []
+  rules:
+    - changes:
+        paths: ["backend/**/*.yml"]
```
alfrunes: This rule would also target compose files and certain test yaml files. Could we make the rule slightly more specific?
```suggestion
        paths: ["backend/**/docs/*.yml", "backend/docs/api/*.yaml"]
```

alfrunes (approved):
:ok_hand:

PR #507, opened 2025-03-05
MEN-7761 - login & forgotten password changes

mzedel:

PR #509, opened 2025-03-05
test(gui): set aria-label explicitly to fix firefox e2e tests

mineralsfree:
Ticket: MEN-8170

### DIFF
--- a/frontend/src/js/common-ui/forms/TimeframePicker.tsx
+++ b/frontend/src/js/common-ui/forms/TimeframePicker.tsx
@@ -69,6 +69,14 @@ export const TimeframePicker = ({ tonight: propsTonight }) => {
         render={({ field: { onChange, value } }) => (
           <DatePicker
             disableFuture
+            slotProps={{
+              textField: props => ({
+                inputProps: {
+                  ...props.inputProps,
+                  'aria-label': 'From'
+                }
+              })
+            }}
             format="YYYY-MM-DD"
             yearsOrder="desc"
             label="From"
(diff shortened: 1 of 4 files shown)


mzedel (changes requested):
this breaks the labeling from MUI - you can see the connection here: https://github.com/mendersoftware/mender-server/blob/main/frontend/src/js/common-ui/forms/__snapshots__/TimeFramePicker.test.tsx.snap#L496-L508 - the proper fix seems to be smaller: https://github.com/mendersoftware/mender-server/pull/507/commits/48375d568546b5bfd3d9ce9123a3baef336100ce

mineralsfree:
> this breaks the labeling from MUI - you can see the connection here: https://github.com/mendersoftware/mender-server/blob/main/frontend/src/js/common-ui/forms/__snapshots__/TimeFramePicker.test.tsx.snap#L496-L508 - the proper fix seems to be smaller: https://github.com/mendersoftware/mender-server/pull/507/commits/48375d568546b5bfd3d9ce9123a3baef336100ce

Is it passing for Firefox in CI? I've tried your approach before proceeding with aria labels, but it was failing locally 🥲

mzedel:
> Is it passing for Firefox in CI? I've tried your approach before proceeding with aria labels, but it was failing locally 🥲

passed locally, but not in a container or in CI - and it fails in all cases if run in isolation => I'll file an issue with playwright to look into this 👌

mzedel (approved):

PR #510, opened 2025-03-06
ci: Remove tag override for `test-integration` Makefile target

alfrunes:
This is only making local tests more confusing. Putting the image tag override in CI to use images built with coverage instrumentation.


merlin-northern (approved):

backend/Makefile
merlin-northern, line 1: do we need this one?
Screenshot 2025-03-06 at 17 43 43

merlin-northern (dismissed):
approved with small comment

backend/Makefile
alfrunes: You're right. Let's remove it.

kjaskiewiczz (approved):

PR #516, opened 2025-03-10
test: in opensource repo we should run opensource tests.

merlin-northern:
Ticket: None


mzedel:
@merlin-northern what's the state of this?

merlin-northern:
> @merlin-northern what's the state of this?

that was a solution to a deep problem of testing, however something tells me that we will soon change the ways in a more fundamental way, so I will close this one.

PR #524, opened 2025-03-12
fix(gui): fixed an issue that could clear search queries when search was run multiple times in non-search searchfields

mzedel:


mineralsfree (approved):

vpodzime (approved):
Looks good to my limited knowledge :)

PR #527, opened 2025-03-13
Fix/snackbar usage

mzedel:


frontend/src/js/common-ui/SharedSnackbar.tsx
```diff
@@ -31,13 +31,13 @@ export const SharedSnackbar = ({ setSnackbar, snackbar }) => {
     setSnackbar('');
```
mineralsfree, line 31: Please use dispatch and selector here and elsewhere instead of passing it as prop

mzedel: Feel free to create a follow up task to change this, but since this would cause many more changes across the codebase + in order to limit the changes in this PR to what's necessary to fix the broken snackbar definitions this will have to happen in a separate PR.

mineralsfree (approved):

PR #528, opened 2025-03-13
MEN-8221 - login follow ups

mzedel:


frontend/src/js/components/login/Password.tsx
```diff
@@ -129,7 +129,7 @@ export const Password = () => {
     });
 
   return (
```
mineralsfree, line 133: might be also nice to unwrap `dispatch(passwordResetStart(formData.email))`. Otherwise both failed and successfull requests will suggest user to wait for an email

mineralsfree (dismissed):

mineralsfree (approved):

PR #529, opened 2025-03-14
Ci/staging

mzedel:
passing staging test runs on 
[firefox](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9413821912)
[chrome](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/9415271265)


danielskinstad (approved):
Nice, seems reasonable to me

PR #530, opened 2025-03-14
fix: Invalid API path for /api/devices/v1/inventory/device/attributes

alfrunes:


kjaskiewiczz (dismissed):

backend/services/inventory/api/http/api_inventory_test.go
```diff
@@ -1036,7 +1036,7 @@ func TestApiInventoryUpsertAttributes(t *testing.T) {
 		"no auth": {
 			inReq: rtest.MakeTestRequest(&rtest.TestRequest{
 				Method: "PATCH",
-				Path:   "http://localhost" + apiUrlDevicesV1 + uriAttributes,
+				Path:   "http://localhost" + apiUrlDevicesV1 + uriDeviceAttributes,
```
merlin-northern, line 1039: okey, I cannot believe that the tests have not caught this. we need to have the acceptance tests covering all the supported URIs.

now the `uriAttributes` which is still supported as I understand, appears only in 
```go
	devicesAPILegacy.Group(".").Use(contenttype.CheckJSON()).
		PATCH(uriAttributes, mgmtHandler.UpdateDeviceAttributesHandler).
		PUT(uriAttributes, mgmtHandler.UpdateDeviceAttributesHandler)
```
which makes it again untested in the units. we need a serious discussion around this.

alfrunes: I should probably add that these routes are not officially in use (yet), and probably won't be until 5.0. Traefik is actively rewriting the request line to point to the legacy `/api/0.1.0/attributes` path.

merlin-northern: it is even worse then. by the time of 5.0 no one will remember. and there is no chance you can deduce that by reading the code. and also there is a task abut traefik, you know.

merlin-northern (approved):

PR #538, opened 2025-03-20
MEN-8229 - chore(gui): made login default to not showing password & wait for env info instead

mzedel:


mineralsfree (changes requested):
https://github.com/user-attachments/assets/5984591e-f6da-4cf4-adca-b494f76867ca

Those changes don't resolve the issue completely, the login form still has a bounce effect

mzedel:
> Screencast.from.2025-03-26.10-51-59.mp4 
> Those changes don't resolve the issue completely, the login form still has a bounce effect

That unfortunately wasn't visible unless I throttled my machine significantly 😕, but it will be visible in some form when we account for the environment the page is running in. Afaict that's also not what's mentioned in the ticket.
Considering the content of the page, waiting for the data parsing to be done seemed unfitting, so I opted for having the hosted setup as the default, accepting the page shifts from the environment deifnition in the OS deployments instead. What do you think?

mineralsfree:
> > Screencast.from.2025-03-26.10-51-59.mp4
> > Those changes don't resolve the issue completely, the login form still has a bounce effect
> 
> That unfortunately wasn't visible unless I throttled my machine significantly 😕, but it will be visible in some form when we account for the environment the page is running in. Afaict that's also not what's mentioned in the ticket. Considering the content of the page, waiting for the data parsing to be done seemed unfitting, so I opted for having the hosted setup as the default, accepting the page shifts from the environment deifnition in the OS deployments instead. What do you think?

makes sense

mineralsfree (approved):

PR #541, opened 2025-03-21
Ci/staging rated

mzedel:
WIP for now:
tried reusing session data in multiple test runs, but we depend on non-authenticated + SP session data, which OTOH breaks test storage definitions + makes it tricky to rely on the test results (e.g. screenshots/ failure info bleeds into other tests based on session storage info + we lost video recordings for most tests due to https://github.com/microsoft/playwright/issues/14813)


mzedel:
closing in favour of #573

PR #543, opened 2025-03-24
ME-469 - keep sp creation form open

mzedel:


frontend/src/js/components/tenants/TenantCreateForm.tsx
```diff
@@ -145,15 +145,22 @@ export const TenantCreateForm = (props: TenantCreateFormProps) => {
...
+    async data => {
+      const { email, password, device_limit, send_reset_password, ...remainder } = data;
+      let selectionState = { device_limit: Number(device_limit), ...remainder };
+      if (adminExists) {
+        selectionState = { users: [{ role: rolesByName.admin, email }], ...selectionState };
+      } else {
+        selectionState = { admin: { password, email, send_reset_password }, ...selectionState };
+      }
+      const success = await dispatch(addTenant(selectionState)).unwrap();
+      if (success) {
```
mineralsfree: `if (success){}` is not necessary, as `.unwrap()` will throw an exception in case of rejection
https://redux-toolkit.js.org/api/createAsyncThunk#checking-errors-after-dispatching

mineralsfree (dismissed):

frontend/src/js/components/tenants/TenantCreateForm.tsx
mzedel: Thanks for the heads up! With the experience from the RTK migration where the unwrapping was missed when it wasn't bound to an assignment, I would rather be explicit - but with the recent change in the UX design the handling is no longer optional 👌

mineralsfree (approved):

PR #545, opened 2025-03-24
MEN-8212 - deployments pagination fix + e2e test

mzedel:


frontend/tests/e2e_tests/integration/04-deployments.spec.ts
```diff
@@ -121,4 +122,24 @@ test.describe('Deployments', () => {
...
+    );
+    await Promise.all(pendingDeploymentRequests);
+    await page.goto(`${baseUrl}ui/deployments`);
+    await expect(page.getByText(/rows/i)).toBeVisible();
+    // 10 clicks as anything leading outside of the 50 + something releases present (considering the 10 item page size)
+    for (let clickAttempt = 0; clickAttempt < 10; clickAttempt++) {
+      const paginationButton = page.getByTestId('KeyboardArrowRightIcon');
+      await paginationButton.click({ noWaitAfter: true, force: true });
+    }
+    await expect(page.getByText(/queued to start/i).first()).toBeVisible();
```
mineralsfree, line 143: maybe it's better to check if `41-50 of 50` is visible, so we actually making sure that pagination works, instead of checking that page didn't collapse?

mineralsfree (approved):

PR #546, opened 2025-03-25
chore(gui): removed axios reliance in e2e tests

mzedel:


frontend/tests/e2e_tests/utils/commands.ts
```diff
@@ -215,24 +215,16 @@ export const stopDockerClient = async () => {
...
-    httpsAgent: new https.Agent({
-      rejectUnauthorized: false
-    })
+export const login = async (username: string, password: string, baseUrl: string, request: APIRequestContext) => {
+  const response = await request.post(`${baseUrl}api/management/v1/useradm/auth/login`, {
+    headers: { Authorization: `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}` }
   });
 
-  if (request.status !== 200) {
+  if (response.status() !== 200) {
```
mineralsfree: Maybe we can use `response.ok()` as you did in `frontend/tests/e2e_tests/integration/01-login.spec.ts` ?

mineralsfree (dismissed):

frontend/tests/e2e_tests/utils/commands.ts
mzedel: 👌 good point, aligned it in the other locations as well

mineralsfree (approved):

PR #571, opened 2025-04-09
ci: Update enterprise merge PR title to skip CI and add instructions

alfrunes:
Requires https://github.com/mendersoftware/integration-test-runner/pull/373


merlin-northern (approved):
I use completely different instructions, but it is ok.

PR #572, opened 2025-04-10
Fix location header and accesslog for deployments service

alfrunes:


kjaskiewiczz:
@alfrunes is that a breaking change?

alfrunes:
> @alfrunes is that a breaking change?

I consider this a bug fix as the previous relative link location is incorrect. For example for creating deployments it would resolve to the absolute path: `/api/management/v1/deployments/deployments/management/v1/deployments/deployments/<id>`.

kjaskiewiczz (approved):

PR #580, opened 2025-04-15
Increase frontend unit tests timeout

alfrunes:


mzedel (changes requested):
The [commit message introducing the timeout](https://github.com/mendersoftware/mender-server/commit/ef8a8d7bb722e413c6bd29ea295cee777484b296) should show the reasoning behind the shorter cutoff - since the failures don't happen locally I suspect there's something to https://northerntech.atlassian.net/browse/QA-983?focusedCommentId=116926

as to the rules, I guess we could align it more with [the backend unit tests rules](https://github.com/mendersoftware/mender-server/blob/main/.gitlab-ci.yml#L210-L216)

alfrunes:
Thanks for the clarification, although the current timeout assumes no load on the CI runners which seems a bit optimistic. I'm closing this PR and opening a new one for the CI rules.

mzedel:
> Thanks for the clarification, although the current timeout assumes no load on the CI runners which seems a bit optimistic. 

@alfrunes the average runtime is around 2mins, so it should be ample headroom for the tests & based on past test runs anything that gets to the 4min mark will never end up in a passing run. So the idea was to shorten the feedback cycle time for re-runs considering the load dependence.
There is some more context in https://northerntech.atlassian.net/browse/QA-983 if you're interested.

PR #581, opened 2025-04-15
docs: Migrate API specs to OpenAPI v3.0.1

alfrunes:


kjaskiewiczz:
@alfrunes Hey! I was playing a bit with different ways of converting our API specs to OpenAPI 3 format and I've found output from swagger editor to be a bit nicer (still not great), comparing to what openapi-generator produces - examples here - https://github.com/mendersoftware/mender-server/pull/593 https://github.com/mendersoftware/mender-server/pull/594 
It's not a huge difference, but whatever we will decide we should at least be consistent and use same tool for all the specs (I also used different naming pattern for spec files, so once decided which looks better we should also align that).
Can you take a look a share your thoughs?

alfrunes:
> @alfrunes Hey! I was playing a bit with different ways of converting our API specs to OpenAPI 3 format and I've found output from swagger editor to be a bit nicer (still not great), comparing to what openapi-generator produces - examples here - #593 #594 It's not a huge difference, but whatever we will decide we should at least be consistent and use same tool for all the specs (I also used different naming pattern for spec files, so once decided which looks better we should also align that). Can you take a look a share your thoughs?

I didn't see the second PR was for inventory service before I converted this PR. Let's use yours, the results from swagger editor produces better structure for the openapi documents.

PR #587, opened 2025-04-15
ci: use dedicated runner for test:frontend:unit

oldgiova:
To avoid potential noisy neighbor issues

Ticket: QA-983


mzedel (dismissed):

alfrunes (commented):
I'm not sure what's the status now for the Gitlab runners. Is this one already obsolete?

oldgiova:
> I'm not sure what's the status now for the Gitlab runners. Is this one already obsolete?

You mean the GCP runners? The docker Machine executor will be dismissed on May 2027: https://docs.gitlab.com/runner/executors/docker_machine/

So far we still don't have an actual replacement plan; maybe the new Hetzner AX42 (Ubuntu 24.04) or the new K8s executor will help replacing it

PR #593, opened 2025-04-18
docs(useradm): Migrate API specs to OpenAPI v3.0.1

kjaskiewiczz:


backend/docs/api/useradm_internal_v1.yaml
```diff
@@ -0,0 +1,424 @@
...
+        "400":
+          description: Missing or malformed request parameters.
+          content: {}
+        "500":
+          description: Unexpected error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+      x-codegen-request-body-name: tenant
```
alfrunes: Do we need these?

alfrunes (dismissed):

backend/docs/api/useradm_internal_v1.yaml
kjaskiewiczz: nope; good catch!

alfrunes (approved):

PR #594, opened 2025-04-22
docs(inventory): Migrate API specs to OpenAPI v3.0.1

kjaskiewiczz:
Ticket: QA-886

### DIFF
--- a/backend/docs/api/inventory_devices_v1.yaml
+++ b/backend/docs/api/inventory_devices_v1.yaml
@@ -1,0 +1,168 @@
+openapi: 3.0.1
+info:
+  title: Device inventory
+  description: |
+    An API for uploading device attributes. Intended for use by devices.
+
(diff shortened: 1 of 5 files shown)


backend/docs/api/inventory_devices_v1.yaml
```diff
@@ -0,0 +1,170 @@
...
+                $ref: '#/components/schemas/Error'
+        "500":
+          description: Internal server error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+      security:
+      - DeviceJWT: []
+      x-codegen-request-body-name: attributes
```
alfrunes: Do we need to keep these tags in the specs?

alfrunes (dismissed):

backend/docs/api/inventory_devices_v1.yaml
kjaskiewiczz: no; code generators are a bit annoying sometimes...
removed

alfrunes (approved):

PR #599, opened 2025-04-23
ci: retry on no docker in the backend:integration tests

merlin-northern:
Changelog: Title
Ticket: None


lluiscampos (approved):

lluiscampos:
From my (legacy) point of view we don't need a `Changelog: Title` for this

merlin-northern:
> From my (legacy) point of view we don't need a `Changelog: Title` for this

I will never learn -- to extend the scripting to Changelog: None for the ci: tag. sorry for that. I fixed it and also added the relevant ref to the ticket: QA-988 (I once created)

PR #598, opened 2025-04-23
test(gui): limited unit test load in CI to improve stability

mzedel:

PR #600, opened 2025-04-23
QA-983: Add retry on frontend tests jobs while we work on a proper fix

lluiscampos:


frontend/pipeline.yml
```diff
@@ -167,6 +170,9 @@ build:frontend:docker:
       - frontend/licenses.json
 
 .template:test:frontend:acceptance:
+  # QA-983: hard-retry on the CI jobs while we work on a proper fix
+  retry:
+    max: 2
   stage: test
```
mzedel, line 176: this is unrelated to the unit test failures, we can leave it as it should get a consistent result - but resource usage might disagree 🙃

frontend/pipeline.yml
```diff
@@ -66,6 +66,9 @@ test:frontend:licenses:
     - hetzner-amd-beefy
 
 test:frontend:unit:
+  # QA-983: hard-retry on the CI jobs while we work on a proper fix
+  retry:
+    max: 2
```
mzedel, line 71: Unfortunately it might take many more attempts, depending on the load on the runner - not sure if we gain a lot with 2 retries (+ the failing runs are limited to 5mins already) - soo maybe `max: 5` 😁?

lluiscampos: I [tried with 3](https://github.com/mendersoftware/mender-server/compare/b7217070d41495d8dbc307406103ccd88d33edc5..3606a01183ca15a7ecb4fa86258b2f6b3ad9461a) first, but :fire: 

image

frontend/pipeline.yml
lluiscampos: From our Slack conversation I deduced that they were related. Okey, let's just leave it for now :see_no_evil:

frontend/pipeline.yml
mzedel: 😱 oh well, then... ✅

mzedel (approved):

frontend/pipeline.yml
lluiscampos: It is not enough... I think this is useless.

@mzedel  What about `allow_failure` instead?

mzedel: 🤷 as a short term solution, but 😬 and in this case only limited to the unit tests

lluiscampos:
Presumably this PR is obsolete after: 
* https://github.com/mendersoftware/mender-server/pull/598

Closing :see_no_evil:

PR #602, opened 2025-04-24
ci: temporarily ignored unit test success until stability improves

mzedel:


frontend/pipeline.yml
```diff
@@ -83,6 +83,7 @@ test:frontend:unit:
     - npm run test -- --reporter=default --reporter=junit --outputFile.junit=./junit.xml --coverage
   after_script:
     - mkdir logs && cp frontend/logs/* logs/
+  allow_failure: true
```
lluiscampos, line 86: I would add a comment here explicitly with the reference to JIRA. It is important that we revert this eventually.

lluiscampos (approved):
You are on top of the issue, so I trust you either way :point_down:

mzedel:
Closing as we seem to have found a way to get stable results now 🎉

PR #605, opened 2025-04-25
ME-515 - clarified SSO related need for password omission on user creation

mzedel:
results in:
Screenshot 2025-04-25 at 11 37 10


TheMeaningfulEngineer (approved):
Can't test it, but the wording and the link to the docs are the fix for the reported issues.
Thanks

PR #607, opened 2025-04-25
Fix/roles on basic

mzedel:
slight regression after https://northerntech.atlassian.net/browse/MEN-8095 - noted in https://northerntech.atlassian.net/browse/ME-508

### DIFF
--- a/frontend/src/js/components/settings/role-management/RoleDefinition.tsx
+++ b/frontend/src/js/components/settings/role-management/RoleDefinition.tsx
@@ -342,11 +342,11 @@ export const RoleDefinition: FunctionComponent<RoleDefinitionProps> = ({
         onClose={onCancel}
         preCloser={
           editing &&
-          !rolesById[selectedRole.id] && (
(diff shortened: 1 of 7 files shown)


frontend/src/js/components/settings/role-management/RoleManagement.tsx
```diff
@@ -109,7 +120,7 @@ export const RoleManagement = () => {
           <DocsTooltip id={DOCSTIPS.rbac.id} />
         </InfoHintContainer>
       </div>
-      {rolesInitialized ? <DetailsTable columns={columns} items={items} onItemClick={onEditRole} /> : <Loader show={true} />}
+      {rolesInitialized || initTimedOut ? <DetailsTable columns={columns} items={items} onItemClick={onEditRole} /> : <Loader show={true} />}
```
mineralsfree: can't we set `rolesInitialized` to `true` in `finally` block of the fetch instead? I would avoid `setTimeout` workarounds if possible

mineralsfree (approved):

PR #608, opened 2025-04-25
fix(gui): restored global settings layout after general input layout adjustments

mzedel:
before:
Screenshot 2025-04-25 at 16 44 43
after:
Screenshot 2025-04-25 at 16 41 50

The delta artifact config would hide the issue due to its `block` nature.


mineralsfree (approved):
:+1:

PR #609, opened 2025-04-28
ci: switch to podman runner

oldgiova:
To prevent build issues with Ubuntu 24.04 and Debian 12, while keeping safety on Hetzner runners

Ticket: QA-866


mzedel (approved):
🙌 - thanks for following up!

PR #611, opened 2025-04-28
MEN-8309 - fixed device inventory attributes  filtering 

mineralsfree:


frontend/tests/e2e_tests/integration/05-deviceDetails.spec.ts
```diff
@@ -66,6 +66,7 @@ test.describe('Device details', () => {
     const nameInput = await page.getByLabel(/value/i);
     await nameInput.fill(demoDeviceName);
     await page.waitForTimeout(timeouts.oneSecond);
+    await page.getByRole('button', { name: /Add a rule/i }).waitFor();
```
mzedel:
```suggestion
    await page.getByRole('button', { name: /Add a rule/i }).waitFor();
```
As the waiting for the button should be sufficient now - although the one extra second won't matter much in these tests.

frontend/src/js/components/devices/widgets/FilterItem.tsx
```diff
@@ -52,15 +52,13 @@ export const FilterItem = ({ attributes, onChange, onSelect, plan, reset }) => {
...
-            })
-          : null,
-      TIMEOUTS.threeSeconds
+        onSelect({
+          key,
+          operator,
+          scope,
+          value
+        }),
+      TIMEOUTS.debounceDefault
```
mzedel, line 61: I guess there is a small chance we get pushback on this for users losing their value input if they intend to change the attribute - but considering the direct attribute change works: 👍

mzedel (dismissed):
Thanks for this!

mzedel (approved):

PR #612, opened 2025-04-29
MEN-7733: Revert rate limit implementation for deviceauth

alfrunes:
Requirements changed, the rate limits needs to be isolated per device.


alfrunes:
Superseded by #771

PR #614, opened 2025-04-29
Global resource groups to avoid running heavy jobs in parallel

oldgiova:
Ticket: QA-983

To prevent load-biased results on tests. This way the jobs that are
    using the full load of the runner host, should go in a queue and run the
    job one at a time. This should help for reliable test results.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -240,6 +240,9 @@ test:backend:unit:
       when: on_success
   tags:
     - hetzner-amd-beefy
+  # heavy_duty: global resource group for all jobs that require most of the
+  # resources of the runner - QA-983
+  resource_group: heavy_duty
   services:
     - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/mongo:6.0
       alias: mongo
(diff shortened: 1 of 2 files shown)


frontend/vitest.config.ts
```diff
@@ -66,8 +66,8 @@ export default defineConfig(() => {
     pool: 'threads',
     poolOptions: {
       threads: {
-        minThreads: threadCount,
-        maxThreads: threadCount,
+        minWorkers: 4,
+        maxWorkers: 4,
         useAtomics: true
       }
     }
```
mzedel: reading your comment in the ticket I think we can replace the `pool` + `poolOptions` altogether with:
```suggestion
    maxWorkers: 4
```
or `maxWorkers: threadCount`, although locally this seems to still create more processes than the maximum 😕 (and not limit the load of the entire test suite either).

oldgiova: > locally this seems to still create more processes than the maximum 😕 (and not limit the load of the entire test suite either).

Yeah, same on CI, that's why I'm experimenting...

mzedel (approved):
Thanks for the investigation & let's hope this does it 🤞

lluiscampos (approved):
This is a good idea. The pipeline failed so it it was not enough.

oldgiova:
> This is a good idea. The pipeline failed so it it was not enough.

I hope it was because another heavy duty job was running at that time, still without the proper resource_group set. In the next days I'll add the resource_groups to all the required jobs from other projects

PR #615, opened 2025-04-29
feat: Implemented content type checker middleware for gin framework

bahaa-ghazal:
Ticket: MEN-8327
Changelog: Title

### DIFF
--- a/backend/pkg/contenttype/middleware.go
+++ b/backend/pkg/contenttype/middleware.go
@@ -1,0 +1,52 @@
+package contenttype
+
+import (
+	"mime"
+	"net/http"
+	"slices"
+
+	"github.com/gin-gonic/gin"
+	"github.com/pkg/errors"
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+func CheckJSON() gin.HandlerFunc {
+	return Middleware("application/json")
+}
+
+func Middleware(contentTypes ...string) gin.HandlerFunc {
+	if len(contentTypes) == 0 {
+		panic("contenttype.Middleware initialized with no parameters")
+	}
+
+	errMsg := "Bad Content-Type, expected "
+	for i, ctype := range contentTypes {
+		if i > 0 {
+			errMsg += ", "
+		}
+		errMsg += "'" + ctype + "'"
+	}
+
+	return func(c *gin.Context) {
+		if c.Request.ContentLength > 0 {
+			contentType, _, err := mime.ParseMediaType(c.GetHeader("Content-Type"))
+			if err != nil {
+				rest.RenderError(c,
+					http.StatusBadRequest,
+					errors.WithMessage(err, "Invalid value for Content-Type header"),
+				)
+				c.Abort()
+				return
+			}
+
+			if !slices.Contains(contentTypes, contentType) {
+				rest.RenderError(c,
+					http.StatusUnsupportedMediaType,
+					errors.New(errMsg),
+				)
+				c.Abort()
+			}
+		}
+	}
+}
--- a/backend/pkg/contenttype/middleware_test.go
+++ b/backend/pkg/contenttype/middleware_test.go
@@ -1,0 +1,90 @@
+package contenttype
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+func init() {
+	gin.SetMode(gin.ReleaseMode)
+}
+func TestGinMiddleware(t *testing.T) {
+	json := "application/json"
+	text := "text/html"
+
+	testCases := []struct {
+		Name string
+
+		Request *http.Request
+		Options []string
+
+		ResponseCode int
+	}{{
+		Name: "ok, json",
+		Request: rest.MakeTestRequest(&rest.TestRequest{
+			Method: http.MethodPost,
+			Path:   "http://localhost/test",
+			Body:   "foo",
+		}),
+		ResponseCode: 200,
+	}, {
+		Name: "ok, no body",
+		Request: rest.MakeTestRequest(&rest.TestRequest{
+			Method: http.MethodPost,
+			Path:   "http://localhost/test",
+		}),
+		ResponseCode: 200,
+	}, {
+		Name: "ok, text/html",
+		Request: rest.MakeTestRequest(&rest.TestRequest{
+			Method:      http.MethodPost,
+			Path:        "http://localhost/test",
+			ContentType: text,
+		}),
+		Options:      []string{text},
+		ResponseCode: 200,
+	}, {
+		Name: "ok, text/html or json",
+		Request: rest.MakeTestRequest(&rest.TestRequest{
+			Method:      http.MethodPost,
+			Path:        "http://localhost/test",
+			ContentType: text,
+		}),
+		Options:      []string{text, json},
+		ResponseCode: 200,
+	}, {
+		Name: "error, conflicting Content-Type",
+		Request: rest.MakeTestRequest(&rest.TestRequest{
+			Method: http.MethodPost,
+			Path:   "http://localhost/test",
+			Body:   "foo",
+		}),
+		Options:      []string{text},
+		ResponseCode: 415,
+	}}
+	for i := range testCases {
+		tc := testCases[i]
+		t.Run(tc.Name, func(t *testing.T) {
+			router := gin.New()
+			if tc.Options != nil {
+				router.Use(Middleware(tc.Options...))
+			} else {
+				router.Use(CheckJSON())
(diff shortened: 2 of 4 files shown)


backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,91 @@
...
+		if o.defaultType != nil {
+			opt.defaultType = o.defaultType
+		}
+	}
+	return func(c *gin.Context) {
+		ctype, params, _ := mime.ParseMediaType(c.GetHeader("Content-Type"))
+		charset, ok := params["charset"]
+		if !ok {
+			charset = "UTF-8"
+		}
```
alfrunes: Checking the `charset` parameter is only applicable to certain content types.

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+type MiddlewareOption struct {
+	ContentType *string
+	Path        string
```
alfrunes: Filtering by path/method shouldn't be necessary as the middleware should only be applied to the routes that use it when setting up the routes.

alfrunes (changes requested):
I think you're trying to do a bit too many things at the same time.

Could we instead simplify things a bit?
For now all we need is a single middleware function with the signature:
```go
func CheckJSON(c *gin.Context) {
   // your middleware implementation with for `application/json`
}
```
There are a few endpoints that accept different content types (e.g. signup accepts both `application/json` and `application/x-www-form-urldata`) in this case I suppose we could use a middleware generator:
```go
func CheckIn(contentTypes ...string) *gin.Handler {
    return func(c *gin.Context) {
        // Check if the header is in contentTypes
    } 
}

**Comment by bahaa-ghazal** [comment]
@alfrunes I'm a bit confused about how the middleware will be used. Will the middleware be applied to each route separately if needed (if the content type is not `application/json`)? If so how are we gonna check if the type is `application/json` for rest of the routes?

**Comment by alfrunes** [comment]
> @alfrunes I'm a bit confused about how the middleware will be used. Will the middleware be applied to each route separately if needed (if the content type is not `application/json`)? If so how are we gonna check if the type is `application/json` for rest of the routes?

I'm sorry about the confusion. My thinking was that we can apply the middleware on the specific routes when building the API. For example, with the gin framework you can do:
```go
// ...
apiGroup := engine.Group("/api/management/v1/service")
apiGroup.GET("/foobar", handlerGet) // GET endpoints does not need to check content type
apiGroup.Group(".").
  Use(contenttype.CheckJSON).
  POST("/foobar", handlerPost).
  PUT("/foobar/:id", handlerPostResource)
 // etc.
// for single endpoints with special requirements:
apiGroup.POST("/special/flower", 
  contenttype.CheckIn("application/json", "application/x-www-form-urlencoded"), 
  handleSpecialFlower)
```
As some APIs have special needs, I would try to keep the implementation simple and handle it in the routing instead of making a catch-all type of middleware.

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+func CheckJSON() gin.HandlerFunc {
+	return Middleware(false, "application/json")
+}
+
+func Middleware(checkCharest bool, contentTypes ...string) gin.HandlerFunc {
+
```
alfrunes, line 30:
```suggestion
	if len(contentTypes) == 0 {
		panic("contenttype.Middleware initialized with no parameters, this is a programming error!")
	}
```

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,45 @@
...
+		contentType := c.GetHeader("Content-Type")
+		if !checkCharest {
+			contentType, _, _ = mime.ParseMediaType(contentType)
+		}
+
+		if c.Request.ContentLength > 0 && !slices.Contains(contentTypes, contentType) {
+			var errMsg string
+
+			if len(contentTypes) == 0 {
+				errMsg = "Bad Content-Type, expected no Content-Type"
```
alfrunes: Could we instead make it more clear that it doesn't make any sense to use no arguments? (see last comment)

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,45 @@
...
+			} else {
+				for i, ctype := range contentTypes {
+					contentTypes[i] = "'" + ctype + "'"
+				}
```
alfrunes: This is modifying the state of the function. The next time this middleware is called it will compare the quoted content type, and then the quoted quoted content type and so on...
Since the error message never changes, maybe we could build the error message once in the parent function (generator) scope?

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,45 @@
...
+	return Middleware(false, "application/json")
+}
+
+func Middleware(checkCharest bool, contentTypes ...string) gin.HandlerFunc {
+
+	return func(c *gin.Context) {
+		contentType := c.GetHeader("Content-Type")
+		if !checkCharest {
+			contentType, _, _ = mime.ParseMediaType(contentType)
+		}
```
alfrunes: I don't think the `checkCharset` argument is doing what I think it should be doing.
By the way, I think it's a good idea to call `mime.ParseMediaType` always, as it strips off any additional parameters to the header value.

bahaa-ghazal:
Thanks for the feedback @alfrunes 🙏. I was just confused about the usage of the middleware. Anyways, i'm not sure if checking the charset is important, but i kept it as an option.

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+func CheckJSON() gin.HandlerFunc {
+	return Middleware(false, "application/json")
+}
+
+// For stronger checking with charset, checkCharest should be true.
+// The charset should be uppercase.
+func Middleware(checkCharest bool, contentTypes ...string) gin.HandlerFunc {
```
alfrunes: Small typo :upside_down_face:
```suggestion
func Middleware(checkCharset bool, contentTypes ...string) gin.HandlerFunc {
```

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+	errMsg := "Bad Content-Type, expected "
+	for i, ctype := range contentTypes {
+		if i > 0 {
+			errMsg += ", "
+		}
+		errMsg += "'" + ctype + "'"
+	}
+
+	return func(c *gin.Context) {
+		contentType, params, _ := mime.ParseMediaType(c.GetHeader("Content-Type"))
```
alfrunes:
```suggestion
		if c.Request.ContentLength > 0 {
			contentType, params, err := mime.ParseMediaType(c.GetHeader("Content-Type"))
			if err != nil {
				// catching this error is crucial to know if we have a valid header at all.
				rest.RenderError(...)
			}
			// ...
		}
```

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+		charset, ok := params["charset"]
+
+		if checkCharest && ok {
+			contentType += "; charset=" + charset
+		}
```
alfrunes: This assumes that the charset parameter is always the first one. Note that the parameters are position independent so you could theoretically get a header:
```
Content-Type: application/json; foo=bar; charset=UTF-8
```
which has the correct charset with some parameter we don't care about.

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+		errMsg += "'" + ctype + "'"
+	}
+
+	return func(c *gin.Context) {
+		if c.Request.ContentLength > 0 {
+			contentType, _, err := mime.ParseMediaType(c.GetHeader("Content-Type"))
+			if err != nil {
+				rest.RenderError(c,
+					http.StatusBadRequest,
+					errors.Wrap(err, "Content-Type error"),
```
alfrunes: Nitpick: The error message could be a little more clear for the consumer.
```suggestion
					errors.WithMessage(err, "invalid value for Content-Type header"),
```

backend/pkg/contenttype/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+				)
+				c.Abort()
```
alfrunes, line 39: You need to break out of the middleware here after abort so you don't end up with multiple calls to `rest.RenderError` inside the next `if` branch.
```suggestion
				c.Abort()
				return
```

alfrunes (commented):
Just a couple more nitpicks, otherwise looking good!

alfrunes (approved):

PR #621, opened 2025-05-01
fix: Ensure email is always encoded in lowercase when stored

alfrunes:
Added a bson codec for model.Email that will ensure that emails are always encoded in lowercase in the database to ensure case insensitive queries.

Changelog: Title
Ticket: MEN-8328

### DIFF
--- a/backend/services/useradm/store/mongo/bson.go
+++ b/backend/services/useradm/store/mongo/bson.go
@@ -1,0 +1,32 @@
+package mongo
+
+import (
+	"reflect"
+	"strings"
+
+	"go.mongodb.org/mongo-driver/bson/bsoncodec"
+	"go.mongodb.org/mongo-driver/bson/bsonrw"
+
+	"github.com/mendersoftware/mender-server/pkg/mongo/codec"
+	"github.com/mendersoftware/mender-server/services/useradm/model"
+)
+
+func newRegistry() *bsoncodec.Registry {
+	registry := codec.NewRegistry()
+	registry.RegisterTypeEncoder(tEmail, bsoncodec.ValueEncoderFunc(encodeEmail))
+	return registry
+}
+
+var tEmail = reflect.TypeOf(model.Email(""))
+
+func encodeEmail(ec bsoncodec.EncodeContext, w bsonrw.ValueWriter, val reflect.Value) error {
+	if !val.IsValid() || val.Type() != tEmail {
+		return bsoncodec.ValueEncoderError{
+			Name:     "EmailCodec",
+			Types:    []reflect.Type{tEmail},
+			Received: val,
+		}
+	}
+	value := val.Interface().(model.Email)
+	return w.WriteString(strings.ToLower(string(value)))
+}
--- a/backend/services/useradm/store/mongo/datastore_mongo.go
+++ b/backend/services/useradm/store/mongo/datastore_mongo.go
@@ -26,7 +26,6 @@ import (
 	mopts "go.mongodb.org/mongo-driver/mongo/options"
 	"golang.org/x/crypto/bcrypt"
 
-	"github.com/mendersoftware/mender-server/pkg/mongo/codec"
 	"github.com/mendersoftware/mender-server/pkg/mongo/oid"
 	mstore "github.com/mendersoftware/mender-server/pkg/store/v2"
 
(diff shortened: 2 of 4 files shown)


backend/services/useradm/store/mongo/bson.go
```diff
@@ -0,0 +1,24 @@
+package mongo
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/mendersoftware/mender-server/services/useradm/model"
+	"go.mongodb.org/mongo-driver/bson/bsoncodec"
+	"go.mongodb.org/mongo-driver/bson/bsonrw"
```
merlin-northern:
```suggestion
	"go.mongodb.org/mongo-driver/bson/bsoncodec"
	"go.mongodb.org/mongo-driver/bson/bsonrw"

	"github.com/mendersoftware/mender-server/services/useradm/model"
```

backend/services/useradm/store/mongo/bson.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+var tEmail = reflect.TypeOf(model.Email(""))
```
merlin-northern, line 20: do you think there is a solution that does not involve reflection?

merlin-northern (commented):
I admire your go expertise, still. my general comment is: I thought the db queries are case insensitive as it is now, and the problem is only in the SSO callbacks. was it a wrong assumption? we do have e2e for sso, and we definitely need a test for the upper/lower case emails.

alfrunes:
> I thought the db queries are case insensitive as it is now, and the problem is only in the SSO callbacks. was it a wrong assumption?

No, what happens is that email is converted to lowercase when deserializing JSON.

https://github.com/mendersoftware/mender-server/blob/d791eaa4175c526cc3c0e8d1080c56bf9d1e156d/backend/services/useradm/model/user.go#L47-L54

What happened here is that the email is deserialized as string and type cast to Email. This PR is not trying to solve the SSO issue in particular, but to fix the bson encoding of emails so that we don't make the same mistake again. It just happens to solve the SSO issue as a side-effect.

backend/services/useradm/store/mongo/bson.go
alfrunes: I couldn't think of any that would not pull `go.mongodb.org/mongo-driver` as a dependency outside the `mongo` package.

merlin-northern: perhaps making `pkg/store` or similar is a good idea?

alfrunes: It's extending `github.com/mendersoftware/mender-server/pkg/mongo/codec` with an additional encoding of `model.Email` type which is local to this service. I'm not sure what else I should to do?

merlin-northern (dismissed):
okey, approving not to block, I think:
* it is hacky
* impacts performance
* has no test extension so we are unsure it solves the original problem

merlin-northern:
> okey, approving not to block, I think:
> 
> * it is hacky
> * impacts performance
> * has no test extension so we are unsure it solves the original problem

sorry for the use of words when I meant something different. what I meant: I approve, but please note the comments I have.

alfrunes:
> sorry for the use of words when I meant something different. what I meant: I approve, but please note the comments I have.

Noted :memo:
Please don't hesitate to block my PRs if you believe it's not up to the standard.

> * it is hacky

Yes, but [it is what it is](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.17.3/bson/bsoncodec#example-Registry-CustomEncoder).

> * impacts performance

Not sure if I understand what you mean. Can you elaborate?

> * has no test extension so we are unsure it solves the original problem

Shame on me :disappointed: 
I extended unit tests for `GetUserByEmail` (the victim interface that caused the issue) to cover upper case emails too.

alfrunes:
Sorry, I also had to rebase to unblock the test issue with seaweedfs.

merlin-northern (approved):
looks good.

merlin-northern:
> > * impacts performance
> 
> Not sure if I understand what you mean. Can you elaborate?

so I have not checked it, but I dont think reflection is the fastest thing in town. I also have not checked how often this will be called, I feared that very often. but on the second thought, I trust you.

alfrunes:
> so I have not checked it, but I dont think reflection is the fastest thing in town. I also have not checked how often this will be called, I feared that very often. but on the second thought, I trust you.

There are certainly ways of (ab)using `reflect` that can be suboptimal, but the `bsoncodec` (and also standard encoding libraries like [encoding/json](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/encoding/json/decode.go;l=168-183)) also depends on `reflect`.

PR #631, opened 2025-05-02
MEN-8337: feat(create-artifact-worker): Update `mender-artifact` to latest v4.1.0

lluiscampos:
Modifying also the integration to build the tool from source instead or repurposing the upstream Debian package. This has the main advantage that we can compile it statically (by disabling a feature that we don't use) and that we have control of the compatibility aspects of the binary.

### DIFF
--- a/backend/services/create-artifact-worker/Dockerfile
+++ b/backend/services/create-artifact-worker/Dockerfile
@@ -3,13 +3,24 @@ ARG TARGETARCH
 ARG TARGETOS
 ARG BUILDFLAGS="-trimpath"
 ARG LDFLAGS="-s -w"
-ARG MENDER_ARTIFACT_VERSION=4.0.0
+ARG MENDER_ARTIFACT_VERSION=4.1.0
 WORKDIR /build
-# Fetch and install mender-artifact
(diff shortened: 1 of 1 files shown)


backend/services/create-artifact-worker/Dockerfile
```diff
@@ -3,13 +3,25 @@ ARG TARGETARCH
...
 WORKDIR /build
-# Fetch and install mender-artifact
-RUN wget "https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/mender-artifact_${MENDER_ARTIFACT_VERSION}-1%2bubuntu%2bnoble_${TARGETARCH}.deb" \
-  -O mender-artifact.deb && \
-  dpkg -x mender-artifact.deb / && \
-  mkdir /var/cache/create-artifact-worker
+
+# Fetch and build mender-artifact
+RUN \
+  apt update && apt install --assume-yes git && \
```
alfrunes: `git` and `make` ships with the standard `golang` (Debian based) images.

alfrunes (dismissed):

alfrunes (approved):

PR #634, opened 2025-05-05
ci: call docker API GET /_ping to detect the need to retry a job.

merlin-northern:
Ticket: QA-988


mzedel:
@merlin-northern is this still relevant?

merlin-northern:
all of them are relevant. but the docker retry problems have stopped, so lets see how it goes. closing.

PR #637, opened 2025-05-06
test: pinned seaweed image to custom header compatible version

mzedel:


oldgiova (approved):

oldgiova:
Great find! note to self:
* check for upstream fix in the future and remove the pin
* add a note in the Mender Helm documentation

PR #641, opened 2025-05-07
ci: retry on lookup docker fails

oldgiova:
Retry when exit code is 2: dial tcp: lookup docker on *:53: no such host

Ticket: QA-1018


oldgiova:
Sorry, this is not necessary: the previous docker check (docker ps) is the way to run it

PR #642, opened 2025-05-07
ci: enforce e2e isolation

oldgiova:
Don't run more than one e2e job on staging at a time globally

Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -385,9 +385,11 @@ test:prep:
   only:
     - main # can't use rules with delays: https://gitlab.com/gitlab-org/gitlab/-/issues/424203
   when: delayed
+  resource_group: test_staging_e2e
   needs:
     - job: test:prep
       artifacts: true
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -414,28 +416,23 @@ test:prep:
 
 test:staging-deployment:chrome:
   extends: .template:test:staging-deployment
-  resource_group: test-staging-deployment-chrome
   script:
     - npx playwright install chromium
     - npm run test -- --project=chromium
   start_in: 15 minutes
```
mzedel: instead of the sleep in the helm bump, we could cut this down to 3 minutes and stagger the other jobs by a little, to ensure the order is still kept as the jobs are supposed to be run in order of browser popularity

oldgiova: Something like this, if I understood it correctly?

mzedel (approved):

PR #648, opened 2025-05-08
ci: preserve the docker check in the before_script

oldgiova:
When the .build_base is included via extends, but a before_script is present, this latter override the before_script present in the .build_base. This change restores the *requires-docker script. Additionally, added debug messages to the docker checks

Ticket: None

### DIFF
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -186,6 +186,7 @@ build:frontend:docker:
     - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-dind
       alias: docker
   before_script:
+    - !reference [.requires-docker]
     - !reference [.dind-login]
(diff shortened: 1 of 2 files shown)


frontend/pipeline.yml
```diff
@@ -6,6 +6,22 @@ variables:
...
+  -   echo "DEBUG - docker is running - continue"
+  -   break
+  -  fi
+  -  sleep "${DOCKER_RETRY_SLEEP_S}"
+  - done
+  - if [ "${DOCKER_RUNNING}" != "true" ]; then
+  -  echo "DEBUG - docker is not running - exiting"
+  -  exit 192
+  - fi
+
```
mzedel: I think you could omit this here and instead change the `*requires-docker` reference in the `before_script` to: `!reference [.requires-docker]`

oldgiova: Wow! TIL

mzedel (approved):

PR #650, opened 2025-05-08
chore(gui): reduced likelyhood pagination buttons rerender during rapid data changes

mzedel:


frontend/pipeline.yml
```diff
@@ -207,7 +207,7 @@ build:frontend:docker:
         - frontend/junit/results.xml
     when: always
   tags:
-    - hetzner-amd-beefy
+    - k8s
```
oldgiova: By moving to k8s runner, you have to define the DOCKER variables:

```
  variables:
    DOCKER_HOST: tcp://docker:2375
    DOCKER_TLS_CERTDIR: ""
```
And also add the `HEALTHCHECK_TCP_PORT` variable to the Docker service:
```
  services:
      alias: docker
      variables:
        HEALTHCHECK_TCP_PORT: "2375"
```

oldgiova (changes requested):
By moving to k8s runner, you have to adapt the pipeline with specific docker variables.
However, I think moving to k8s is useless: the load on this kind of test is still on the Staging Cluster, and not in the Gitlab Runner, and it ends up paying for an ephemeral k8s runner. But we can try

mzedel:
@oldgiova if there's not that much to gain I'll remove the commit 👌 - the pipeline before that https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/1806921227 worked fine for the e2e tests, but OTOH I haven't seen the test failing that frequently in the last few weeks, so I'm retrying a bit to see if there's something

oldgiova:
> @oldgiova if there's not that much to gain I'll remove the commit 👌 - the pipeline before that https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/1806921227 worked fine for the e2e tests, but OTOH I haven't seen the test failing that frequently in the last few weeks, so I'm retrying a bit to see if there's something

It's fine, let's move to k8s and see what happens. At this stage we have to move things around to find the most stable solution, then optimize for costs later... 
The change request is just for the variables ;)

oldgiova (dismissed):

oldgiova (approved):

PR #651, opened 2025-05-08
test(gui): silenced more warnings about non-seriables

mzedel:
- referencing uploads where we need file references until the upload is done + device twin integration references that come with an explanatory info component

### DIFF
--- a/frontend/src/js/store/store.ts
+++ b/frontend/src/js/store/store.ts
@@ -121,7 +121,7 @@ export const getConfiguredStore = (options = {}) => {
         },
         serializableCheck: {
           ignoredActions: [organizationActions.receiveExternalDeviceIntegrations.name, setSnackbar.name, uploadProgress.name],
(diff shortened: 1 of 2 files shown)


frontend/src/js/store/store.ts
```diff
@@ -121,7 +121,7 @@ export const getConfiguredStore = (options = {}) => {
         },
         serializableCheck: {
           ignoredActions: [organizationActions.receiveExternalDeviceIntegrations.name, setSnackbar.name, uploadProgress.name],
-          ignoredActionPaths: ['uploads', 'snackbar', /payload\..*$/],
+          ignoredActionPaths: ['uploads', 'snackbar', /payload\..*$/, 'meta.arg.file', 'meta.arg.integration.configHint'],
```
mineralsfree: We might also want to remove snackbar related actions from here, as we are not passing jsx anymore

mineralsfree (dismissed):
:+1:

mineralsfree (approved):

PR #653, opened 2025-05-08
refactor(inventory): Migrate from ant0nie/go-json-rest to gin-gonic/gin

bahaa-ghazal:
Ticket: MEN-8236
Changelog: Title


backend/services/inventory/api/http/api_inventory.go
```diff
@@ -564,286 +470,316 @@ func (i *inventoryHandlers) UpdateDeviceTagsHandler(w rest.ResponseWriter, r *re
...
 	hasNext := totalCount > int(page*perPage)
 
-	links := utils.MakePageLinkHdrs(r, page, perPage, hasNext)
+	hints := rest.NewPagingHints().
+		SetPage(page).
+		SetPerPage(perPage).
+		SetHasNext(hasNext).
+		SetTotalCount(int64(totalCount))
+
+	links, err := rest.MakePagingHeaders(c.Request, hints)
```
bahaa-ghazal, line 641: I migrated from using the local implention to the one in rest.utils. The only different is this new logic will also create a link for last page.

backend/pkg/rest.utils/paging.go
```diff
@@ -37,6 +37,14 @@ var (
 	)
 )
 
+func MsgQueryParmInvalid(param, name string) string {
+	return fmt.Sprintf("invalid %s query: \"%s\"", param, name)
+}
+
+func MsgQueryParmLimit(name string) string {
+	return fmt.Sprintf("invalid %s query: value must be a non-zero positive integer", name)
+}
```
alfrunes: This could be tidied up a little:
```suggestion
func ErrQueryParmInvalid(param, name string) error {
	return fmt.Errorf("invalid %s query: \"%s\"", param, name)
}

func ErrQueryParmLimit(name string) error {
	return fmt.Errorf("invalid %s query: value must be a non-zero positive integer", name)
}
```

backend/services/inventory/api/http/routing.go
```diff
@@ -38,27 +46,124 @@ func supportsMethod(method string, methods []string) bool {
...
+	)
+
+	mgmtAPIV1Legacy.GET(uriDevices, mgmtHandler.GetDevicesHandler)
+	mgmtAPIV1Legacy.GET(uriDevice, mgmtHandler.GetDeviceHandler)
+	mgmtAPIV1Legacy.GET(uriDeviceGroups, mgmtHandler.GetDeviceGroupHandler)
+	mgmtAPIV1Legacy.GET(uriGroups, mgmtHandler.GetGroupsHandler)
+	mgmtAPIV1Legacy.GET(uriGroupsDevices, mgmtHandler.GetDevicesByGroupHandler)
+	rest.ApplyMiddlewareToRoutes(
+		contenttype.CheckJSON(),
+		mgmtAPIV1Legacy.DELETE(uriDevice, mgmtHandler.DeleteDeviceInventoryHandler),
```
bahaa-ghazal: I'm not sure if the checking the content type for delete endpoints is necessary

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
 
-	devs, totalCount, err := i.inventory.ListDevices(ctx, ld)
+	devs, totalCount, err := i.App.ListDevices(ctx, ld)
 
 	if err != nil {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: Can you please push the underlying error on the stack here?
```suggestion
		c.Error(err)
		rest.RenderError(c,
			http.StatusInternalServerError,
			errors.New("internal error"),
		)
```

---

Alternatively, you could extend the `rest.utils` package with a new function that does this:
```go
// Same as RenderError but pushed `err` on the stack and outputs `apiMessage` to the caller.
func RenderErrorWithMessage(c *gin.Context, code int, err error, apiMessage string) {
	ctx := c.Request.Context()
	_ = c.Error(err)
	err = &Error{
		Err:       apiMessage,
		RequestID: requestid.FromContext(ctx),
	}
	c.JSON(code, err)
}
```
and then do:
```suggestion
		rest.RenderErrorWithMessage(c,
			http.StatusInternalServerError,
			err,
			"internal error",
		)
```
Which will render a nicer accesslog entry.

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
-	deviceID := r.PathParam("id")
-
-	dev, err := i.inventory.GetDevice(ctx, model.DeviceID(deviceID))
+	dev, err := i.App.GetDevice(ctx, model.DeviceID(deviceID))
 	if err != nil {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: Please see previous comment (`err` not logged).

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
 
-	err := i.inventory.ReplaceAttributes(ctx, model.DeviceID(deviceID),
+	err := i.App.ReplaceAttributes(ctx, model.DeviceID(deviceID),
 		model.DeviceAttributes{}, model.AttrScopeInventory, "")
 	if err != nil && err != store.ErrDevNotFound {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: ^

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
-	deviceID := r.PathParam("device_id")
-
-	err := i.inventory.DeleteDevice(ctx, model.DeviceID(deviceID))
+	err := i.App.DeleteDevice(ctx, model.DeviceID(deviceID))
 	if err != nil && err != store.ErrDevNotFound {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: ^

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
 	}
 
-	err = i.inventory.AddDevice(ctx, dev)
+	err = i.App.AddDevice(ctx, dev)
 	if err != nil {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: ^

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -403,155 +271,193 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
...
+func (i *ManagementAPI) UpdateDeviceAttributesHandler(c *gin.Context) {
+	ctx := c.Request.Context()
+
 	var idata *identity.Identity
 	if idata = identity.FromContext(ctx); idata == nil || !idata.IsDevice {
-		u.RestErrWithLog(w, r, l, errors.New("unauthorized"), http.StatusUnauthorized)
+		rest.RenderError(c,
+			http.StatusUnauthorized,
+			errors.New("unauthorized"),
+		)
```
alfrunes, line 407: Same here, you could do:
```suggestion
		rest.RenderErrorWithMessage(c,
			http.StatusUnauthorized,
			err, "unauthorized",
		)
```

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -564,286 +470,316 @@ func (i *inventoryHandlers) UpdateDeviceTagsHandler(w rest.ResponseWriter, r *re
...
+			cause,
+		)
 		return
 	}
 	if err != nil {
-		u.RestErrWithLogInternal(w, r, l, err)
+		rest.RenderError(c,
+			http.StatusInternalServerError,
+			errors.New("internal error"),
+		)
```
alfrunes: ^
I will stop here, could you go through all places where there is an error that is being replaced and make sure it ends up in the logs?

backend/services/inventory/api/http/routing.go
```diff
@@ -38,27 +46,124 @@ func supportsMethod(method string, methods []string) bool {
...
+	mgmtAPIV1.GET(uriDevice, mgmtHandler.GetDeviceHandler)
+	mgmtAPIV1.GET(uriDeviceGroups, mgmtHandler.GetDeviceGroupHandler)
+	mgmtAPIV1.GET(uriGroups, mgmtHandler.GetGroupsHandler)
+	mgmtAPIV1.GET(uriGroupsDevices, mgmtHandler.GetDevicesByGroupHandler)
+	rest.ApplyMiddlewareToRoutes(
+		contenttype.CheckJSON(),
+		mgmtAPIV1.DELETE(uriDevice, mgmtHandler.DeleteDeviceInventoryHandler),
+		mgmtAPIV1.DELETE(uriDeviceGroup, mgmtHandler.DeleteDeviceGroupHandler),
+		mgmtAPIV1.DELETE(uriGroupsName, mgmtHandler.DeleteGroupHandler),
+		mgmtAPIV1.DELETE(uriGroupsDevices, mgmtHandler.ClearDevicesGroupHandler),
```
alfrunes: `DELETE` is like `GET` and should not have a body at all.

backend/services/inventory/api/http/routing.go
```diff
@@ -38,27 +46,124 @@ func supportsMethod(method string, methods []string) bool {
...
+		mgmtAPIV1.PUT(uriDeviceGroups, mgmtHandler.AddDeviceToGroupHandler),
+		mgmtAPIV1.PATCH(uriGroupsDevices, mgmtHandler.AppendDevicesToGroup),
+		mgmtAPIV1.PUT(uriDeviceTags, mgmtHandler.UpdateDeviceTagsHandler),
+		mgmtAPIV1.PATCH(uriDeviceTags, mgmtHandler.UpdateDeviceTagsHandler),
+	)
```
alfrunes: Sorry I didn't catch this in your previous PR earlier. I took another look at the gin reference and it turns out that the call to make routes (`GET`, `DELETE`, etc.) returns the group object itself. In turn, the call to `ApplyMiddlewareToRoutes` will apply the middleware as many times as there are arguments.

Instead you could do:
```suggestion
	mgmtAPIV1.DELETE(uriDevice, mgmtHandler.DeleteDeviceInventoryHandler),
	mgmtAPIV1.DELETE(uriDeviceGroup, mgmtHandler.DeleteDeviceGroupHandler),
	mgmtAPIV1.DELETE(uriGroupsName, mgmtHandler.DeleteGroupHandler),
	mgmtAPIV1.DELETE(uriGroupsDevices, mgmtHandler.ClearDevicesGroupHandler),

	mgmtAPIV1.Group(".").Use(contenttype.CheckJSON()).
		.PUT(uriDeviceGroups, mgmtHandler.AddDeviceToGroupHandler)
		.PATCH(uriGroupsDevices, mgmtHandler.AppendDevicesToGroup)
		.PUT(uriDeviceTags, mgmtHandler.UpdateDeviceTagsHandler)
		.PATCH(uriDeviceTags, mgmtHandler.UpdateDeviceTagsHandler)
```

alfrunes (changes requested):
First round of preliminary feedback, I have to stop here.
Great job with the translation of the logic from rest to gin. As far as I can see everything looks one to one.
There is big issue in the routes that needs fixing (sorry, I should've caught that in the middleware PR).

backend/services/inventory/api/http/routing.go
alfrunes: I created https://github.com/mendersoftware/mender-server/pull/656

backend/services/inventory/api/http/api_inventory.go
bahaa-ghazal: There is no error to return here. `err` is not defined.

alfrunes: Sorry, you're right. Please disregard this one.

backend/pkg/rest.utils/paging.go
alfrunes, line 1: That's cleaner! :sparkles:
Thanks

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -564,286 +452,292 @@ func (i *inventoryHandlers) UpdateDeviceTagsHandler(w rest.ResponseWriter, r *re
...
+		SetPage(page).
+		SetPerPage(perPage).
+		SetHasNext(hasNext).
+		SetTotalCount(int64(totalCount))
+
+	links, err := rest.MakePagingHeaders(c.Request, hints)
+	if err != nil {
+		rest.RenderInternalError(c, err)
+		return
+	}
```
alfrunes: In the (impossible) event that this function returns an error it should be because of a client error. However, since the PerPage and Page parameters are set in the hints, I don't believe the above call can return an error :shrug:
```suggestion
	if err != nil {
		rest.RenderError(c, http.StatusBadRequest, err)
		return
	}
```

backend/services/inventory/api/http/api_inventory_test.go
```diff
@@ -54,49 +68,47 @@ func ToJson(data interface{}) string {
...
+		} else {
+			req.Header.Set("Authorization", r.Token)
+		}
+	}
+
+	if r.ContentType == "" {
+		req.Header.Set("Content-Type", "application/json")
+	} else {
+		req.Header.Set("Content-Type", r.ContentType)
+	}
```
alfrunes: Technically `Content-Type` doesn't make sense if there's no content.
```suggestion
	if body != nil && r.ContentType == "" {
		req.Header.Set("Content-Type", "application/json")
	} else {
		req.Header.Set("Content-Type", r.ContentType)
	}
```

backend/services/inventory/api/http/api_inventory_test.go
```diff
@@ -14,27 +14,30 @@
...
 	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/ant0ine/go-json-rest/rest/test"
```
alfrunes: The dependency is still here :point_up:
```suggestion
```
It's used here: https://github.com/mendersoftware/mender-server/blob/c8d7ebff5c9addee9c266a6e05f62fde339297f3/backend/services/inventory/api/http/api_inventory_test.go#L3138-L3142

alfrunes (commented):
This is looking pretty good, well done! :ok_hand:
I found one instance of the old rest package in the test, but otherwise I only have a couple nitpicks.

It would be great if I could get a second pair of eyes on this one (@kjaskiewiczz or @merlin-northern) since it involves a lot of changes.

I'll also create a feature branch upstream to stage this pull request until it's ready to be merged into enterprise.

backend/pkg/rest.utils/paging.go
```diff
@@ -69,14 +72,9 @@ func ParsePagingParameters(r *http.Request) (int64, int64, error) {
...
-			return -1, -1, errors.Errorf(
-				"invalid per_page query: \"%s\"",
-				qPerPage,
-			)
-		} else if perPage < 1 {
-			return -1, -1, errors.New("invalid per_page query: " +
-				"value must be a non-zero positive integer",
-			)
+			return -1, -1, ErrQueryParmInvalid("per_page", qPerPage)
+		} else if page < 1 {
```
alfrunes: This condition is using the wrong parameter.
```suggestion
		} else if perPage < 1 {
```
Why didn't the CI catch this? :thinking:

backend/pkg/rest.utils/request.go
alfrunes, line 1: Can we move this to a dedicated test package please? We shouldn't expose hard-coded tokens and test utilities in code linked to the application binary.

alfrunes (changes requested):
Found an issue in the paging parsing that I missed earlier.
I'll check why CI is not running the unit tests :monocle_face:

backend/pkg/rest.utils/request.go
alfrunes: I created a PR to your fork here: https://github.com/bahaa-ghazal/mender-server/pull/1

alfrunes:
> I'll check why CI is not running the unit tests 🧐

Found it! https://github.com/mendersoftware/mender-server/pull/673

alfrunes (approved):

PR #656, opened 2025-05-08
revert: Gin utility for applying middleware to routes

alfrunes:
The func does not implement what it is intended to do and instead applies the middleware to all routes in the group.


bahaa-ghazal (approved):
Oh, i see. Thanks for clarifying 👍

PR #657, opened 2025-05-08
Cherry pick pipeline improvements for 4.0.x

oldgiova:


mzedel (approved):
0d54def7009d9f7bbc622a12373fdbd8b3fb0732 can probably be left out as these will only run `main`, but it won't hurt either and the rest: 👌

oldgiova:
> [0d54def](https://github.com/mendersoftware/mender-server/commit/0d54def7009d9f7bbc622a12373fdbd8b3fb0732) can probably be left out as these will only run `main`, but it won't hurt either and the rest: 👌

Ah, right!

PR #670, opened 2025-05-14
ci: aligned sentry release publishing with current setup

mzedel:
passing pipeline with passing finalizing step: https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/1817774661
+ uploaded releases can be found in sentry at /organizations/sentry/releases/?project=2


oldgiova (approved):
:partying_face:

PR #673, opened 2025-05-14
Fix `pkg` unit tests not run in CI

alfrunes:
Since the default target changed to `generate`, the unit tests has not been running in CI.
Also added `pkg` unit tests to the root `test-unit` Makefile target.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -40,7 +40,7 @@ $(BUILD_TARGETS):
 
 
 .PHONY: test
-test: $(TEST_TARGETS) pkg-test test-integration
+test: test-unit test-integration
 
 .PHONY: $(TEST_TARGETS)
 $(TEST_TARGETS):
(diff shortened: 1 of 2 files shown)


backend/pkg/Makefile
```diff
@@ -2,10 +2,11 @@ LDFLAGS ?= "-s -w"
...
+test:
+	go test $(BUILDFLAGS) $(TESTFLAGS) ./...
+
 .PHONY: generate
 generate:
 	go generate ./...
 
-.PHONY: test
-test:
-	go test $(BUILDFLAGS) $(TESTFLAGS) ./...
```
merlin-northern, line 11: just for me, is this a needed change?

backend/Makefile
```diff
@@ -40,18 +40,18 @@ $(BUILD_TARGETS):
...
 	@$(MAKE) -C services/$(subst -test,,$@) test
 
 .PHONY: pkg-test
 pkg-test:
-	@$(MAKE) -C pkg/
+	@$(MAKE) -C pkg/ test
 
 .PHONY: test-unit
-test-unit: $(TEST_UNIT_TARGETS)
+test-unit: pkg-test $(TEST_UNIT_TARGETS)
```
merlin-northern, line 54: wiat, test-unit reqs pkg-test, pkg-test calls text and test needs test-unit. luckly from the pkg directory. good that I am not lost

merlin-northern (approved):

backend/pkg/Makefile
alfrunes: I prefer test is the default target rather than `generate`, but maybe we should make a target that executes both as default?

PR #674, opened 2025-05-16
ME-529, ME-528 - adjustments to device tag editing

mzedel:
some of this also comes from DK-76

the empty state is more a RFC for ME-528

### DIFF
--- a/frontend/src/js/common-ui/forms/KeyValueEditor.tsx
+++ b/frontend/src/js/common-ui/forms/KeyValueEditor.tsx
@@ -17,6 +17,8 @@ import { Clear as ClearIcon, Add as ContentAddIcon } from '@mui/icons-material';
 import { Fab, FormControl, FormHelperText, IconButton, OutlinedInput } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 
+import { yes } from '@northern.tech/store/constants';
+
 type HelptipProps = {
   [key: string]: any;
   style?: CSSProperties;
@@ -51,7 +53,7 @@ const useStyles = makeStyles()(theme => ({
   }
 }));
 
-export const KeyValueEditor = ({ disabled, errortext, initialInput = {}, inputHelpTipsMap = {}, onInputChange, reset }) => {
+export const KeyValueEditor = ({ disabled, errortext, initialInput = {}, inputHelpTipsMap = {}, onInputChange, onInputUpdate = yes, reset }) => {
   const { classes } = useStyles();
   const [inputs, setInputs] = useState([{ ...emptyInput }]);
   const [error, setError] = useState('');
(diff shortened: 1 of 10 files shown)


frontend/src/js/helptips/HelpTooltips.tsx
```diff
@@ -350,7 +359,12 @@ export const HELPTOOLTIPS = {
   webhookSecret: { id: 'webhookSecret', Component: WebhookSecret }
 };
 
-export const MenderHelpTooltip = props => {
+type MenderHelpTooltipProps = {
+  contentProps?: Record<string, unknown>;
+  id: string;
+} & Omit<HelpTooltipProps, 'setAllTooltipsReadState' | 'setTooltipReadState' | 'tooltip'>;
+
+export const MenderHelpTooltip: React.FC<MenderHelpTooltipProps> = props => {
```
mineralsfree: Maybe `import { FC } from 'React'`?

I am on the side of typing props directly instead of using `FC`, but it's not a big deal

frontend/src/js/common-ui/forms/KeyValueEditor.tsx
```diff
@@ -11,13 +11,31 @@
...
+type InputHelptip = {
+  component: ComponentType<HelptipProps>;
+  position?: string;
+  props?: HelptipProps;
+};
+
+type InputLineItem = {
+  helptip: InputHelptip | null;
+  key: string;
+  ref?: RefObject<unknown>;
```
mineralsfree: is there a purpose of this ref? How is it used?

frontend/src/js/common-ui/forms/KeyValueEditor.tsx
```diff
@@ -85,6 +88,9 @@ export const KeyValueEditor = ({ disabled, errortext, initialInput = {}, inputHe
 
   const reducePairs = listOfPairs => listOfPairs.reduce((accu, item) => ({ ...accu, ...(item.value ? { [item.key]: item.value } : {}) }), {});
 
+  // have this for now, the intention is to move the editor to RHF and then we can reconsider this & the related usage in device config + preauth
+  const reducePairsIndiscriminately = listOfPairs => listOfPairs.reduce((accu, item) => ({ ...accu, [item.key]: item.value }), {});
```
mineralsfree: I find 
```const reducePairsIndiscriminately = (listOfPairs) =>
  Object.fromEntries(listOfPairs.map(({ key, value }) => [key, value]));
  ```
  slightly more readable, but it's once again preference.

mineralsfree (dismissed):
A few small things to consider

frontend/src/js/common-ui/forms/KeyValueEditor.tsx
mzedel: Good question! This _used_ to be the way to position the tooltips showing alongside the editor => removed it now 👌

frontend/src/js/helptips/HelpTooltips.tsx
mzedel: 🤷 here I don't have a preference & made sure to remove the other references there 👌

mzedel:
@mineralsfree as a potential addition to this I added the migration of the KeyValueEditor to RHF https://github.com/mzedel/mender-server/pull/1 - just don't want to attach it to this right away...

mineralsfree (approved):
LGTM

PR #675, opened 2025-05-16
MEN-8397 - removed default organization token visibility

mzedel:
image
image

### DIFF
--- a/frontend/src/js/components/settings/organization/Organization.tsx
+++ b/frontend/src/js/components/settings/organization/Organization.tsx
@@ -187,7 +187,7 @@ export const Organization = () => {
                 dividerDisabled
                 key="org_token"
                 onExpansion={onTokenExpansion}
-                secondary={tenant_token}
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/settings/organization/Organization.tsx
```diff
@@ -187,7 +187,7 @@ export const Organization = () => {
                 dividerDisabled
                 key="org_token"
                 onExpansion={onTokenExpansion}
-                secondary={org.tenant_token}
+                secondary={showTokenWarning ? tenant_token : `${tenant_token.substring(0, 5)}...${tenant_token.substring(tenant_token.length - 5)}`}
```
mineralsfree, line 190: IMO if we use asterisks instead of dots, we'll signify that it's a sensitive information, not just information that didn't fit in the container

mineralsfree (approved):

frontend/src/js/components/settings/organization/Organization.tsx
mzedel: @michaelntech or @chiachenglu maybe - we don't use them anywhere in the UI, but maybe for the upcoming changes..?

PR #677, opened 2025-05-19
Upgrade dependencies

alfrunes:


kjaskiewiczz (approved):

mzedel (approved):
I assume the differences in the indirect dependencies will be taken into account already, the module & image version changes look very reasonable

PR #679, opened 2025-05-19
test: Make `test_get_limit_default_limit` reentrant

alfrunes:
This test is not reentrant with cache and depends on execution order of other tests because of colliding namespace.


merlin-northern (approved):
what it has to do with the reentrancy as stated by the commit message I have no idea.

alfrunes:
> what it has to do with the reentrancy as stated by the commit message I have no idea.

Not the most accurate commit message. It solves an issue that the tests within the class affect each other over executions and this test class does not work well with cache.

PR #690, opened 2025-05-27
QA-995 - chore/use eslint 9

mzedel:


mineralsfree (approved):
Assuming files are auto-fixed, LGTM :)

PR #720, opened 2025-06-11
MEN-8411 - billing/ organization split

mzedel:


frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -151,19 +250,13 @@ export const Billing = () => {
...
-        let addonPrice = '';
-        if (!organization.trial && !isEnterprise) {
-          const planAddon = ADDONS[addon.name][currentPlan] ? ADDONS[addon.name][currentPlan] : ADDONS[addon.name].os;
-          addonPrice = ` - ${planAddon.price}`;
-        }
-        accu.push(`${title}${addonPrice}`);
-      }
-      return accu;
-    }, []) || [];
+  const enabledAddOns = addons.reduce((accu: string[], addon) => {
```
mineralsfree:
```
const enabledAddOns = addons
  .filter(addon => addon.enabled)
  .map(addon => ADDONS[addon.name].title); 
  ```
  maybe this can do the job?

File: frontend/src/js/components/settings/organization/Billing.tsx [hunk 2]
```diff
@@ -139,10 +237,11 @@ export const Billing = () => {
   const isEnterprise = useSelector(getIsEnterprise);
   const organization = useSelector(getOrganization);
   const card = useSelector(getCard);
+  const deviceLimit = useSelector(getDeviceLimit);
   const billing = useSelector(getBillingProfile);
-  const { plan: currentPlan = PLANS.os.id } = organization;
+  const { addons = [], plan: currentPlan = PLANS.os.id, trial, trial_expiration } = organization;
+  const isTrial = !!trial;
```
mineralsfree: Isn't trial boolean already?

frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -179,72 +272,55 @@ export const Billing = () => {
...
-              </a>
-            </InfoText>
-          )}
-        </div>
-      </List>
+            {!billing && !isTrial && (
+              <Alert severity="warning">
+                Your account is not set up for automatic billing. If you believe this is a mistake, please contact{' '}
+                <a href="mailto:support@mender.io" target="_blank" rel="noopener noreferrer">
+                  support@mender.io
```
mineralsfree, line 307: I believe this <a> support element would benefit from being  reusable common component 😄

mineralsfree (dismissed):
Minor things caught my attention, looks good 👍

frontend/src/js/components/settings/organization/Billing.tsx
mzedel: yep, in that case I would also move the title lookup part to the `AddOnDescriptor` component to have the rendering part encapsulated - it would iterate that (very huge) list three times though (which was why I didn't do it in the first place) 🙈

changing 👌

frontend/src/js/components/settings/organization/Billing.tsx
mzedel: yeah - but there are some variations for the content of the element (contact our team, contact support, contact our sales team & as here)... - I'll create a follow up PR 👌

mzedel: #730

mineralsfree (approved):
:+1:

frontend/src/js/components/settings/organization/Billing.tsx
mineralsfree: Yeah, but I think in general we should prioritize readability over performance unless it becomes an issue, and I doubt we get a noticeable improvement from using `reduce` instead of `filter` + `map`

PR #734, opened 2025-06-14
test(deployments): moving deployments tests to Open API 3

merlin-northern:
Ticket: QA-883

### DIFF
--- a/backend/docs/api/devices_v1_deployments.yaml
+++ b/backend/docs/api/devices_v1_deployments.yaml
@@ -248,25 +248,6 @@ paths:
       tags:
       - Device API
 components:
-  responses:
-    NotFoundError:
-      content:
-        application/json:
-          schema:
-            $ref: '#/components/schemas/Error'
-      description: Not Found.
-    InternalServerError:
-      content:
-        application/json:
-          schema:
-            $ref: '#/components/schemas/Error'
-      description: Internal Server Error.
-    InvalidRequestError:
-      content:
-        application/json:
-          schema:
-            $ref: '#/components/schemas/Error'
-      description: Invalid Request.
   schemas:
     Error:
       description: Error descriptor.
@@ -320,11 +301,12 @@ components:
           description: Deployment ID
           type: string
         artifact:
-          $ref: '#/components/schemas/DeploymentInstructions_artifact'
+          $ref: '#/components/schemas/DeploymentInstructionsArtifact'
       required:
       - artifact
       - id
       type: object
+      nullable: true
     DeploymentLog:
       example:
         messages:
@@ -338,12 +320,12 @@ components:
         messages:
           description: Array of log entries of a deployment
           items:
-            $ref: '#/components/schemas/DeploymentLog_messages_inner'
+            $ref: '#/components/schemas/DeploymentLogMessagesInner'
           type: array
       required:
       - messages
       type: object
-    DeploymentInstructions_artifact_source:
+    DeploymentInstructionsArtifactSource:
       properties:
         uri:
           description: URL to fetch the artifact from
@@ -354,12 +336,12 @@ components:
           format: date-time
           type: string
       type: object
-    DeploymentInstructions_artifact:
+    DeploymentInstructionsArtifact:
       properties:
         id:
           type: string
         source:
-          $ref: '#/components/schemas/DeploymentInstructions_artifact_source'
+          $ref: '#/components/schemas/DeploymentInstructionsArtifactSource'
         device_types_compatible:
           description: Compatible device types
           items:
@@ -372,7 +354,7 @@ components:
       - device_types_compatible
       - source
       type: object
-    DeploymentLog_messages_inner:
+    DeploymentLogMessagesInner:
       properties:
         timestamp:
           format: date-time
--- a/backend/services/deployments/Makefile
+++ b/backend/services/deployments/Makefile
@@ -2,4 +2,26 @@ COMPONENT := deployments
 
 override BUILDTAGS_REQUIRED := nopkcs11
 
+DOCFILES := $(wildcard ../../../backend/docs/api/*deployments.yaml)
+MAKEDIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+
+tests/%: ../../../backend/docs/api/%_deployments.yaml
+	[ -e $@ ] && rm -r $@; \
+	docker run --rm -t -v $(MAKEDIR):/work -v $(MAKEDIR)/../../../backend/docs/api:/api -w /work \
+		--ulimit nofile=65536:65536 \
+		-u $(shell id -u):$(shell id -g) \
+		openapitools/openapi-generator-cli:v4.3.1 generate \
+		-g python -i /api/$(notdir $<) \
+		-c tests/.openapi-generator.yml \
+		-o $(dir $@) \
+		--additional-properties=packageName=$* && \
+		sed -i.gitignore -e 's/"log":false/"log":False/g' tests/*/test/test_*.py
+		# see https://github.com/OpenAPITools/openapi-generator/issues/11421
+		# there is a bug in the python generator which produces non-python syntax
+
+.PHONY: docs
+docs: $(patsubst ../../../backend/docs/api/%_deployments.yaml,tests/%,$(DOCFILES))
+
+test_acceptance_run_deps := docs
+
 include ../Makefile.common
--- a/backend/services/deployments/tests/tests/client.py
+++ b/backend/services/deployments/tests/tests/client.py
@@ -22,6 +22,7 @@ import os.path
 import random
 import socket
 
+import jwt
 from base64 import b64encode
 from datetime import datetime
 from collections import OrderedDict
@@ -33,11 +34,17 @@ import pytest  # noqa
 import pytz
 import requests
 
-from bravado.swagger_model import load_file
-from bravado.client import SwaggerClient, RequestsClient
-from bravado.exception import HTTPUnprocessableEntity
-
 from config import pytest_config
+
+from bson.objectid import ObjectId
+from base64 import b64encode
+from datetime import datetime, timedelta
+import hmac
+import uuid
+import management_v2 as mv2
+import management_v1 as mv1
+import devices_v1 as dv1
+import internal_v1 as iv1
 
 DEPLOYMENTS_BASE_URL = "http://{}/api/{}/v1/deployments"
 
@@ -56,35 +63,6 @@ class RequestsApiClient(requests.Session):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.verify = False
-
-
-class SwaggerApiClient(BaseApiClient):
-    """Class that it based on swagger spec. Swagger support is initialized on call
-    to setup_swagger(). Class has no constructor, hence can be used with Pytest"""
-
-    config = {
-        "also_return_response": True,
-        "validate_responses": True,
-        "validate_requests": False,
-        "validate_swagger_spec": False,
-        "use_models": True,
-    }
-
-    log = logging.getLogger("client.Client")
-    spec_option = "spec"
-
-    def __init__(self):
-        self.setup_swagger()
-
-    def setup_swagger(self):
-        self.http_client = RequestsClient()
-        self.http_client.session.verify = False
-
-        spec = pytest_config.getoption(self.spec_option)
-        self.client = SwaggerClient.from_spec(
-            load_file(spec), config=dict(self.config), http_client=self.http_client
-        )
-        self.client.swagger_spec.api_url = self.make_api_url()
 
 
 class ArtifactsClientError(Exception):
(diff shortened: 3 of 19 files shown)


alfrunes (commented):
This looks a lot like the PR to deployments-enterprise. Is this a cherry-pick? If so, could you add the trailer showing it's a cherry-pick (`-x`)?
Also, could you please change the base to a feature branch until a merge resolution is ready for `mender-server-enterprise`?

merlin-northern:
> This looks a lot like the PR to deployments-enterprise. Is this a cherry-pick? If so, could you add the trailer showing it's a cherry-pick (`-x`)? Also, could you please change the base to a feature branch until a merge resolution is ready for `mender-server-enterprise`?

this is not a cherry pick. this is more or less manual pick.

backend/services/deployments/Makefile
```diff
@@ -2,4 +2,26 @@ COMPONENT := deployments
...
+		--ulimit nofile=65536:65536 \
+		-u $(shell id -u):$(shell id -g) \
+		openapitools/openapi-generator-cli:v4.3.1 generate \
+		-g python -i /api/$(notdir $<) \
+		-c tests/.openapi-generator.yml \
+		-o $(dir $@) \
+		--additional-properties=packageName=$* && \
+		sed -i.gitignore -e 's/"log":false/"log":False/g' tests/*/test/test_*.py
+		# see https://github.com/OpenAPITools/openapi-generator/issues/11421
+		# there is a bug in the python generator which produces non-python syntax
```
dhaustein, line 20: Question: This is not called in the CI, is it?

merlin-northern:
> This looks a lot like the PR to deployments-enterprise. Is this a cherry-pick? If so, could you add the trailer showing it's a cherry-pick (`-x`)? Also, could you please change the base to a feature branch until a merge resolution is ready for `mender-server-enterprise`?

done, btw we are missing the rules for the feature branches. also: the merge resolution will be simple: you take nothing from OS keep the ENT. thats why I did it this way. so we dont waste time.

alfrunes:
> > This looks a lot like the PR to deployments-enterprise. Is this a cherry-pick? If so, could you add the trailer showing it's a cherry-pick (`-x`)? Also, could you please change the base to a feature branch until a merge resolution is ready for `mender-server-enterprise`?
> 
> done, btw we are missing the rules for the feature branches. also: the merge resolution will be simple: you take nothing from OS keep the ENT. thats why I did it this way. so we dont waste time.

That's what I figured, but that can easily get complicated once you have multiple conflicts in parallel.

backend/docs/api/devices_v1_deployments.yaml
```diff
@@ -1,4 +1,8 @@
 openapi: 3.0.1
+tags:
+- name: Device API
+  x-displayName: Deployments Device API
+  description: device api
```
alfrunes: This section is unique to only devices v1 and is not present in enterprise.

backend/services/deployments/tests/tests/test_api_internal.py
```diff
@@ -33,13 +31,14 @@
 )
 from client import SimpleArtifactsClient, ArtifactsClientError
 
+from client import internal_v1_client
+from internal_v1.rest import ApiException
+
 
 class TestInternalApiTenantCreate:
     def test_create_ok(self, api_client_int, clean_db):
```
alfrunes, line 39: Unused fixture (also removed in enterprise). Applies to the rest of the tests in this class too.
```suggestion
    def test_create_ok(self, clean_db):
```

backend/services/deployments/tests/tests/client.py
```diff
@@ -232,51 +214,50 @@ def generate_artifact(
...
+    def with_added_artifact(self, description="", size=0, data=None):
+        """Acts as a context manager, adds artifact and yields artifact ID and deletes
+        it upon completion"""
+        artid = None
+        try:
+            artid = self.add_artifact(description=description, size=size, data=data)
+            yield artid
+        finally:
+            if artid is not None:
+                self.delete_artifact(artid)
```
alfrunes: Duplicate
```suggestion
```

backend/services/deployments/tests/tests/test_release_v2.py
```diff
@@ -33,6 +32,49 @@
...
+        "alg": "HS256",
+        "typ": "JWT",
+    }
+    hdr64 = (
+        b64encode(json.dumps(hdr).encode(), altchars=b"-_").decode("ascii").rstrip("=")
+    )
+
+    claims = {
+        "sub": subject,
+        "exp": (datetime.utcnow() + timedelta(hours=1)).isoformat("T"),
```
dhaustein, line 57: Fix: The `exp` takes on a `NumericDate` which the RFC 7519 defines as "the number of seconds from 1970-01-01T00:00:00Z UTC" aka unix epoch timestamp, not an ISO format with a "T" separator. https://datatracker.ietf.org/doc/html/rfc7519#section-2

backend/services/deployments/tests/tests/test_release_v2.py
```diff
@@ -33,6 +32,49 @@
...
 
+import management_v2 as mv2
+from bson.objectid import ObjectId
+from base64 import b64encode
+from datetime import datetime, timedelta
+import hmac
+import uuid
+
+
+def generate_jwt(tenant_id: str = "", subject: str = "", is_user: bool = True) -> str:
```
dhaustein, line 43: Suggestion: We should use something like https://pyjwt.readthedocs.io/en/stable/ instead of doing our own implementation. 
```python
import jwt
from datetime import datetime, timedelta
import uuid

def generate_jwt(tenant_id: str = "", subject: str = "", is_user: bool = True) -> str:
    if len(subject) == 0:
        subject = str(uuid.uuid4())

    payload = {
        "sub": subject,
        "exp": datetime.utcnow() + timedelta(hours=1),  # PyJWT will handle the time conversion
        "mender.user": is_user,
        "mender.device": not is_user,
        "mender.tenant": tenant_id,
    }
    
    return jwt.encode(
        payload=payload,
        key="secretJWTkey",
        algorithm="HS256"
    )
```

dhaustein (commented):
I skimmed over the migrated tests, they seem to be OK if messy. But refactoring wasn't the point of this PR, so I will just keep them noted for myself to look at later.

Otherwise I have only one concern and that is the custom implementation of JWT.

backend/services/deployments/Makefile
merlin-northern: this is a make target, called during the build, and before the tests start.

dhaustein: So this will be called every time the tests are ran in the CI/CD? Meaning, the API client SDK will be regenerated on every test run.

backend/services/deployments/tests/tests/test_deployment.py
```diff
@@ -421,16 +410,20 @@ def test_device_deployments_already_installed(self, mongo):
...
-                    )
+                    nextdep = {}
+                    try:
+                        dc.get_next_deployment(
+                            dev.fake_token,
+                            artifact_name=artifact_name,
+                            device_type=dev.device_type,
+                        )
+                    except ValueError:
+                        nextdep = None
```
alfrunes: I guess this is one of those instances where the generated client fails to comprehend an empty response. However, one small nitpick, nextdep is not assigned any other value if `get_next_deployment` would return a valid response, perhaps it should be like this?
```suggestion
                    nextdep = {}
                    try:
                        nextdep = dc.get_next_deployment(
                            dev.fake_token,
                            artifact_name=artifact_name,
                            device_type=dev.device_type,
                        )
                    except ValueError:
                        nextdep = None
```

Or like this?
```suggestion
                    with pytest.raises(ValueError):
                        dc.get_next_deployment(
                            dev.fake_token,
                            artifact_name=artifact_name,
                            device_type=dev.device_type,
                        )
                        # assert nextdep == None can be dropped in this case.
```

backend/services/deployments/tests/tests/test_deployment.py
```diff
@@ -527,11 +514,15 @@ def test_device_deployments_full(self, mongo):
...
-                    )
+                    nodep = {}
+                    try:
+                        dc.get_next_deployment(
+                            dev.fake_token,
+                            artifact_name="other {}".format(artifact_name),
+                            device_type=dev.device_type,
+                        )
+                    except ValueError:
+                        nodep = None
```
alfrunes: Same here...

backend/services/deployments/tests/tests/test_release_v2.py
merlin-northern: thanks. this was a copy-paste from other services (should be in common really). this has no impact on the tests, I will perhaps copy-paste from other place.

backend/services/deployments/tests/tests/test_release_v2.py
```diff
@@ -119,39 +124,13 @@ def test_get_all_releases_types(self, mongo, cli):
...
-                + f"/api/management/v2/deployments/releases/all/types"
-            )
-
-            r = requests.get(
-                get_types_url, verify=False, headers={"Authorization": "Bearer foo"},
-            )
-            types = json.loads(r.text)
-            assert types == []
+            types = management_v2_client(jwt="foo").list_release_types()
+            assert len(types) == 0
```
alfrunes, line 136: This test seems to be exclusively in open source only :monocle_face:

backend/services/deployments/tests/tests/test_release_v2.py
merlin-northern: I will try with that.

alfrunes (commented):
I tried to compare the same changes to enterprise to review what actually changed. I know some of my suggestions are predating your changes, but I also found a couple inconsistencies that would be great to clear out before proceeding.

backend/services/deployments/tests/tests/test_release_v2.py
merlin-northern: okey thanks it works.

backend/docs/api/devices_v1_deployments.yaml
merlin-northern: oh, yes, sorry, a left over after linter runs (with warnings on). I added an MSTM agenda.

backend/services/deployments/Makefile
merlin-northern: yes. we had a discussion on it. atm we are not adding the auto-generated code to the repo. feel free to bring it up on Wed.

backend/services/deployments/tests/tests/test_api_internal.py
merlin-northern: this fixture actually is still there. I ported it to the oa3 client (left the requests though for the followup).

backend/services/deployments/tests/tests/client.py
merlin-northern: removed.

backend/services/deployments/tests/tests/test_deployment.py
merlin-northern: let me see what I can do with that lovely piece. and I like the raises approach -- I took it. thank you.

backend/services/deployments/tests/tests/test_deployment.py
merlin-northern: right.

backend/services/deployments/tests/tests/test_release_v2.py
merlin-northern: it was some kind of messup. I will bring it to ent as well. but in a followup PR.

merlin-northern: https://github.com/mendersoftware/mender-server-enterprise/pull/524

merlin-northern:
> I tried to compare the same changes to enterprise to review what actually changed. I know some of my suggestions are predating your changes, but I also found a couple inconsistencies that would be great to clear out before proceeding.

do not worry. thanks for the comments.

backend/docs/api/devices_v1_deployments.yaml
```diff
@@ -180,6 +180,7 @@ paths:
   /download/configuration/{deployment_id}/{device_type}/{device_id}:
     get:
       operationId: Fetch Configuration
+      description: some
```
alfrunes: Not a very helpful description and also not present in enterprise.

backend/tests/requirements-acceptance.txt
```diff
@@ -26,3 +26,5 @@ twisted
 tz
 urllib3<3
 websocket_client >= 0.57
+pydantic==2.11.5
+pyjwt
```
alfrunes: Could you pin that version too so dependabot can track it?

merlin-northern: fixed.

backend/docs/api/devices_v1_deployments.yaml
merlin-northern: fixed. sorry for that I was fighting with the linter on warning level, and got carried away.

backend/services/deployments/Makefile
dhaustein: Thanks! I am a bit concerned, not because I disagree with the auto-generation, but about doing it every time a pipeline runs. 

Like this, we are essentially recreating the contract (the generated API client) between the API and the caller while having no say in it. The API and the tests can diverge and you start getting weird errors in the pipeline. Good luck to anyone who will go debug a pipeline failing like this. 

I'll put it up for Wednesday.

dhaustein (approved):

alfrunes (approved):

PR #742, opened 2025-06-17
MEN-8498 - fix(gui): fixed a problem that could apply RBAC restrictions on non-RBAC plans

mzedel:
the plan check was outdated as it was carried over from the "before plans" time & would be defeated by the check for multitenancy which is always truthy on both hosted & enterprise


mineralsfree (approved):
LGTM

PR #743, opened 2025-06-17
refactor(deployments): Migrate from ant0nie/go-json-rest to gin-gonic/gin

bahaa-ghazal:


backend/services/deployments/api/http/routing.go
```diff
@@ -232,152 +170,151 @@ func NewImagesResourceRoutes(controller *DeploymentsApiHandlers, cfg *Config) []
...
+	// Configuration deployments (internal)
+	router.POST(ApiUrlInternalDeviceConfigurationDeployments,
+		controller.PostDeviceConfigurationDeployment)
+
+	// Last device deployment status deployments (internal)
+	router.POST(ApiUrlInternalDeviceDeploymentLastStatusDeployments,
+		controller.GetDeviceDeploymentLastStatus)
+	// Health Check
+	router.GET(ApiUrlInternalAlive, controller.AliveHandler)
+	router.GET(ApiUrlInternalHealth, controller.HealthHandler)
```
alfrunes: Would be nice if these endpoints (alive and health) would be excluded from the access log if the status code is 2XX as they tend to generate some noise.


```go
accesslogErrorsOnly := accesslog.Middleware{
	DisableLog: func(c *gin.Context) bool {
		if c.Status() / 100 == 2 {
			// 2XX
			return true
		}
		return false
	},
}
```

backend/services/deployments/utils/restutil/options_external_test.go
```diff
@@ -58,130 +54,145 @@ func TestOptionsHandle(t *testing.T) {
...
-
-				if !reflect.DeepEqual(exp.methods, recived.methods) {
-					t.FailNow()
-				}
+			h, ok := v[r.Method]
+			if ok != true {
+				t.Errorf("failed with route %+v, method not present", r)
 			}
 
+			if funcName(r.HandlerFunc) != funcName(h) {
```
alfrunes: Could also compare the function address here instead of using reflection.
```suggestion
			if &r.HandlerFunc != &h {
```

backend/services/deployments/utils/restutil/options_external_test.go
```diff
@@ -58,130 +54,145 @@ func TestOptionsHandle(t *testing.T) {
...
 	}
 }
+
+// We can't compare functions, so let's take the hard way and extract
+// func name from runtime
+func funcName(f interface{}) string {
+	p := reflect.ValueOf(f).Pointer()
+	rfunc := runtime.FuncForPC(p)
+	return rfunc.Name()
+}
```
alfrunes, line 193: This is not very safe. If you pass anything that's not a function it will panic. But since this is only used in a specific test it's ok.

backend/services/deployments/utils/restutil/options_external_test.go
```diff
@@ -58,130 +54,145 @@ func TestOptionsHandle(t *testing.T) {
...
-
-				recived := &MockResponseWriter{}
-				out[idx].Func(recived, nil)
-
-				if !reflect.DeepEqual(exp.methods, recived.methods) {
-					t.FailNow()
-				}
+			h, ok := v[r.Method]
+			if ok != true {
+				t.Errorf("failed with route %+v, method not present", r)
```
alfrunes: If the test fails here it will panic on the next condition which might lead to misleading output.
```suggestion
				t.Errorf("failed with route %+v, method not present", r)
				return
```

alfrunes (approved):
Looks good. :+1:
Found a few nitpicks, but no blockers.

PR #746, opened 2025-06-18
fix(gui): fixed another situation that could cause the organization to overflow

mzedel:
it would happen when expanded too, this should now be addressed:
Screenshot 2025-06-18 at 20 08 16


oldgiova (approved):
I completely trust you

PR #750, opened 2025-06-23
refactor(useradm): Migrate from ant0nie/go-json-rest to gin-gonic/gin

bahaa-ghazal:
Ticket: MEN-8237


backend/pkg/routing/routing.go
alfrunes, line 1: Interesting. Normally this doing a lot of breaking changes and would have to be done separately, but since this package is only used in useradm this is fine.

backend/services/useradm/api/http/api_useradm.go
```diff
@@ -93,359 +108,365 @@ func NewUserAdmApiHandlers(
...
+		userAdm:    userAdm,
+		db:         db,
+		jwth:       jwth,
+		config:     config,
+		authorizer: authorizer,
+	}
+}
+
+// Getting false means that a rest error allready been renderd
+func (i *UserAdmApiHandlers) authTokenExtractor(c *gin.Context) (*jwt.Token, bool) {
```
alfrunes: The boolean return argument is not really needed as the function returns `nil` whenever it responded. There's also the `gin.Context.ResponseWriter.Written()` that reports whether the header is written.
```suggestion
func (i *UserAdmApiHandlers) authTokenExtractor(c *gin.Context) *jwt.Token {
```

backend/services/useradm/api/http/api_useradm.go
```diff
@@ -93,359 +108,365 @@ func NewUserAdmApiHandlers(
...
+		// we have not found the corresponding handler for the key by id
+		// on purpose we do not return common.ErrKeyIdNotFound -- to not allow
+		// the enumeration attack
+		rest.RenderError(c, http.StatusUnauthorized, ErrAuthzTokenInvalid)
+		return nil, false
 	}
-	api.SetApp(app)
 
-	return api.MakeHandler(), nil
+	// parse token, insert into env
```
alfrunes:
```suggestion
	// parse token
```

backend/services/useradm/api/http/api_useradm_test.go
```diff
@@ -1781,27 +1796,33 @@ func TestUserAdmApiGetSettings(t *testing.T) {
...
 	}
 }
 
 func makeReq(method, url, auth string, body interface{}) *http.Request {
-	req := test.MakeSimpleRequest(method, url, body)
-
-	if auth != "" {
-		req.Header.Set("Authorization", auth)
+	if auth == "token" {
+		auth = rtest.DEFAULT_AUTH
```
alfrunes: This is very confusing and arbitrary. Why not just pass `rtest.DEFAULT_AUTH` as an argument?
Had to scroll down here to see what `"token"` in the call signature above meant...

backend/services/useradm/tests/tests/test_auth.py
```diff
@@ -90,9 +90,7 @@ def test_ok(self, api_client_int, api_client_mgmt, init_users):
 
     def test_internal_error(self, api_client_mgmt, init_users):
         try:
-            _, r = api_client_mgmt.logout(
-                auth={"Authorization": "Bearer invalid-token"}
-            )
+            _, r = api_client_mgmt.logout()
```
alfrunes, line 93: Interesting, does this return 500 and not 401? :monocle_face:

backend/services/useradm/api/http/router.go
```diff
@@ -0,0 +1,98 @@
+package http
+
+import (
+	"net/http"
+	"os"
+
+	"github.com/gin-gonic/gin"
```
alfrunes, line 7: The linter wants some "air" here.
```suggestion
	"github.com/gin-gonic/gin"

```

alfrunes (commented):
Generally looks ok.
Just found a few minor things...

backend/services/useradm/api/http/api_useradm.go
bahaa-ghazal: I have done that the first time, but then changed it to be safe

backend/services/useradm/tests/tests/test_auth.py
bahaa-ghazal: It has a token because of this : https://github.com/mendersoftware/mender-server/blob/6a369e5367d22b1ff7dc7a77210239da0769f88b/backend/services/useradm/tests/common.py#L105
It is trying to logout without loging in, which returns 500.

backend/pkg/routing/routing.go
bahaa-ghazal: yeah i have checked before commiting the change

backend/services/useradm/tests/tests/test_auth.py
alfrunes: Fair enough, but based on the documentation I would expect a 401 from this endpoint.
https://github.com/mendersoftware/mender-server/blob/6a369e5367d22b1ff7dc7a77210239da0769f88b/backend/services/useradm/docs/management_api.yml#L79-L107
On the other hand, I think the lack of error handling is primarily because of the assumption of the userauth middleware imposed by the gateway should render a 401 for missing authentication.

alfrunes (approved):

PR #751, opened 2025-06-23
ME-542 - device reports per widget only

mzedel:
This should limit device data retrieval to the visible device reports only + stagger the retrieval to reduce the likelihood of running into rate limits while doing so. Still there might be data retrieval duplication if devices correspond to multiple groups with different selected attributes/ all devices get retrieved with multiple attributes/ ...

Device retrieval is also further limited with this change + points users to contact support for clarification until backend support lands.

### DIFF
--- a/frontend/src/js/components/dashboard/widgets/Distribution.tsx
+++ b/frontend/src/js/components/dashboard/widgets/Distribution.tsx
@@ -12,6 +12,7 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import { useNavigate } from 'react-router-dom';
 
 import { Clear as ClearIcon, Settings, Square } from '@mui/icons-material';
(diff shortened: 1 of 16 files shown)


frontend/src/js/store/devicesSlice/thunks.tsx
```diff
@@ -753,30 +755,43 @@ const initializeDistributionData = (report, groups, devices, totalDeviceCount) =
...
+    return;
+  }
+  const { attribute, group = '', software = '' } = report;
+  const effectiveAttribute = software ? software : attribute;
+  const devicesById = getDevicesById(getState());
+  const acceptedDevices = getAcceptedDevices(getState());
+  const groups = getGroupsById(getState());
+  const { deviceIds = [], filters = [], total = 0 } = groups[group] || {};
+  let groupDevicesRequest = Promise.resolve({
+    payload: { groupName: '', group: { deviceIds: Object.keys(devicesById), total: Object.keys(devicesById).length } }
```
mineralsfree, line 780: It looks like this payload is not used later, was that intentional?

frontend/src/js/components/dashboard/widgets/Distribution.tsx
```diff
@@ -201,28 +205,32 @@ const initDistribution = ({ data, theme }) => {
...
-  const [chartType, setChartType] = useState(chartTypes.bar.key);
-  const [software, setSoftware] = useState('');
-  const [group, setGroup] = useState('');
   const navigate = useNavigate();
   const { classes, theme } = useStyles();
+  const reportsData = useSelector(getDeviceReports);
+  const groupsById = useSelector(getGroupsById);
+  const dispatch = useDispatch();
+  const hasGroupDefinition = !!groupsById[group];
+  const data = reportsData[reportIndex];
```
mineralsfree: maybe this can be named better? For e.g. `report`?

mineralsfree (dismissed):
🤞

frontend/src/js/store/devicesSlice/thunks.tsx
mzedel: It was once 😅 - will clean it up right away!

mineralsfree (approved):

PR #756, opened 2025-06-26
fix(tests): await docker start and remove

dhaustein:
Attempts to reduce flakiness around test setup and `startDockerClient` and `stopDockerClient` where `child_process.spawn` was immediately returning `Promise.resolve()` without waiting for the docker commands to finish

This was causing intermittent fails in the CI
```
stopping: docker
undefined
stderr docker: Error response from daemon: No such container: connect-client
child process exited with code 1
```
Job https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/10456515622

**Summary**
- refactored `startDockerClient` and `stopDockerClient` to use `await execPromise()` and wait for docker

### DIFF
--- a/frontend/tests/e2e_tests/utils/commands.ts
+++ b/frontend/tests/e2e_tests/utils/commands.ts
@@ -13,7 +13,7 @@
 //    limitations under the License.
 import type { APIRequestContext, Browser, BrowserContext, Page } from '@playwright/test';
 import { runServer } from '@sidewinder1138/saml-idp';
-import { spawn } from 'child_process';
+import { exec, spawn } from 'child_process';
 import * as fs from 'fs';
 import { jwtDecode } from 'jwt-decode';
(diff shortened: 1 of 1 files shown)


dhaustein:
Left some TODOs in there that are up to discussion, let me know if you want me to pursue them here or in a separate PR please

mzedel (changes requested):
As said on Slack already, that's neither needed nor the source for failures in the tests... + the commit message with a `fix` type is meant for user facing changes, whereas this is tests/ internal only.

mzedel:
@dhaustein while the discussion is still ongoing you could follow up on the commit type adjustment. We use conventional commits which again bases the commit types on the [Angular guidelines](https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md) - if you're in doubt what type to use.

PR #758, opened 2025-06-26
refactor(useradm): Avoid inconsistencies between OS and enterprise  

bahaa-ghazal:


backend/services/useradm/tests/tests/client.py
```diff
@@ -94,11 +94,16 @@ class ManagementApiClient(ApiClient):
     log = logging.getLogger("client.ManagementClient")
     spec_option = "management_spec"
 
-    # default user auth - single user, single tenant
+    # default user auth
```
bahaa-ghazal, line 97: I chose to keep a instance of a raw jwt-token instead of taking it through the constructor (like the existing implementation in enterprise).

PR #761, opened 2025-06-26
chore: Added security fix to changelog

olehermanse:


olehermanse:
This one is enterprise only, so not the right changelog to put it in.

PR #765, opened 2025-06-27
MEN-7711 - delta artifact generation settings

mzedel:


frontend/src/js/components/settings/ArtifactGeneration.test.tsx
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 Northern.tech AS
```
mineralsfree: probably makes no difference, but it's 2025 :smile:

frontend/src/js/components/settings/ArtifactGeneration.tsx
```diff
@@ -69,157 +73,151 @@ const NumberInputLimited = ({ limit, onChange, value: propsValue, ...remainder }
...
+export const ArtifactGenerationSettings = ({ onClose, open }) => {
+  const { binaryDelta: deltaConfig = {}, binaryDeltaLimits: deltaLimits = {} } = useSelector(state => state.deployments.config) ?? {};
+  const dispatch = useDispatch();
   const { classes } = useStyles();
 
+  const initialValues = useMemo(() => ({ ...formDefaults, ...deltaConfig }), [deltaConfig]);
+
+  const defaultValues = useMemo(
+    () =>
+      Object.entries(deltaLimits).reduce(
```
mineralsfree, line 178: Maybe? 
```
    () => Object.fromEntries(Object.entries(formDefaults).map(([key, value]) => [key, deltaLimits[key] ? deltaLimits[key].default : value])),

```

frontend/src/js/components/settings/ArtifactGeneration.tsx
```diff
@@ -11,52 +11,56 @@
...
   const debouncedValue = useDebounce(value, TIMEOUTS.oneSecond);
   const { default: defaultValue, max, min } = limit;
 
+  useEffect(() => {
+    setValue(propsValue);
+  }, [propsValue]);
+
   useEffect(() => {
     const minimum = Math.max(min, debouncedValue);
     const allowedValue = Math.min(max ?? minimum, minimum);
```
mineralsfree, line 66: In my opinion, this UX with debounce should be reconsidered. I think we should show an error if the value is not allowed, as resetting it is annoying for a user. This is probably out of scope for the task

frontend/src/js/components/settings/Global.tsx
```diff
@@ -183,44 +221,55 @@ export const GlobalSettingsDialog = ({
...
+            title="Deployments confirmation"
+            description="Always require confirmation on deployment creation"
+            onClick={toggleDeploymentConfirmation}
+            value={needsDeploymentConfirmation}
+          />
+        )}
+        {canManageReleases && (
+          <div>
+            <div className="flexbox">
+              <ToggleSetting title="Delta Artifacts generation" disabled={!isEnterprise} onClick={yes} value={hasDeltaArtifactGeneration} />
```
mineralsfree: Maybe it should be always disabled, since user can't change it anyway?

mineralsfree (dismissed):
Few small things, totally non-blocking, just suggestions

frontend/src/js/components/settings/ArtifactGeneration.tsx
mzedel: Yep, fully agree - I guess we'll get to iterate over this with the UX team 🤞

frontend/src/js/components/settings/Global.tsx
mzedel: That would conflict with the design, but OTOH that might just be due to the age difference between the decisions on the task 🙈 and indeed that seem to be more accurate. 👌 adjusted.

mineralsfree (approved):

PR #767, opened 2025-06-27
Fix/tooltip read state toggle

mzedel:


frontend/src/js/store/usersSlice/selectors.ts
```diff
@@ -32,9 +32,16 @@ export const getSelectedDeviceAttribute = createSelector([getUserSettings], ({ c
 );
 export const getIsDarkMode = createSelector([getUserSettings], ({ mode }) => isDarkMode(mode));
 
-export const getShowHelptips = createSelector([getTooltipsById], tooltips =>
-  Object.values(tooltips).reduce((accu, { readState }) => accu || readState === READ_STATES.unread, false)
-);
+export const getReadAllHelptips = createSelector([getTooltipsById], tooltips => {
```
mineralsfree: maybe 
```
export const getReadAllHelptips = createSelector([getTooltipsById], tooltips => {
  const readTooltips = Object.values(tooltips).reduce((accu, { readState }) => (readState === READ_STATES.read ? accu + 1 : accu), 0);
  return readTooltips === Object.keys(tooltips).length;
});

mineralsfree (dismissed):
👍

frontend/src/js/store/usersSlice/selectors.ts
mzedel: > ```
> export const getReadAllHelptips = createSelector([getTooltipsById], tooltips => Object.values(tooltips).every(({ readState }) => (readState === READ_STATES.read)));
> ```

maybe even that?

mineralsfree (approved):

PR #769, opened 2025-06-30
MEN-7744: Rate limits for authenticated requests

alfrunes:
:warning: This PR uses #782

To test the rate limiting, you can use the following compose override which limits all deployments next calls to once per minute per device:
```bash
# Execute in the root of mender-server repository.
cat > docker-compose.override.yml <<EOF
services:
  deviceauth:
    command:
      - --debug
      - server
    environment:
      DEVICEAUTH_REDIS_CONNECTION_STRING: "redis://redis"
      DEVICEAUTH_RATELIMITS_AUTH_ENABLE: "1"
      DEVICEAUTH_RATELIMITS_AUTH_GROUPS: |
        {"name": "slow", "quota": 1, "interval": "1m", "event_expression": "{{ with .Identity }}{{ .Subject }}{{ end }}"}
      DEVICEAUTH_RATELIMITS_AUTH_MATCH: |
        {"api_pattern": "/api/devices/v1/deployments/", "group_expression": "slow"}
        {"api_pattern": "/api/devices/v2/deployments/", "group_expression": "slow"}
        {"api_pattern": "/api/devices/v1/deployments/device/deployments/next", "group_expression": "slow"}
  redis:
    image: redis:7.2

EOF
docker compose build deviceauth
docker compose up -d
```

<details>
<summary>docker-compose.override.yml</summary>

Following is an alternative compose override for configuring using YAML file:

```yaml
services:
  deviceauth:
    command:
      - --debug
      - server
    environment:
      DEVICEAUTH_REDIS_CONNECTION_STRING: "redis://redis"
    configs:
      - source: config.yaml
        target: /etc/deviceauth/config.yaml
  redis:
    image: redis:7.2

configs:
  config.yaml:
    content: |
      ratelimits:
        auth:
          enable: true
          groups:
            - name: slow
              quota: 1
              interval: 1m
              event_expression: "{{with .Identity}}{{.Subject}}{{end}}"
          match:
            - api_pattern: /api/devices/v1/deployments/
              group_expression: "slow"
```

</detalis>

### DIFF
--- a/backend/services/deviceauth/api/http/routing.go
+++ b/backend/services/deviceauth/api/http/routing.go
@@ -76,7 +76,6 @@ func supportsMethod(method string, methods []string) bool {
 // Automatically add OPTIONS method support for each defined route,
 // only if there's no OPTIONS handler for that route yet
 func AutogenOptionsRoutes(router *gin.Engine, gen HttpOptionsGenerator) {
-
 	routes := router.Routes()
 	methodGroups := make(map[string][]string, len(routes))
 
@@ -98,13 +97,28 @@ func AutogenOptionsRoutes(router *gin.Engine, gen HttpOptionsGenerator) {
 			router.OPTIONS(route, gen(methods))
 		}
 	}
-
 }
 
-func NewRouter(app devauth.App, db store.DataStore) http.Handler {
-	router := routing.NewGinRouter()
+type Config struct {
+	AuthVerifyRatelimits gin.HandlerFunc
+}
 
-	d := NewDevAuthApiHandlers(app, db)
+type Option func(c *Config)
+
+func ConfigAuthVerifyRatelimits(handler gin.HandlerFunc) Option {
+	return func(c *Config) {
+		c.AuthVerifyRatelimits = handler
+	}
+}
+
+func NewRouter(app devauth.App, db store.DataStore, options ...Option) http.Handler {
+	router := routing.NewGinRouter()
+	cfg := new(Config)
+	for _, option := range options {
+		option(cfg)
+	}
+
+	d := NewDevAuthApiHandlers(app, db, options...)
 
 	publicAPIs := router.Group(".")
 	publicAPIs.Use(identity.Middleware())
@@ -137,12 +151,11 @@ func NewRouter(app devauth.App, db store.DataStore) http.Handler {
 
 	intrnlAPIV1.GET(uriAlive, d.AliveHandler)
 	intrnlAPIV1.GET(uriHealth, d.HealthCheckHandler)
-	intrnlAPIV1.GET(uriTokenVerify,
-		identity.Middleware(),
-		d.VerifyTokenHandler)
-	intrnlAPIV1.POST(uriTokenVerify,
-		identity.Middleware(),
-		d.VerifyTokenHandler)
+
+	intrnlAPIV1.Group(".").
+		Use(identity.Middleware()).
+		GET(uriTokenVerify, d.VerifyTokenHandler).
+		POST(uriTokenVerify, d.VerifyTokenHandler)
 	intrnlAPIV1.DELETE(uriTokens, d.DeleteTokensHandler)
 	intrnlAPIV1.PUT(uriTenantLimit, d.PutTenantLimitHandler)
 	intrnlAPIV1.GET(uriTenantLimit, d.GetTenantLimitHandler)
--- a/backend/services/deviceauth/config.yaml
+++ b/backend/services/deviceauth/config.yaml
@@ -147,8 +147,48 @@
 # Overwrite with environment variable: DEVICEAUTH_REDIS_LIMITS_EXPIRE_SEC
 # redis_limits_expire_sec: "1800"
 
-
 #    Enable addon feature restrictions.
 #    Defaults to: false
 #    Overwrite with environment variable: DEVICEAUTH_HAVE_ADDONS
 # have_addons: false
+
+# ratelimits:
+#   # auth configures ratelimits for authenticated requests.
+#   auth:
+#     # enable rate limiting also requires redis_connection_string to be effective.
+#     enable: false
+#     # default rate limit group is the default ratelimiting parameters used when
+#     # no group `match`es the expressions.
+#     default:
+#       quota: 100    # number of request per
+#       interval: 60s # interval
+#       # event_expression is a go template for grouping requests.
+#       # The following attributes are available in the context:
+#       # Identity - contains a subset of the JWT claims:
+#         - .Subject  (jwt:"sub")          string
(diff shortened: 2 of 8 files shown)


alfrunes:
:warning: Will rebase this PR, I'm also creating a new PR with the common stuff to lower the review bar. :barber:

alfrunes:
To try to aid the review experience I put all the common parts (`pkg` stuff) in a separate PR: #782

backend/services/deviceauth/api/http/routing.go
```diff
@@ -112,12 +110,29 @@ func init() {
...
-func NewRouter(app devauth.App, db store.DataStore) http.Handler {
+type Config struct {
+	AuthVerifyRatelimits gin.HandlerFunc
+}
+
+type Option func(c *Config)
+
+func ConfigAuthVerifyRatelimits(handler gin.HandlerFunc) Option {
+	return func(c *Config) {
+		c.AuthVerifyRatelimits = handler
```
merlin-northern, line 124: what if c is nil?

backend/services/deviceauth/server.go
```diff
@@ -108,25 +113,37 @@ func RunServer(c config.Reader) error {
 	if cacheConnStr != "" {
 		l.Infof("setting up redis cache")
 
-		cache, err := cache.NewRedisCache(
-			context.TODO(),
-			cacheConnStr,
-			c.GetString(dconfig.SettingRedisKeyPrefix),
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		redisClient, err := redis.ClientFromConnectionString(ctx, cacheConnStr)
+		cancel()
```
merlin-northern: why do we run cancel() here?

merlin-northern (dismissed):
not bad!

backend/services/deviceauth/server.go
alfrunes: I was setting up a trap :mouse_trap: :imp: 
Just kidding, but it's sloppy work. I refactored it to a separate function and defer the cancel.

backend/services/deviceauth/api/http/routing.go
alfrunes: A new config is always allocated 5 lines below before calling any Option:
https://github.com/mendersoftware/mender-server/blob/d68fa42c79c3dbfc6a39314a370ae6aeb1b1e43e/backend/services/deviceauth/api/http/routing.go#L126

alfrunes:
Thanks to your observation I fixed a bug in this PR https://github.com/mendersoftware/mender-server/pull/772#discussion_r2201136830
I also did a small refactoring to handle the poor use of the Context.

backend/services/deviceauth/api/http/routing.go
```diff
@@ -150,12 +165,11 @@ func NewRouter(app devauth.App, db store.DataStore) http.Handler {
...
 	intrnlAPIV1.GET(uriAlive, d.AliveHandler)
 	intrnlAPIV1.GET(uriHealth, d.HealthCheckHandler)
-	intrnlAPIV1.GET(uriTokenVerify,
-		identity.Middleware(),
-		d.VerifyTokenHandler)
-	intrnlAPIV1.POST(uriTokenVerify,
-		identity.Middleware(),
-		d.VerifyTokenHandler)
+
+	intrnlAPIV1.Group(".").
```
merlin-northern, line 171: this will not add "./" right? (this is just paranoid_mode:on)

backend/services/deviceauth/server.go
```diff
@@ -100,33 +104,27 @@ func RunServer(c config.Reader) error {
...
-			c.GetInt(dconfig.SettingRedisLimitsExpSec),
-		)
-
+		srvCache, rateLimits, err := setupRedis(c, cacheConnStr)
 		if err != nil {
 			return err
 		}
-
-		devauth = devauth.WithCache(cache)
+		devauth.WithCache(srvCache)
```
merlin-northern: just and observation: we always use it with assignment `devauth = `, so technically this should be SetCache. okey, ignore this.

merlin-northern (dismissed):
you are doing great job with all that.

backend/services/deviceauth/server.go
```diff
@@ -160,3 +158,30 @@ func RunServer(c config.Reader) error {
 	}
 	return nil
 }
+
+func setupRedis(c config.Reader, connStr string) (cache.Cache, *rate.HTTPLimiter, error) {
+	l := log.NewEmpty()
```
merlin-northern, line 165: technically this one is called where a logger instance already exists.

backend/services/deviceauth/server.go
```diff
@@ -160,3 +158,30 @@ func RunServer(c config.Reader) error {
...
+	l.Infof("setting up redis cache")
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
```
merlin-northern, line 169: will this cancel the context right after setup? do we need this context after this call?

alfrunes: The context only exists in the scope of the setup function, so we call cancel when leaving the scope to make sure the context deadline is terminated.

backend/services/deviceauth/api/http/routing.go
alfrunes: No, it will expand relative paths effectively removing the dot.

backend/services/deviceauth/server.go
alfrunes: The two are the same since the `With` just returns an instance of itself. However, that diff was unnecessary so I reverted it back to how it was.

backend/services/deviceauth/config.yaml
```diff
@@ -147,12 +147,52 @@
...
-
 #    Enable addon feature restrictions.
 #    Defaults to: false
 #    Overwrite with environment variable: DEVICEAUTH_HAVE_ADDONS
 # have_addons: false
 
+# ratelimits:
+#   # auth configures ratelimits for authenticated requests.
+#   auth:
+#     # enable rate limiting also requires redis_connection_string to be effective.
```
merlin-northern, line 158:
```suggestion
#     # enable rate limiting also requires redis_connection_string to be in effect.
```

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -1380,7 +1380,7 @@ func (d *DevAuth) cacheThrottleVerify(
 	origMethod,
 	origUri string,
 ) (string, error) {
-	if d.cache == nil {
+	if d.cache == nil || d.cTenant == nil {
```
merlin-northern, line 1383: it will also panic if `token` is nil.

merlin-northern (approved):
looks good. you could rebase on top of main.

PR #770, opened 2025-06-30
docs(deployments): Create API spec for v2 GET `artifacts/list` endpoint

bahaa-ghazal:
Ticket: MEN-8180

### DIFF
--- a/backend/docs/api/management_v2_deployments.yaml
+++ b/backend/docs/api/management_v2_deployments.yaml
@@ -571,6 +571,97 @@ paths:
         Lists all release update types.
       tags:
       - Management API
+  /artifacts:
+    get:
+      description: |
+        Returns a collection of all artifacts sorted by name.
+      operationId: List Artifacts with pagination
+      parameters:
+      - description: |
+          Artifact(s) name(s) filter.
+          Multiple names can be provided (e.g., `?name=foo&name=bar`).
+          Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+        in: query
+        name: name
+        schema:
+          type: array
+          items:
+            type: string
+        style: form
+        explode: true
+        examples:
+          single:
+            summary: Get artifact with the given name.
+            value:
+              - foo
+          multi:
+            summary: Get artifact matching multiple names by repeating the query parameter.
+            value:
+              - foo
+              - bar
+          prefix:
+            summary: Get artifact with the specified prefix
+            value:
+              - foo*
+      - description: |
+          Artifact description filter.
+          Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+        in: query
+        name: description
+        schema:
+          type: string
+      - description: |
+          Artifact device type filter.
+          Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+        in: query
+        name: device_type
+        schema:
+          type: string
+      - description: Starting page.
+        in: query
+        name: page
+        schema:
+          default: 1
+          type: integer
+      - description: Maximum number of results per page.
+        in: query
+        name: per_page
+        schema:
+          default: 20
+          maximum: 500
+          type: integer
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                items:
+                  $ref: '#/components/schemas/Artifact'
+                type: array
+          description: OK
+          headers:
+            Link:
+              description: "Standard header, we support 'first', 'next', and 'prev'."
+              schema:
+                type: string
+        "401":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Unauthorized.
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Internal Server Error.
+      security:
+      - ManagementJWT: []
+      summary: |
+        Lists known artifacts.
+      tags:
+      - Management API
 components:
   responses:
     InvalidRequestError:
--- a/backend/services/deployments/docs/management_api_v2.yml
+++ b/backend/services/deployments/docs/management_api_v2.yml
@@ -522,6 +522,75 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /artifacts:
+    get:
+      operationId: List Artifacts with pagination
+      tags:
+      - Management API
+      security:
+      - ManagementJWT: []
+      summary: |
+        Lists known artifacts.
+      description: |
+        Returns a collection of all artifacts sorted by name.
+      parameters:
+        - name: name
+          in: query
+          description: | 
+            Artifact(s) name(s) filter.
+            Multiple names can be provided (e.g., `?name=foo&name=bar`).
+            Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+          required: false
+          type: array
+          items:
+            type: string
+          collectionFormat: multi
+        - name: description
+          in: query
+          description: |
+            Artifact description filter.
+            Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+          required: false
+          type: string
+        - name: device_type
+          in: query
+          description: |
+            Artifact device type filter.
+            Supports exact matching or prefix matching by adding `*` to the end (e.g., `foo*`).
+          required: false
+          type: string
+        - name: page
+          in: query
+          description: Starting page.
+          required: false
+          type: number
+          format: integer
+          default: 1
+        - name: per_page
+          in: query
+          description: Maximum number of results per page.
+          required: false
+          type: number
+          format: integer
+          default: 20
+          maximum: 500
+      produces:
+        - application/json
+      responses:
+        200:
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/Artifact"
+          headers:
+            Link:
+              type: string
+              description: Standard header, we support 'first', 'next', and 'prev'.
+        401:
+          $ref: '#/responses/UnauthorizedError'
+        500:
+          $ref: "#/responses/InternalServerError"
 
 definitions:
   Artifact:


backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,94 @@ paths:
...
+          Artifact(s) exact name(s) or prefix filter. 
+          Add `*` to the end for prefix matching.
+        in: query
+        name: name
+        schema:
+          type: array
+          items:
+            type: string
+        style: form
+        explode: true
```
merlin-northern: Screenshot 2025-06-30 at 11 25 07
is it only me or does it look strange a bit?

merlin-northern: perhaps it is only the github render it strangely. 🤔

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,94 @@ paths:
...
+      parameters:
+      - description: | 
+          Artifact(s) exact name(s) or prefix filter. 
+          Add `*` to the end for prefix matching.
+        in: query
+        name: name
+        schema:
+          type: array
+          items:
+            type: string
```
merlin-northern, line 589: you know reading that, and assuming I know nothing, which is not far from what I do know, I have no idea how to pass an array of strings/prefixes in a query. example would be nice.

backend/docs/api/management_v2_deployments.yaml
bahaa-ghazal: This shows that you can pass multiple name prams at ones, like `?name="foo"&name="bar"`. The `style` and `explode` is just is defaults values, but are there to make things clearer

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,94 @@ paths:
...
+      - description: Artifact device type filter.
+        in: query
+        name: device_type
+        schema:
+          type: string
+      - description: Starting page.
+        in: query
+        name: page
+        schema:
+          default: 1.0
```
merlin-northern: it does not smell of integer to me.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,94 @@ paths:
...
+          type: integer
+      - description: Maximum number of results per page.
+        in: query
+        name: per_page
+        schema:
+          default: 20.0
```
merlin-northern: same story. it will break the generation.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,94 @@ paths:
...
+                  $ref: '#/components/schemas/Artifact'
+                type: array
+          description: OK
+          headers:
+            X-Total-Count:
+              description: Total number of releases matching query.
+              schema:
+                type: integer
+            Link:
+              description: "Standard header, we support 'first', 'next', and 'prev'."
```
merlin-northern, line 644: why is this one in double quotes? I would see some more explicit information here.

backend/docs/api/management_v2_deployments.yaml
merlin-northern: so, just reading that there is a parameter in query of type array and that the items are strings I would never guess that I can specify name multiple times. this is how we described similar situation: https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/services/deployments/docs/management_api_v2.yml#L77 it is important to be clear you know especially that this is something that works a bit outside the RFC for HTTP.

backend/docs/api/management_v2_deployments.yaml
bahaa-ghazal: I copied the pagination and sorting from the existing v1 `artifacts/list`. Guess they could use some work

bahaa-ghazal: All the other endpoints that uses pagination use this same description. I'll keep it to keep the consistency

merlin-northern: okey, but I find it confusing.

backend/docs/api/management_v2_deployments.yaml
alfrunes: The `form` and `style` parameters should be sufficient to describe how the parameter is defined. The documentation generator should render the examples (below) in the style specified.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
+              - foo
+              - bar
+          Prefix Search:
+            value:
+              - foo*
+      - description: Artifact exact description filter.
+        in: query
+        name: description
+        schema:
+          type: string
```
alfrunes: I do not see the usefulness of this query parameter.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
+        in: query
+        name: sort
+        schema:
+          default: name:asc
+          enum:
+          - name:asc
+          - name:desc
+          - modified:asc
+          - modified:desc
+          type: string
```
alfrunes: I suppose this is from the ticket, but I would argue that sorting by created (oldest first) should be the default as it's the only sorting criteria that produce stable pagination outputs.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
+        "500":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          description: Internal Server Error.
+      security:
+      - ManagementJWT: []
+      summary: |
+        Lists known actifacts.
```
alfrunes:
```suggestion
        Lists known artifacts.
```

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
+              schema:
+                items:
+                  $ref: '#/components/schemas/Artifact'
+                type: array
+          description: OK
+          headers:
+            X-Total-Count:
+              description: Total number of releases matching query.
+              schema:
+                type: integer
```
alfrunes: @kjaskiewiczz @merlin-northern; Are we still doing this?

backend/docs/api/management_v2_deployments.yaml
alfrunes: If we go through the effort of getting the total count we could also have a reference to the last page, although I would prefer not to.

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
+          Single Name:
+            value:
+              - foo
+          Multiple Names:
+            value:
+              - foo
+              - bar
+          Prefix Search:
+            value:
+              - foo*
```
alfrunes: A summary would be nice :)
```suggestion
          single:
            summary: Get release with the given name.
            value:
              - foo
          multi:
            summary: Get releases matching multiple names by repeating the query parameter.
            value:
              - foo
              - bar
          prefix:
            summary: Get releases with the specified prefix
            value:
              - foo*
```

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
...
       - Management API
+  /artifacts/list:
+    get:
+      description: |
+        Returns a collection of all artifacts.
+      operationId: List Artifacts with pagination
+      parameters:
+      - description: | 
+          Artifact(s) exact name(s) or prefix filter. 
+          Add `*` to the end for prefix matching.
```
alfrunes: Just be mindful that the actual wildcard character is `%2A` in the escaped query string (the only non-alphanumeric characters that are not escaped are `-` `_` `.` `~`). Not sure if that could lead to confusion. The implementation also needs an escape sequence for the literal escape character (e.g. `\*`).

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,105 @@ paths:
         Lists all release update types.
       tags:
       - Management API
+  /artifacts/list:
```
alfrunes: Can we make the `artifacts` RESTful again? :upside_down_face:
```suggestion
  /artifacts:
```

alfrunes (commented):
Great job creating your first API specs! :rocket:
The specs look pretty good to me, I only have a few nitpicks here and there. I also have a few questions regarding the acceptance criteria for the endpoint itself, but perhaps we can start with the minimum set of requirements and build from there?
@kjaskiewiczz would be great if you could also take a look.

backend/docs/api/management_v2_deployments.yaml
kjaskiewiczz: +1

backend/docs/api/management_v2_deployments.yaml
kjaskiewiczz: I agree with @merlin-northern. I know `form` and `style` should be enough, but there is no harm in adding extra line of description to avoid confusion.

backend/docs/api/management_v2_deployments.yaml
kjaskiewiczz: Me neither, but this was direct request from a customer. UI is not consuming the old endpoint and will not consume this one, so we prepare a new endpoint to satisfy customers using the old endpoint directly and we want to fully cover their use case.
If they are using the old endpoint this way (or similar way, but the intention was to exact match the description) let them do this with a new endpoint.
ref: https://github.com/mendersoftware/deployments-enterprise/commit/c48b8f6fa44f6cb54286d4a73aae786fffee2b59

backend/docs/api/management_v2_deployments.yaml
```diff
@@ -571,6 +571,108 @@ paths:
...
+            value:
+              - foo
+              - bar
+          prefix:
+            summary: Get artifact with the specified prefix
+            value:
+              - foo\*
+      - description: Artifact exact description filter.
+        in: query
+        name: description
```
kjaskiewiczz, line 610: Since with the old endpoint we allow regex search for name, description and device_type, I'd allow exact and prefix search for them in the new endpoint (would do the same to description and device_type as you did for name).

backend/docs/api/management_v2_deployments.yaml
kjaskiewiczz: Yes, if it's part of requirements/UX design. Here we don't need it.

backend/docs/api/management_v2_deployments.yaml
kjaskiewiczz: +1
I'd set modified:asc as default

kjaskiewiczz (changes requested):

backend/docs/api/management_v2_deployments.yaml
alfrunes: Sorry about the confusion, but I'm reading a contradiction: first part says yes, the other part says we don't need it (no). Could you please rephrase that? :upside_down_face:

bahaa-ghazal: I think what he meant is that the endpoint isn't used by the frontend and is only used by customers calling the endpoint directly.

kjaskiewiczz: yes, if it's part of requirements/UX design, here it isn't

kjaskiewiczz: or in other words - we try to avoid that, but if it's enforced through explicit requirement, we do this

kjaskiewiczz (dismissed):

alfrunes (approved):

PR #771, opened 2025-06-30
revert: Deviceauth ratelimiting for authenticated requests (MEN-7744)

alfrunes:
This PR reverts the ratelimiting feature implemented for authenticated deviceauth requests in favor of a new implementation that satisfies the updated acceptance criteira.


merlin-northern (approved):
only reverts, right? I totally trust you. what is the scenario to get it to ent?

PR #773, opened 2025-07-01
fix(deviceauth): Fix the pagination logic in devices search endpoint

bahaa-ghazal:
Ticket: MEN-8521
Changelog: Title

### DIFF
--- a/backend/services/deviceauth/api/http/api_devauth.go
+++ b/backend/services/deviceauth/api/http/api_devauth.go
@@ -237,17 +237,18 @@ func (i *DevAuthApiHandlers) SearchDevicesV2Handler(c *gin.Context) {
 		return
 	}
 
-	numDevs := len(devs)
+	numDevs := int64(len(devs))
 	hasNext := false
-	if int64(numDevs) > perPage {
+
+	if numDevs > perPage {
 		hasNext = true
-		numDevs = int(perPage)
+		numDevs -= 1
 	}
+
 	hints := rest.NewPagingHints().
 		SetPage(page).
 		SetPerPage(perPage).
-		SetHasNext(hasNext).
-		SetTotalCount(int64(numDevs))
+		SetHasNext(hasNext)
 	links, err := rest.MakePagingHeaders(c.Request, hints)
 	if err != nil {
 		rest.RenderError(c, http.StatusBadRequest, err)
--- a/backend/services/deviceauth/api/http/api_devauth_test.go
+++ b/backend/services/deviceauth/api/http/api_devauth_test.go
@@ -1009,6 +1009,57 @@ func TestSearchDevices(t *testing.T) {
 			return b
 		}(),
 	}, {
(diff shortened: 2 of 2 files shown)


backend/services/deviceauth/api/http/api_devauth_test.go
```diff
@@ -1223,7 +1279,7 @@ func TestApiV2GetDevices(t *testing.T) {
 				Path:   "http://localhost/api/management/v2/devauth/devices?page=2&per_page=2",
 				Auth:   true,
 			}),
-			devices: devs,
+			devices: devs[:2],
```
bahaa-ghazal, line 1277: This test actually does not test the limits because it is handled in `GetDevices`, which is being mocked

backend/services/deviceauth/api/http/api_devauth_test.go
```diff
@@ -2188,11 +2244,11 @@ func TestApiGetTenantDevicesV2(t *testing.T) {
 				Method: "GET",
 				Path:   "http://localhost/api/internal/v1/devauth/tenants/powerpuff123/devices?page=2&per_page=2",
 			}),
-			devices: devs,
+			devices: devs[:2],
```
bahaa-ghazal, line 2242: same here

backend/services/deviceauth/api/http/api_devauth.go
```diff
@@ -237,17 +237,14 @@ func (i *DevAuthApiHandlers) SearchDevicesV2Handler(c *gin.Context) {
...
-	}
+	numDevs := int64(len(devs))
+	hasNext := numDevs > perPage
+
 	hints := rest.NewPagingHints().
 		SetPage(page).
 		SetPerPage(perPage).
 		SetHasNext(hasNext).
-		SetTotalCount(int64(numDevs))
+		SetTotalCount(numDevs)
```
alfrunes: This line needs to be removed as it is not the actual total count (see above). We only know if it has one device more than current page.
```suggestion
```

alfrunes (changes requested):

backend/services/deviceauth/api/http/api_devauth.go
```diff
@@ -237,17 +237,13 @@ func (i *DevAuthApiHandlers) SearchDevicesV2Handler(c *gin.Context) {
...
 
-	numDevs := len(devs)
-	hasNext := false
-	if int64(numDevs) > perPage {
-		hasNext = true
-		numDevs = int(perPage)
-	}
+	numDevs := int64(len(devs))
+	hasNext := numDevs > perPage
+
```
alfrunes, line 247: We're returning one more device than requested (hence the test failures).
```suggestion
	if hasMore {
		numDevs---
	}
```

alfrunes (changes requested):

alfrunes (approved):

PR #781, opened 2025-07-02
Created interface and implementation of distributed locks

alfrunes:
... and a MongoDB implementation.

### DIFF
--- a/backend/pkg/mongo/lock.go
+++ b/backend/pkg/mongo/lock.go
@@ -1,0 +1,86 @@
+package mongo
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/writeconcern"
+
+	"github.com/mendersoftware/mender-server/pkg/sync"
+)
+
+type distributedLock struct {
+	coll   *mongo.Collection
+	lockID string
+	lock   lockDoc
+}
+
+func NewLockGenerator(db *mongo.Database, collName string) sync.DistributedLockGenerator {
+	return func(lockID string) (sync.DistributedLock, error) {
+		return NewLock(db, collName, lockID), nil
+	}
+}
+
+// NewLock creates a DistributedLock using a single document in a collection.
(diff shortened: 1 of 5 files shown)


backend/pkg/mongo/lock.go
```diff
@@ -0,0 +1,86 @@
...
+	var lock lockDoc
+	err := l.coll.FindOneAndUpdate(ctx,
+		bson.M{"_id": l.lockID, "exp": bson.M{"$lt": now}},
+		bson.M{
+			"$set": bson.M{"_id": l.lockID, "exp": deadline, "created": now},
+			"$inc": bson.M{"version": 1},
+		},
+		options.FindOneAndUpdate().
+			SetReturnDocument(options.After).
+			SetUpsert(true)).
```
merlin-northern, line 72: do we plan to remove the documents after a couple of years?

backend/pkg/mongo/lock.go
```diff
@@ -0,0 +1,86 @@
...
+	if !ok {
+		deadline = time.Now().Add(defaultTTL)
+	}
+	now := time.Now()
+	var lock lockDoc
+	err := l.coll.FindOneAndUpdate(ctx,
+		bson.M{"_id": l.lockID, "exp": bson.M{"$lt": now}},
+		bson.M{
+			"$set": bson.M{"_id": l.lockID, "exp": deadline, "created": now},
+			"$inc": bson.M{"version": 1},
```
merlin-northern: there is ``bson:"v"`` above. also: where do we use the version/v? also: at one point any numerical int will overflow, do we need to worry about it here?

alfrunes: Right that would miss the point of the document versioning.
I don't think there will be any issues with int overflows (even if it were to happen). We're using the $inc simply to guarantee uniqueness for the caller that aquires the lock.

backend/pkg/mongo/lock.go
alfrunes: I don't believe there will be a problem with too many documents (there will be one per resource in locked state). We can make a trivial TTL index on the `exp` field, but I don't want to implement a breaking change just to create the index.

merlin-northern: stale locks removal index sounds good.

merlin-northern (approved):
looks good to me

PR #772, opened 2025-06-30
MEN-7745: Add rate limiting support to authenticated useradm requests

alfrunes:
:warning: NOTE: The 9 first commits are this separate PR: #782 
Please review first before proceeding with this one.

Example override for testing this feature:
```yaml
# docker-compose.override.yml
services:
  useradm:
    command:
      - --debug
      - server
      - --automigrate
    image: localhost:5000/useradm:MEN-7455
    environment:
      USERADM_REDIS_CONNECTION_STRING: "redis://redis"
      USERADM_RATELIMITS_AUTH_ENABLE: "1"
      USERADM_RATELIMITS_AUTH_GROUPS: |
        {"name": "slow", "quota": 1, "interval": "1m", "event_expression": "{{ with .Identity }}{{ .Subject }}{{ end }}"}
      USERADM_RATELIMITS_AUTH_MATCH: |
        {"api_pattern": "/api/management/v1/useradm/slow", "group_expression": "slow"}
        {"api_pattern": "/api/management/v1/useradm/slow/", "group_expression": "slow"}
  redis:
    image: redis:7.2
```
Any request to sub-paths of `/api/management/v1/useradm/slow` will be limited to 1 request per second.

PR #782, opened 2025-07-02
MEN-7743: Distributed authentication rate limits (common parts)

alfrunes:
This PR contains the common parts for useradm and deviceauth for creating distributed ratelimits for authenticated requests (for token verification). The PR contains the following components:
* Extending the existing rate limiting interface with `EventLimiter`
* A new HTTPLimiter implementation
  * Maps request patterns to different rate limiting instances
    * Allows independent rate limiting (with different parameters) for different API groups
  * Is currently only using the existing Redis-based fixed-window rate limiting algorithm 
* A shared configuration component that takes care of parsing the configuration for ratelimits

### DIFF
--- a/backend/pkg/config/ratelimits/config.go
+++ b/backend/pkg/config/ratelimits/config.go
@@ -1,0 +1,128 @@
+// Copyright 2025 Northern.tech AS
+//
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
+//
+//	    http://www.apache.org/licenses/LICENSE-2.0
+//
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
+
+package ratelimits
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/mendersoftware/mender-server/pkg/config"
+	"github.com/mendersoftware/mender-server/pkg/log"
+	"github.com/mendersoftware/mender-server/pkg/rate"
+	"github.com/mendersoftware/mender-server/pkg/redis"
+)
+
+type ConfigDisabledError struct {
+	Path string
+}
+
+func (err *ConfigDisabledError) Error() string {
+	return `configuration "` + err.Path + `" disabled`
+}
+
+func init() {
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultInterval, "1m")
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultQuota, "120")
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultEventExpression,
+		"{{with .Identity}}{{.Subject}}{{end}}")
+}
+
+const (
+	SettingRatelimits                           = "ratelimits"
+	SettingRatelimitsAuth                       = SettingRatelimits + ".auth"
+	SettingRatelimitsAuthEnable                 = SettingRatelimitsAuth + ".enable"
+	SettingRatelimitsAuthGroups                 = SettingRatelimitsAuth + ".groups"
+	SettingRatelimitsAuthMatch                  = SettingRatelimitsAuth + ".match"
+	SettingRatelimitsAuthDefault                = SettingRatelimitsAuth + ".default"
+	SettingRatelimitsAuthDefaultQuota           = SettingRatelimitsAuthDefault + ".quota"
+	SettingRatelimitsAuthDefaultInterval        = SettingRatelimitsAuthDefault + ".interval"
+	SettingRatelimitsAuthDefaultEventExpression = SettingRatelimitsAuthDefault +
+		".event_expression"
+)
+
+func LoadRatelimits(c config.Reader) (*RatelimitConfig, error) {
+	if !c.GetBool(SettingRatelimitsAuthEnable) {
+		return nil, nil
+	}
+	ratelimitConfig := RatelimitConfig{
+		DefaultGroup: RatelimitParams{
+			Quota:           int64(c.GetInt(SettingRatelimitsAuthDefaultQuota)),
+			Interval:        config.Duration(c.GetDuration(SettingRatelimitsAuthDefaultInterval)),
+			EventExpression: c.GetString(SettingRatelimitsAuthDefaultEventExpression),
+		},
+	}
+	err := config.UnmarshalSliceSetting(c,
+		SettingRatelimitsAuthGroups,
+		&ratelimitConfig.RatelimitGroups,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error loading rate limit groups: %w", err)
+	}
+
+	err = config.UnmarshalSliceSetting(c,
+		SettingRatelimitsAuthMatch,
+		&ratelimitConfig.MatchExpressions,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error loading rate limit match expressions: %w", err)
+	}
+	return &ratelimitConfig, nil
+}
+
+func SetupRedisRateLimits(
+	redisClient redis.Client,
+	keyPrefix string,
+	c config.Reader,
+) (*rate.HTTPLimiter, error) {
+	if !c.GetBool(SettingRatelimitsAuthEnable) {
+		return nil, &ConfigDisabledError{
+			Path: SettingRatelimitsAuthEnable,
+		}
+	}
+	lims, err := LoadRatelimits(c)
+	if err != nil {
+		return nil, err
+	}
+	log.NewEmpty().Debugf("loaded rate limit configuration: %v", lims)
+	defaultPrefix := fmt.Sprintf("%s:rate:default", keyPrefix)
+	defaultLimiter := redis.NewFixedWindowRateLimiter(
+		redisClient, defaultPrefix,
+		time.Duration(lims.DefaultGroup.Interval), lims.DefaultGroup.Quota,
+	)
+	mux, err := rate.NewHTTPLimiter(
+		defaultLimiter,
+		c.GetString(SettingRatelimitsAuthDefaultEventExpression),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error setting up rate limits: %w", err)
+	}
+	for _, group := range lims.RatelimitGroups {
+		groupPrefix := fmt.Sprintf("%s:rate:g:%s", keyPrefix, group.Name)
+		limiter := redis.NewFixedWindowRateLimiter(
+			redisClient, groupPrefix, time.Duration(group.Interval), group.Quota,
+		)
+		err = mux.AddRateLimitGroup(limiter, group.Name, group.EventExpression)
+		if err != nil {
+			return nil, fmt.Errorf("error setting up rate limit group %s: %w", group.Name, err)
+		}
+	}
+	for _, expr := range lims.MatchExpressions {
+		err = mux.AddMatchExpression(expr.APIPattern, expr.GroupExpression)
+		if err != nil {
+			return nil, fmt.Errorf("error setting up match patterns: %w", err)
+		}
+	}
+	return mux, nil
+}
--- a/backend/pkg/config/unmarshal.go
+++ b/backend/pkg/config/unmarshal.go
@@ -19,10 +19,22 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/pkg/errors"
 )
+
+type Duration time.Duration
+
+func (duration *Duration) UnmarshalText(b []byte) error {
+	d, err := time.ParseDuration(string(b))
+	if err != nil {
+		return err
+	}
+	*duration = Duration(d)
+	return nil
+}
 
 // UnmarshalSliceSetting will unmarshal an array of objects into the result T
 // using either newline separated json objects for strings (from env) or
--- a/backend/pkg/config/unmarshal.go
+++ b/backend/pkg/config/unmarshal.go
@@ -1,0 +1,74 @@
+// Copyright 2025 Northern.tech AS
+//
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
+//
+//	    http://www.apache.org/licenses/LICENSE-2.0
+//
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
+
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/go-viper/mapstructure/v2"
+	"github.com/pkg/errors"
+)
+
+// UnmarshalSliceSetting will unmarshal an array of objects into the result T
+// using either newline separated json objects for strings (from env) or
+// an array of objects (using json struct tag) for any configuration source
+// that is able to express a slice of objects.
+func UnmarshalSliceSetting[T any](c Reader, path string, result *[]T) error {
+	if result == nil {
+		return errors.New("cannot unmarshal setting to nil result")
+	}
+	value := c.Get(path)
+	var err error
+	switch cfg := value.(type) {
+	case string:
+		decoder := json.NewDecoder(strings.NewReader(cfg))
+		for {
+			var elem T
+			err = decoder.Decode(&elem)
+			if err != nil {
+				break
+			}
+			*result = append(*result, elem)
+		}
+		if errors.Is(err, io.EOF) {
+			err = nil
+		}
+	case []any:
+		var decoder *mapstructure.Decoder
+		decoder, err = mapstructure.NewDecoder(
+			&mapstructure.DecoderConfig{
+				TagName: "json",
+				Result:  &result,
+				Squash:  true,
+				DecodeHook: mapstructure.ComposeDecodeHookFunc(
+					mapstructure.TextUnmarshallerHookFunc(),
+					mapstructure.StringToTimeDurationHookFunc(),
+				),
+			},
+		)
+		if err != nil {
+			return err
+		}
+		err = decoder.Decode(value)
+	case nil:
+		// pass (empty config)
+	default:
+		err = fmt.Errorf("invalid config type %T", cfg)
+	}
+	return err
+}
--- a/backend/pkg/config/unmarshal_test.go
+++ b/backend/pkg/config/unmarshal_test.go
@@ -1,0 +1,187 @@
+// Copyright 2025 Northern.tech AS
+//
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
+//
+//	    http://www.apache.org/licenses/LICENSE-2.0
+//
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
+
+package config
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/go-viper/mapstructure/v2"
+	"github.com/spf13/viper"
+)
+
+type testConfig struct {
+	Foo string   `json:"foo"`
+	Bar float64  `json:"bar"`
+	Dur Duration `json:"dur"`
+}
+
+func TestUnmarshalSliceSetting(t *testing.T) {
+	configData := []testConfig{
+		{Foo: "first_foo", Bar: 9.1, Dur: Duration(time.Nanosecond)},
+		{Foo: "second_foo", Bar: 8.2, Dur: Duration(time.Microsecond)},
+		{Foo: "third_foo", Bar: 7.3, Dur: Duration(time.Millisecond)},
+		{Foo: "fourth_foo", Bar: 6.4, Dur: Duration(time.Second)},
+		{Foo: "fifth_foo", Bar: 5.5, Dur: Duration(time.Minute)},
+		{Foo: "sixth_foo", Bar: 4.6, Dur: Duration(time.Hour)},
+		{Foo: "seventh_foo", Bar: 3.7},
+		{Foo: "eighth_foo", Bar: 2.8},
+		{Foo: "ninth_foo", Bar: 1.9},
+		{Foo: "tenth_foo", Bar: 0.0},
+	}
+	const configFileData = `
+---
+test:
+- foo: first_foo
+  bar: 9.1
+  dur: 1ns
+- foo: second_foo
+  bar: 8.2
+  dur: 1us
+- foo: third_foo
+  bar: 7.3
+  dur: 1ms
+- foo: fourth_foo
+  bar: 6.4
+  dur: 1s
+- foo: fifth_foo
+  bar: 5.5
+  dur: 1m
+- foo: sixth_foo
+  bar: 4.6
+  dur: 1h
+- foo: seventh_foo
+  bar: 3.7
+- foo: eighth_foo
+  bar: 2.8
+- foo: ninth_foo
+  bar: 1.9
+- foo: tenth_foo
+  bar: 0.0
+`
+	const configEnvDataProducts = `
+{"foo": "first_foo",   "bar": 9.1, "dur": "1ns"}
+{"foo": "second_foo",  "bar": 8.2, "dur": "1us"}
+{"foo": "third_foo",   "bar": 7.3, "dur": "1ms"}
+{"foo": "fourth_foo",  "bar": 6.4, "dur": "1s"}
+{"foo": "fifth_foo",   "bar": 5.5, "dur": "1m"}
+{"foo": "sixth_foo",   "bar": 4.6, "dur": "1h"}
+{"foo": "seventh_foo", "bar": 3.7}
+{"foo": "eighth_foo",  "bar": 2.8}
+{"foo": "ninth_foo",   "bar": 1.9}
+{"foo": "tenth_foo",   "bar": 0.0}
+`
+	evalAndCompare := func(cfg *viper.Viper) {
+		var result []testConfig
+		err := UnmarshalSliceSetting(cfg, "test", &result)
+		if err != nil {
+			t.Errorf("error loading offers: %s", err.Error())
+			return
+		}
+		for i, item := range configData {
+			if i >= len(result) {
+				t.Errorf("missing item[%d]: %v", i, item)
+				continue
+			} else if !reflect.DeepEqual(item, result[i]) {
+				t.Errorf("item[%d] does not match expectations: %v (actual) != %v (expected)",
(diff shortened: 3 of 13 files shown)


alfrunes:
Rebasing this PR so I can get the updated useradm migrations.

commitlint.config.js
```diff
@@ -6,6 +6,7 @@ const commonScopes = [
   'deviceconnect',
   'inventory',
   'iot-manager',
+  'pkg',
```
merlin-northern, line 9: let's make it public! I mean: for sure KJ will not notice that, we have to tell him somehow. others as well, me included, I will forget :>

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+	switch cfg := value.(type) {
+	case string:
+		decoder := json.NewDecoder(strings.NewReader(cfg))
+		for {
+			var elem T
+			err = decoder.Decode(&elem)
+			if err != nil {
+				break
+			}
+			*result = append(*result, elem)
```
merlin-northern, line 58: what if result is nil?

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+// UnmarshalSliceSetting will unmarshal an array of objects into the result T
+// using either newline separated json objects for strings (from env) or
+// an array of objects (using json struct tag) for any configuration source
+// that is able to express a slice of objects.
+func UnmarshalSliceSetting[T any](c Reader, path string, result *[]T) error {
+	value := c.Get(path)
+	var err error
+	switch cfg := value.(type) {
+	case string:
+		decoder := json.NewDecoder(strings.NewReader(cfg))
```
merlin-northern, line 51: I am having security doubts here. so without any checks we feed the decoder whatever is in the env. do you have an opinion on that?

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+		for {
+			var elem T
+			err = decoder.Decode(&elem)
+			if err != nil {
+				break
+			}
```
merlin-northern, line 57: are you sure that err === EOF doesnt mean that there was something decoded and then it was end of input? most readers do it.

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		if errors.Is(err, io.EOF) {
+			err = nil
+		}
+	case []any:
+		var decoder *mapstructure.Decoder
+		decoder, err = mapstructure.NewDecoder(
+			&mapstructure.DecoderConfig{
+				TagName: "json",
+				Result:  &result,
```
merlin-northern, line 68: just to double check this will pass **

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/go-viper/mapstructure/v2"
+	"github.com/pkg/errors"
+)
+
+// UnmarshalSliceSetting will unmarshal an array of objects into the result T
+// using either newline separated json objects for strings (from env) or
+// an array of objects (using json struct tag) for any configuration source
+// that is able to express a slice of objects.
+func UnmarshalSliceSetting[T any](c Reader, path string, result *[]T) error {
```
merlin-northern, line 43: this needs thorough units.

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+	value := c.Get(path)
+	var err error
+	switch cfg := value.(type) {
+	case string:
```
merlin-northern, line 50: I find this function confusing. using type of what viper Get returns to determine how to proceed with the decoding, feeding the decoders arbitrary strings from env, it is a nice riddle. doesnt the caller know the type so we can have a simpler decoder(s)?

backend/pkg/config/unmarshal.go
merlin-northern: I just saw the units in the following commits.

backend/pkg/config/unmarshal.go
```diff
@@ -0,0 +1,71 @@
...
+			*result = append(*result, elem)
+		}
+		if errors.Is(err, io.EOF) {
+			err = nil
+		}
+	case []any:
+		var decoder *mapstructure.Decoder
+		decoder, err = mapstructure.NewDecoder(
+			&mapstructure.DecoderConfig{
+				TagName: "json",
```
merlin-northern, line 67: I am highly unfamiliar with the mapstructure decoder, can this config be defined as some variable at least? looks like it is const.

backend/pkg/config/unmarshal_test.go
```diff
@@ -0,0 +1,127 @@
...
+					i, result[i], item)
+			}
+		}
+	}
+
+	t.Run("load from file", func(t *testing.T) {
+		t.Parallel()
+		tmpDir := t.TempDir()
+		configFile := filepath.Join(tmpDir, "config1.yaml")
+		err := os.WriteFile(configFile, []byte(configFileData), 0644)
```
merlin-northern: no need for 644 probably 600 is enough.

backend/pkg/config/unmarshal_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/spf13/viper"
+)
+
+type testConfig struct {
+	Foo string  `json:"foo"`
+	Bar float64 `json:"bar"`
+}
+
+func TestUnmarshalSliceSetting(t *testing.T) {
```
merlin-northern, line 37: this only tests happy paths, what about parse errors, illegal characters, 4Gb of data here and there?

backend/pkg/redis/ratelimit.go
```diff
@@ -15,37 +15,36 @@ func NewFixedWindowRateLimiter(
...
+	_ rate.EventLimiter = &FixedWindowRateLimiter{}
+)
 
-type fixedWindowRatelimiter struct {
-	client     Client
-	paramsFunc RatelimitParamsFunc
-	nowFunc    func() time.Time
-	keyPrefix  string
+// FixedWindowRateLimiter implements a version of the algorithm described
+// at https://redis.io/glossary/rate-limiting/
```
merlin-northern, line 35:
```suggestion
// FixedWindowRateLimiter implements Fixed-window rate limiting as described
// at https://redis.io/glossary/rate-limiting/
```

backend/pkg/redis/ratelimit.go
```diff
@@ -15,37 +15,36 @@ func NewFixedWindowRateLimiter(
...
+	keyPrefix string
 
 	interval time.Duration
+	quota    int64
 }
 
 type simpleReservation struct {
 	ok     bool
-	tokens uint64
+	tokens int64
```
merlin-northern, line 50: so we have both `Token bucket rate limiting` and `Fixed-window rate limiting` (I am referring to https://redis.io/glossary/rate-limiting/)?

backend/pkg/redis/ratelimit.go
merlin-northern, line 1: it is veery hard to review. I have not been in the rate limits at all, and getting this is really challenging to properly review.

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type HTTPLimiter struct {
+	templateData  func(r *http.Request) any
+	template      *template.Template
+	mux           *http.ServeMux
+	DefaultGroup  *limiterGroup
+	limiterGroups map[string]*limiterGroup
+
+	rewriteRequests bool
```
merlin-northern: I know I am usually saying that, and it maybe annoying, but we have to leave some comments. out of those 6 fields at least a half is a mistery to anyone reading from the "outside".

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func defaultTemplateData(r *http.Request) any {
+	id := identity.FromContext(r.Context())
+	ctx := map[string]any{
+		"Identity": id,
+		// "Request":  r, // not needed, but could be handy
```
merlin-northern, line 78: so we need this comment?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/mendersoftware/mender-server/pkg/requestid"
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+var ErrTooManyRequests = errors.New("too many requests")
+
+type limiterGroup struct {
+	limiter       EventLimiter
+	eventTemplate *template.Template
+}
```
merlin-northern: same as below; we need comments here. what is a group? why do we need it? what is the architecture of the limiters? is that documented anywhere?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+	h.limiterGroups[group] = &limiterGroup{
+		limiter:       limiter,
+		eventTemplate: t,
+	}
+	return nil
+}
+
+func (h *HTTPLimiter) MatchHTTPPattern(
+	pattern, groupTemplate string,
+) error {
```
merlin-northern: how often are we going to call this? have we given some thought about the seep of execution of this function?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		if err != nil {
+			return fmt.Errorf("error parsing group_template: %w", err)
+		}
+	}
+	limiterHandle := handle{
+		HTTPLimiter:   h,
+		groupTemplate: t,
+	}
+	h.mux.Handle(pattern, limiterHandle)
```
merlin-northern: this panics if the patterns conflicts; we are aware of that, right?

backend/pkg/rate/limit_http.go
merlin-northern, line 1: there is no changelog in this commit, on purpose?

backend/pkg/redis/ratelimit.go
```diff
@@ -15,37 +15,36 @@ func NewFixedWindowRateLimiter(
...
-		client:     client,
-		paramsFunc: limitFromContext,
-		nowFunc:    time.Now,
-		keyPrefix:  keyPrefix,
-		interval:   interval,
+	quota int64,
+) *FixedWindowRateLimiter {
+	return &FixedWindowRateLimiter{
+		client:    client,
+		nowFunc:   time.Now,
```
merlin-northern, line 22: note that we have to use monotonic clock for the limits.

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+			RequestID: requestid.FromContext(ctx),
+		})
+	}
+}
+
+func (h *HTTPLimiter) MiddlewareGin(c *gin.Context) {
+	r := c.Request
+	if h.rewriteRequests {
+		r = rest.RewriteForwardedRequest(r)
+	}
```
merlin-northern: why does a limiter rewrite requests?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+		})
+		return
+	}
+	if res == nil || res.OK() {
+		return
+	} else {
+		hdr := w.Header()
+		hdr.Set("Content-Type", "application/json")
+		retryAfter := math.Ceil(res.Delay().Abs().Seconds())
+		hdr.Set("Retry-After", fmt.Sprint(retryAfter))
```
merlin-northern: we did use ceil on it, so setting Retry-After to `fmt.Sprintf("%d",int64(retryAfter))` seems more fair.

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+		retryAfter := math.Ceil(res.Delay().Abs().Seconds())
+		c.Header("Retry-After", fmt.Sprint(retryAfter))
+		rest.RenderError(c, http.StatusTooManyRequests, ErrTooManyRequests)
+		c.Abort()
+	}
+}
+
+func (m *HTTPLimiter) Reserve(r *http.Request) (Reservation, error) {
+	var b bytes.Buffer
+	limiter := m.DefaultGroup
```
merlin-northern: three is a small shadowing here with the `type limiter rate.Limiter` could we change the name of the variable?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+				return nil, fmt.Errorf("error executing ratelimit group template: %w", err)
+			}
+			group := m.limiterGroups[b.String()]
+			if group != nil {
+				limiter = group
+			}
+			b.Reset()
+		}
+	}
+	err := limiter.eventTemplate.Execute(&b, templateData)
```
merlin-northern: so we do this template dance only to pass a fancy eventId? why?

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error executing template for event ID: %w", err)
+	}
+	if limiter == nil {
+		return nil, nil
```
merlin-northern: I am not a fan of nil,nil in calls like Reserve are you sure this is what we want? btw we already dereferenced the limiter a couple of lines above.

backend/pkg/rate/limit_http.go
```diff
@@ -0,0 +1,204 @@
...
+func (m *HTTPLimiter) Reserve(r *http.Request) (Reservation, error) {
+	var b bytes.Buffer
+	limiter := m.DefaultGroup
+	templateData := m.templateData(r)
+	ctx := r.Context()
+	h, _ := m.mux.Handler(r)
+	hh, ok := h.(handle)
+	if ok {
+		if hh.groupTemplate != nil {
+			err := hh.groupTemplate.Execute(&b, templateData)
```
merlin-northern, line 238: where can a person who sees this code for the first time, and you are on PTO, read about the whole idea of groups?

backend/pkg/rest.utils/request.go
```diff
@@ -0,0 +1,48 @@
...
+	host := request.Header.Get("X-Forwarded-Host")
+	if host != "" {
+		newRequest.URL.Host = host
+	}
+	method := request.Header.Get("X-Forwarded-Method")
+	if method != "" {
+		newRequest.Method = method
+	}
+	return newRequest
+}
```
merlin-northern: perfect time for the consts for all the X- headers.

backend/pkg/rest.utils/request.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"net/http"
+	"net/url"
+)
+
+// RewriteForwardedRequest makes a shallow clone of request and replaces
+// the URL and Method with X-Forwarded-* headers.
+func RewriteForwardedRequest(request *http.Request) *http.Request {
+	newRequest := new(http.Request)
+	*newRequest = *request
```
merlin-northern, line 35: what if request is nil?

backend/pkg/rate/limit_http_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/mendersoftware/mender-server/pkg/accesslog"
+	"github.com/mendersoftware/mender-server/pkg/identity"
```
merlin-northern:
```suggestion
	"github.com/gin-gonic/gin"
	
	"github.com/mendersoftware/mender-server/pkg/accesslog"
	"github.com/mendersoftware/mender-server/pkg/identity"
```

backend/pkg/config/ratelimits/config.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/mendersoftware/mender-server/pkg/log"
+	"github.com/mendersoftware/mender-server/pkg/rate"
+	"github.com/mendersoftware/mender-server/pkg/redis"
+)
+
+func init() {
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultInterval, "1m")
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultQuota, "120")
+	config.Config.SetDefault(SettingRatelimitsAuthDefaultEventExpression,
+		"{{with .Identity}}{{.Subject}}{{end}}")
```
merlin-northern, line 39: in the defaultTemplate data I saw only the Identity, where is Subject handled?

backend/pkg/config/ratelimits/config.go
```diff
@@ -0,0 +1,118 @@
...
+	SettingRatelimitsAuthDefault                = SettingRatelimitsAuth + ".default"
+	SettingRatelimitsAuthDefaultQuota           = SettingRatelimitsAuthDefault + ".quota"
+	SettingRatelimitsAuthDefaultInterval        = SettingRatelimitsAuthDefault + ".interval"
+	SettingRatelimitsAuthDefaultEventExpression = SettingRatelimitsAuthDefault +
+		".event_expression"
+)
+
+func LoadRatelimits(c config.Reader) (*RatelimitConfig, error) {
+	if !c.GetBool(SettingRatelimitsAuthEnable) {
+		return nil, nil
```
merlin-northern, line 57: just to double check, nil,nil here, ErrNotEnabled would be easier on the caller.

merlin-northern (commented):
github hanged i.e.: gave up when I tried to submit the review. proof of the fact that it is not easy on anyone 😘 I really tried to be constructive, treat as suggestions.

backend/pkg/config/unmarshal.go
alfrunes: Not sure if I understand the concern. This is configured by the operator, the input is not exposed to the public. What do you suggest to do?

backend/pkg/config/unmarshal.go
alfrunes: Good call. I didn't encounter the issue, but it could be the case https://pkg.go.dev/io#Reader

alfrunes: Had to double check. The `json.Decoder` takes care of buffering and handles EOF gracefully: https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/encoding/json/stream.go;l=164-168

backend/pkg/config/unmarshal.go
alfrunes: I don't understand, can you elaborate?

backend/pkg/config/unmarshal.go
alfrunes: The problem I was trying to solve is one that `viper.Unmarshal` partially solves. What I want to achieve is to use a structured config which if configured through environment (the value is a string) uses a series of JSON objects, and if it is configured using a file uses the mapslice from viper and mapstructure to reuse the `json` tags.

backend/pkg/config/unmarshal_test.go
alfrunes: Will add a few unhappy cases. For the big data test I don't understand the motivation. You could put a huge value in any configuration (how about a 10TiB MongoDB connection string?).

backend/pkg/rate/limit_http.go
alfrunes: It was before I split this up into smaller PRs to make it "easier" to review. You can see #772 and #769 for PRs to manually testing the package and some documentation of the rate limiter. I'll add some inline comments as well. :memo:

backend/pkg/rate/limit_http.go
alfrunes: This is called a few times on start-up when configuring the rate limiting router groups.

backend/pkg/rate/limit_http.go
alfrunes: Yes, that would be a configuration error. I don't want to mess with the underlying implementation and instead let it panic as this would result from a configuration error.

backend/pkg/rate/limit_http.go
alfrunes: Yes, it is not a public feature yet, it's only the library part. Not sure what's the policy here though, but I kept it out as I think most consumers are primarily interested in public facing stuff.

backend/pkg/rate/limit_http.go
alfrunes: :memo:

backend/pkg/redis/ratelimit.go
alfrunes: That would be quite a challenge to implement in a distributed system. The implementation uses the Unix timestamps (wall clock) to generate a epoch count for the current fixed window. If Redis provides an API for a monotonic clock, we could utilize that as a clock source but that would come at the cost of performance.

backend/pkg/rate/limit_http.go
alfrunes: It's because it is used in the auth verify APIs (deviceauth/useradm).

backend/pkg/rate/limit_http.go
alfrunes: I can add some inline docs, but should be apparent from the config docs from the parent PR. It is so that you could make the call be grouped by different means. For example the two templates:
* `{{with .Identity}}{{.Subject}}{{end}}`
* `{{with .Identity}}{{.Tenant}}{{end}}`
would make the rate limit count per client (user/device) or per tenant respectively.

backend/pkg/rate/limit_http.go
alfrunes: This was a leftover, I will remove it. I believe this is a constant condition since it should always be set. :memo:

backend/pkg/rate/limit_http.go
alfrunes: :memo:

backend/pkg/rest.utils/request.go
alfrunes: I consciously let it panic :scream_cat:

backend/pkg/config/ratelimits/config.go
alfrunes: The Identity attribute is an identity.Identity struct (please read the commit message: https://github.com/mendersoftware/mender-server/pull/772/commits/56c52a984e67e9bd031ba80868ee8ee8b543375a)

backend/pkg/rate/limit_http.go
alfrunes: Added some comments...

commitlint.config.js
alfrunes: I'm confused, is this related to the commitlint config?

alfrunes:
@merlin-northern I tried my best to address your comment, please see the fixup commits. I will do a `git rebase --autosquash` before merging.
Edit: I'll take care of the static check too.

commitlint.config.js
merlin-northern: no I just meant that there is a new scope -- how to spread the knowledge about it. sorry fot the confusion.

backend/pkg/config/unmarshal.go
merlin-northern: maybe paranoia mode is on, but I would at least check the size, but if you dont share my concerns lets just forget about it.

backend/pkg/config/unmarshal.go
merlin-northern: I meant the type of the Result, as the result is a pointer.

backend/pkg/redis/ratelimit.go
merlin-northern: I understand, but building on non-monotonic clocks has obvious built-in problems. but if you thought about it, it is ok.btw time.Now does track the monotonic clock, and if you calculate the time difference it will use it.

backend/pkg/rate/limit_http.go
```diff
@@ -98,88 +139,92 @@ func (h *HTTPLimiter) MatchHTTPPattern(
...
-	if h.rewriteRequests {
-		r = rest.RewriteForwardedRequest(r)
-	}
-	res, err := h.Reserve(r)
+func handleError(w http.ResponseWriter, err error) {
+	var tooManyRequests *TooManyRequestsError
+	status := http.StatusInternalServerError
+	hdr := w.Header()
+	hdr.Set("Content-Type", "application/json")
+	if errors.As(err, &tooManyRequests) {
```
merlin-northern, line 189: this will work because `As` will look via .Error() call to match the err, right?

backend/pkg/rate/limit_http.go
```diff
@@ -98,88 +139,92 @@ func (h *HTTPLimiter) MatchHTTPPattern(
...
-	r := c.Request
-	if h.rewriteRequests {
-		r = rest.RewriteForwardedRequest(r)
-	}
-	res, err := h.Reserve(r)
+func handleError(w http.ResponseWriter, err error) {
+	var tooManyRequests *TooManyRequestsError
+	status := http.StatusInternalServerError
+	hdr := w.Header()
+	hdr.Set("Content-Type", "application/json")
```
merlin-northern, line 188: I just wondered now, do we return any json here?

backend/pkg/rate/limit_http.go
```diff
@@ -188,17 +233,17 @@ func (m *HTTPLimiter) Reserve(r *http.Request) (Reservation, error) {
...
-	err := limiter.eventTemplate.Execute(&b, templateData)
+	err := eventLimiter.eventTemplate.Execute(&b, templateData)
 	if err != nil {
 		return nil, fmt.Errorf("error executing template for event ID: %w", err)
 	}
-	if limiter == nil {
-		return nil, nil
+	if eventLimiter == nil {
+		return okReservation{}, nil
 	}
```
merlin-northern: why do we first call `err := eventLimiter.eventTemplate.Execute(&b, templateData)` and then check `if eventLimiter == nil`?

backend/pkg/config/ratelimits/config.go
```diff
@@ -24,6 +24,14 @@ import (
 	"github.com/mendersoftware/mender-server/pkg/redis"
 )
 
+type ConfigDisabledError struct {
+	Path string
+}
+
+func (err *ConfigDisabledError) Error() string {
+	return fmt.Sprintf("configuration %q disabled", err.Path)
```
merlin-northern: this is %q and we want it because we need it? if not, the simple string concat would be much faster here.

merlin-northern (dismissed):
left last comments. looks good.

backend/pkg/rate/limit_http.go
alfrunes: It will recursively call unwrap and try to do a type assertion to the same type as `tooManyRequests` (`*TooManyRequestsError`).

backend/pkg/rate/limit_http.go
alfrunes: Ohh man, all this refactoring made me remove the part writing the body :exploding_head:

backend/pkg/rate/limit_http.go
alfrunes: Ouch, another code shuffle mistake :memo:

backend/pkg/rate/limit_http_test.go
```diff
@@ -0,0 +1,181 @@
...
+		if w.Code != http.StatusInternalServerError {
+			t.Errorf("unexpected status code: %d, expected: %d", w.Code, http.StatusInternalServerError)
+			t.Error("expected template execution to fail")
+		}
+		req.URL.Path = "/bad_group/"
+		w = httptest.NewRecorder()
+		handler.ServeHTTP(w, req)
+		if w.Code != http.StatusInternalServerError {
+			t.Errorf("unexpected status code: %d, expected: %d", w.Code, http.StatusInternalServerError)
+			t.Error("expected template execution to fail")
```
merlin-northern, line 178: the execution will never reach this line, Errorf calls Fail() or at least thats what I think.

merlin-northern (approved):
looks good; thanks for reading my comments.

backend/pkg/rate/limit_http_test.go
alfrunes: `Error[f]` prints the message and marks the test as failed and continues execution.
`Fatal[f]` is doing as you describe but calls FailNow and aborts the test.

merlin-northern: ah, yes, I think the fact that Error[f]* calls Fail() vs Fatal calling FailNow helped me here. sorry for the noise.

PR #784, opened 2025-07-02
refactor: Remove github.com/ant0ine/go-json-rest from the backend  

bahaa-ghazal:
This might break the enterprise repo. Best if we merge them together in two separate PRs


backend/pkg/rest.utils/error.go
```diff
@@ -22,3 +27,19 @@ type Error struct {
 func (err Error) Error() string {
 	return err.Err
 }
+
+func IsApiError(e error) bool {
+	_, ok := e.(*Error)
+	return ok
```
alfrunes: `errors.As` might be more stable as it will recursively check wrapped errors if it's the right type.
```suggestion
	var target *Error
	return errors.As(e, &target)
```

backend/pkg/accesslog/middleware_gin.go
```diff
@@ -27,14 +31,71 @@ import (
...
+			)
+		}
+		if more {
+			fmt.Fprintln(&traceback)
+		}
+		n--
+	}
+	return traceback.String()
+}
+
```
alfrunes: Can you please move this back to the original file instead of moving it around?
It makes it harder to review what's actually being removed and it could lead to more difficulties if we were to revert this change.

backend/pkg/accesslog/middleware_gin.go
```diff
@@ -27,14 +31,71 @@ import (
...
+	"github.com/mendersoftware/mender-server/pkg/netutils"
 	"github.com/mendersoftware/mender-server/pkg/rest.utils"
 )
 
+const (
+	StatusClientClosedConnection = 499
+
+	// nolint:lll
+	DefaultLogFormat = "%t %S\033[0m \033[36;1m%Dμs\033[0m \"%r\" \033[1;30m%u \"%{User-Agent}i\"\033[0m"
+	SimpleLogFormat  = "%s %Dμs %r %u %{User-Agent}i"
```
alfrunes: I believe these two consts are no longer in use.

backend/go.mod
```diff
@@ -6,7 +6,6 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.18.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.1
 	github.com/alicebob/miniredis v2.5.0+incompatible
-	github.com/ant0ine/go-json-rest v3.3.2+incompatible
```
alfrunes, line 9: :rocket:

alfrunes (commented):
Overall looks very good.
Could you please fixup the first commit (just move back the parts to the original source files)?

alfrunes (approved):
Amazing, thank you!

PR #785, opened 2025-07-03
fix(inventory): Broken links for legacy endpoints for listing devices

alfrunes:
When using the /api/managment/v1/inventory/devices to list devices, the path prefix is rewritten to /api/0.1.0/devices. This commit fixes the regression.

Ticket: MEN-8517

### DIFF
--- a/backend/services/inventory/api/http/routing.go
+++ b/backend/services/inventory/api/http/routing.go
@@ -39,6 +39,15 @@ func AllowHeaderOptionsGenerator(methods []string) gin.HandlerFunc {
 	}
 }
 
+func rewritePathPrefix(old, new string) gin.HandlerFunc {
+	n := len(old)
+	return func(ctx *gin.Context) {
+		if strings.HasPrefix(ctx.Request.URL.Path, old) {
+			ctx.Request.URL.Path = new + ctx.Request.URL.Path[n:]
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


bahaa-ghazal (dismissed):

backend/services/inventory/api/http/routing.go
```diff
@@ -39,6 +39,15 @@ func AllowHeaderOptionsGenerator(methods []string) gin.HandlerFunc {
 	}
 }
 
+func rewritePathPrefix(old, new string) gin.HandlerFunc {
+	n := len(old)
+	return func(ctx *gin.Context) {
+		if strings.HasPrefix(ctx.Request.URL.Path, old) {
+			ctx.Request.URL.Path = new + ctx.Request.URL.Path[n:]
+		}
```
merlin-northern, line 47: I wonder what would be faster, Replace with limit 1? have you given strings.CutPrefix a look?

merlin-northern (approved):
left some cheeky comment, I cannot believe it how the tests have not caught that?

backend/services/inventory/api/http/routing.go
alfrunes: I was actually looking into the implementation of the strings.Replace, but decided to go with this approach to save scan for counting occurrences and make it position dependent while only using a single allocation.

PR #788, opened 2025-07-07
test(e2e): Fix issue with expect: Property 'then' not found

alfrunes:


frontend/tests/e2e_tests/integration/04-deployments.spec.ts
```diff
@@ -51,7 +51,7 @@ test.describe('Deployments', () => {
     await page.waitForSelector(selectors.releaseSelect, { timeout: timeouts.fiveSeconds });
     const releaseSelect = await page.getByPlaceholder(/select a release/i);
     await releaseSelect.focus();
-    await expect(page.locator(`#deployment-release-selection-listbox li:has-text('mender-demo-artifact')`));
+    await expect(page.locator(`#deployment-release-selection-listbox li:has-text('mender-demo-artifact')`)).toBeTruthy();
```
mineralsfree:
```suggestion
    await expect(page.locator(`#deployment-release-selection-listbox li:has-text('mender-demo-artifact')`)).toBeVisible();
```

mineralsfree (changes requested):
I believe we should use `toBeVisible()` here instead

frontend/tests/e2e_tests/integration/04-deployments.spec.ts
alfrunes: Ahh, ok. I used that initially, but didn't know if it made any sense so I fell back on the more relaxed toBeTruthy :sloth:

mineralsfree (approved):

PR #790, opened 2025-07-10
feat(deployments): Implement new v2 GET `/artifacts` endpoint

bahaa-ghazal:
Ticket: MEN-8181
Changelog: Title

### DIFF
--- a/backend/services/deployments/api/http/api_deployments.go
+++ b/backend/services/deployments/api/http/api_deployments.go
@@ -321,6 +321,56 @@ func getReleaseOrImageFilter(r *http.Request, version listReleasesVersion,
 	return filter
 }
 
+func getImageFilter(c *gin.Context, paginated bool) *model.ImageFilter {
+
+	q := c.Request.URL.Query()
+
+	names := c.QueryArray(ParamName)
+
+	var exactNames []string
+	var namePrefixes []string
+
+	for _, name := range names {
+		if strings.HasSuffix(name, `\*`) {
+			name = strings.TrimSuffix(name, `\*`) + "*"
+			exactNames = append(exactNames, name)
+		} else if strings.HasSuffix(name, "*") {
+			namePrefixes = append(namePrefixes, strings.TrimSuffix(name, `*`))
+		} else {
+			exactNames = append(exactNames, name)
+		}
+	}
+
+	filter := &model.ImageFilter{
+		ExactNames:   exactNames,
+		NamePrefixes: namePrefixes,
+		Description:  q.Get(ParamDescription),
+		DeviceType:   q.Get(ParamDeviceType),
+	}
+
+	if paginated {
+		// filter.Sort = q.Get(ParamSort)
+		if page := q.Get(ParamPage); page != "" {
+			if i, err := strconv.Atoi(page); err == nil {
+				filter.Page = i
+			}
+		}
+		if perPage := q.Get(ParamPerPage); perPage != "" {
+			if i, err := strconv.Atoi(perPage); err == nil {
+				filter.PerPage = i
+			}
+		}
+		if filter.Page <= 0 {
+			filter.Page = 1
+		}
+		if filter.PerPage <= 0 || filter.PerPage > MaximumPerPage {
+			filter.PerPage = DefaultPerPage
+		}
+	}
+
+	return filter
+}
+
 type limitResponse struct {
 	Limit uint64 `json:"limit"`
 	Usage uint64 `json:"usage"`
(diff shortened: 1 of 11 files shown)


backend/services/deployments/api/http/api_deployments.go
```diff
@@ -310,6 +310,53 @@ func getReleaseOrImageFilter(r *http.Request, version listReleasesVersion,
...
+
+	q := c.Request.URL.Query()
+
+	names := c.QueryArray(ParamName)
+
+	var exactNames []string
+	var nameprefixes []string
+
+	for _, name := range names {
+		if strings.HasSuffix(name, "*") {
```
alfrunes: Maybe add the possibility to escape the last character?
```suggestion
		if strings.HasSuffix(name, `\*`) {
			name = strings.TrimSuffix(name, `\*`) + "*"
			exactNames = append(exactNames, name)
		} else if strings.HasSuffix(name, "*") {
```

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -310,6 +310,53 @@ func getReleaseOrImageFilter(r *http.Request, version listReleasesVersion,
...
 }
 
+func getImageFilter(c *gin.Context, paginated bool) *model.ImageFilter {
+
+	q := c.Request.URL.Query()
+
+	names := c.QueryArray(ParamName)
+
+	var exactNames []string
+	var nameprefixes []string
```
alfrunes: I think some limits are in order here. How about applying the following limitations to start with:
* `len(exactNames) <= 100`
* `len(namePrefixes) <= 1`

Should we also make the two options mutually exclusive? What do you think @kjaskiewiczz?

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -1284,6 +1284,138 @@ func (db *DataStoreMongo) ListImages(
...
+			return nil, 0, nil
+		}
+		return nil, 0, err
+	}
+
+	count, err := collImg.CountDocuments(ctx, filters)
+	if err != nil {
+		return nil, -1, ErrDevicesCountFailed
+	}
+
```
alfrunes: My understanding is that we won't do that for the new endpoint.

kjaskiewiczz: that's correct; please remove it (it's not in the spec)

backend/services/deployments/api/http/api_deployments.go
kjaskiewiczz: yes, I think it make sense to make exact names and name prefix mutually exclusive and allows only single name prefix in single query;
and 100 feels also reasonable as limit for number of exact names in single query;
@alfrunes thanks for your comments

backend/services/deployments/model/image.go
```diff
@@ -315,3 +316,34 @@ func (p *ProvidesIdx) UnmarshalBSONValue(t bsontype.Type, b []byte) error {
...
+
+func (filter ImageFilter) Validate() error {
+	return validation.ValidateStruct(&filter,
+		validation.Field(&filter.ExactNames,
+			validation.Length(0, 100),
+			validation.Each(lengthLessThan4096)),
+
+		validation.Field(&filter.NamePrefixes,
+			validation.Length(0, 1),
+			validation.Each(lengthLessThan4096)),
```
alfrunes, line 342: Can you also make the two parameters mutually exclusive?

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -410,6 +460,50 @@ func (d *DeploymentsApiHandlers) ListImages(c *gin.Context) {
...
+	// add one more item to the limit to check if there is a next page.
+	filter.Limit = filter.PerPage + 1
+
+	list, err := d.app.ListImagesV2(c.Request.Context(), filter)
+	if err != nil {
+		d.view.RenderInternalError(c, err)
+		return
+	}
+
+	len := len(list)
```
alfrunes: Overloads built-in `len` with a variable.

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -1284,6 +1284,137 @@ func (db *DataStoreMongo) ListImages(
...
+					"$regex": primitive.Regex{
+						Pattern: "^" + regexp.QuoteMeta(name) + ".*",
+						Options: "i",
+					},
+				},
+			})
+		}
+		if len(nameFilter) > 0 {
+			filters["$or"] = nameFilter
+		}
```
alfrunes: (suggestion) Since the options are mutually exclusive this could be simplified.
```suggestion
	if len(filt.ExactNames) > 0 {
		filters[StorageKeyName] = bson.M{"$in": filt.ExactNames}
	} else if filt.NamePrefix != "" {
		filters[StorageKeyName] = bson.M{"$regex": "^" + regexp.QuoteMeta(filt.NamePrefix)}	
	}
```

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -1284,6 +1284,137 @@ func (db *DataStoreMongo) ListImages(
...
+
+		if filt.Description != "" {
+			if strings.HasSuffix(filt.Description, "*") {
+				description := strings.TrimSuffix(filt.Description, "*")
```
alfrunes: Missing check for escape sequence in the end `\*`

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -1284,6 +1284,137 @@ func (db *DataStoreMongo) ListImages(
...
+						Options: "i",
+					},
+				}
+			} else {
+				filters[StorageKeyImageDescription] = filt.Description
+			}
+		}
+
+		if filt.DeviceType != "" {
+			if strings.HasSuffix(filt.DeviceType, "*") {
```
alfrunes: Also here.

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -1284,6 +1284,137 @@ func (db *DataStoreMongo) ListImages(
...
+		}
+		if sortOrder == 0 {
+			sortOrder = 1
+		}
+	*/
+	sortField := StorageKeyImageName
+	sortOrder := 1
+	findOptions.SetSort(bson.D{
+		{Key: sortField, Value: sortOrder},
+		{Key: "_id", Value: sortOrder},
```
alfrunes: The ID secondary sort prevents use of the existing index on the name field.
```suggestion
```

alfrunes (commented):
Sorry, just a couple more minor I missed in my previous review.

alfrunes (dismissed):

alfrunes (approved):

PR #792, opened 2025-07-10
test: backend unit tests with parallel 1 for stability of pipelines

merlin-northern:
Changelog: None
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -266,7 +266,7 @@ test:backend:unit:
   script:
     - |
       make -C backend test-unit \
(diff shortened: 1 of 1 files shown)


oldgiova (approved):

alfrunes (approved):

alfrunes (changes requested):
Sorry I was a bit fast clicking the green button yesterday. It just hit me that I was confusing the `test` flag `-parallel` with the `build` flag `-p`. It's the latter that is building and running tests from different packages in parallel which is causing issues when the unit test depends on a shared resource (like mongodb).

mzedel:
@merlin-northern, @alfrunes is this still relevant?

alfrunes:
No, I think this one is outdated.

PR #793, opened 2025-07-13
chore(gui): describe the offline check better.

merlin-northern:
Changelog: None
Ticket: ME-547


merlin-northern:
there has been a change of plans.

PR #794, opened 2025-07-14
MEN-8523: Make all routes response with JSON when route or method missing

bahaa-ghazal:


backend/pkg/routing/routing.go
alfrunes, line 1: Finally! :smile_cat:

backend/pkg/routing/routing.go
```diff
@@ -64,3 +67,37 @@ func AutogenOptionsRoutes(router *gin.Engine, gen HttpOptionsGenerator) {
...
+
+func handleNoMethod(c *gin.Context) {
+	c.JSON(http.StatusMethodNotAllowed, rest.Error{
+		Err:       "method not allowed",
+		RequestID: requestid.FromContext(c.Request.Context()),
+	})
+}
+
+func SwitchToReleaseMode() {
+	gin.SetMode(gin.ReleaseMode)
```
alfrunes: It could still be handy for debugging locally to be able to override the gin mode using the environment.
```suggestion
	if mode := os.Getenv(gin.EnvGinMode); mode != "" {
		gin.SetMode(mode)
	} else {
		gin.SetMode(gin.ReleaseMode)
	}
	gin.DisableConsoleColor()
```

alfrunes (dismissed):
Great work. It would be nice to keep the `GIN_MODE` override, but this is a good step towards consistency.

alfrunes (approved):

PR #796, opened 2025-07-15
chore: extract dir paths to common vars

dhaustein:
- centralized directory variables (FRONTEND_DIR, LOGS_DIR, COVERAGE_DIR, SCREENSHOTS_DIR, JUNIT_DIR) at pipeline level
  - replaced hardcoded "frontend" dir paths with `${FRONTEND_DIR}` variable throughout
  - changed the npm commands to use `--prefix` instead of `cd` commands
- updated the Deno image from `debian-1.46.3` to `debian-2.0.2` for the licenseCheck job
- update the `licenseCheck` script to not be hardcoded to use only current directory
- added `license-checker-rseidelsohn` to `deno.json` (not sure if strictly needed but I saw some import errors without it)
- removed archiving frontend/logs from the `test:frontend:unit` job

### DIFF
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -81,13 +81,10 @@ test:frontend:unit:
     - cd frontend
     - npm ci
     - npm run test -- --reporter=default --reporter=junit --outputFile.junit=./junit.xml --coverage
-  after_script:
-    - mkdir logs && cp frontend/logs/* logs/
   artifacts:
     expire_in: 2w
     paths:
       - frontend/coverage
-      - logs
     reports:
       junit: frontend/junit.xml
     when: always
@@ -204,7 +201,6 @@ build:frontend:docker:
     paths:
       - frontend/coverage
       - frontend/screenshots
-      - frontend/logs
       - frontend/junit
     reports:
       junit:
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -43,7 +43,7 @@ test:frontend:license-headers:
     - git diff --exit-code frontend/tests
 
 test:frontend:licenses:
-  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-1.46.3
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-2.0.2
   stage: test
   rules:
     - if: $CI_COMMIT_REF_NAME =~ /^(staging|saas-[a-zA-Z0-9.]+)$/
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -6,6 +6,7 @@ variables:
   DOCS_VERSION: development
   MENDER_IMAGE_GUI: ${MENDER_IMAGE_REGISTRY}/${MENDER_IMAGE_REPOSITORY}/gui:${MENDER_IMAGE_TAG}
   FF_TIMESTAMPS: true
+  FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
 
 test:frontend:lint:
   stage: test
@@ -18,10 +19,9 @@ test:frontend:lint:
         compare_to: '${RULES_CHANGES_COMPARE_TO_REF}'
   needs: []
   script:
-    - cd frontend
-    - npm ci --cache .npm --prefer-offline
-    - cd tests/e2e_tests && npm ci && cd ../..
-    - npm run lint
+    - npm ci --prefix ${FRONTEND_DIR} --cache ${FRONTEND_DIR}/.npm --prefer-offline
+    - npm ci --prefix ${FRONTEND_DIR}/tests/e2e_tests
+    - npm run lint --prefix ${FRONTEND_DIR}
   tags:
     - k8s
 
@@ -38,9 +38,9 @@ test:frontend:license-headers:
   before_script:
     - apt update && apt install git -yq
   script:
-    - deno task --cwd frontend/scripts licenseCheck --rootDir $(pwd)
-    - git diff --exit-code frontend/src
-    - git diff --exit-code frontend/tests
+    - deno task --cwd ${FRONTEND_DIR}/scripts licenseCheck --rootDir ${CI_PROJECT_DIR}
+    - git diff --exit-code ${FRONTEND_DIR}/src
+    - git diff --exit-code ${FRONTEND_DIR}/tests
 
 test:frontend:licenses:
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-2.0.2
@@ -55,13 +55,12 @@ test:frontend:licenses:
     - job: build:frontend:docker
       artifacts: true
   before_script:
-    - cd frontend
     - apt-get update && apt-get install -yq curl
     - curl -fsSL https://deb.nodesource.com/setup_23.x | bash
     - apt-get install -yq nodejs
-    - npm ci
+    - npm ci --prefix ${FRONTEND_DIR}
   script:
-    - deno run --allow-env --allow-read --allow-sys tests/licenses/licenseCheck.ts --rootDir $(pwd)
+    - deno run --allow-env --allow-read --allow-sys --config ${FRONTEND_DIR}/scripts/deno.json ${FRONTEND_DIR}/tests/licenses/licenseCheck.ts --rootDir ${FRONTEND_DIR}
   tags:
     - k8s
 
@@ -78,15 +77,14 @@ test:frontend:unit:
   needs: []
   timeout: 10 minutes
   script:
-    - cd frontend
-    - npm ci
-    - npm run test -- --reporter=default --reporter=junit --outputFile.junit=./junit.xml --coverage
+    - npm ci --prefix ${FRONTEND_DIR}
+    - npm run test --prefix ${FRONTEND_DIR} -- --reporter=default --reporter=junit --outputFile.junit=${FRONTEND_DIR}/junit/junit.xml --coverage
   artifacts:
     expire_in: 2w
     paths:
-      - frontend/coverage
+      - ${FRONTEND_DIR}/coverage
     reports:
-      junit: frontend/junit.xml
+      junit: ${FRONTEND_DIR}/junit/junit.xml
     when: always
   tags:
     - k8s
@@ -103,9 +101,8 @@ test:frontend:docs-links:
   needs: []
   before_script:
     - apk add --no-cache curl
-    - cd frontend
   script:
-    - links=$(grep -r docs.mender.io src/ | grep -v snapshots | sed -e 's,\${docsVersion},'${DOCS_VERSION}'/,g' | sed -e 's,\${path},''/,g' | egrep -o 'https?://[^ `"]+' | sort | uniq)
+    - links=$(grep -r docs.mender.io ${FRONTEND_DIR}/src/ | grep -v snapshots | sed -e 's,\${docsVersion},'${DOCS_VERSION}'/,g' | sed -e 's,\${path},''/,g' | egrep -o 'https?://[^ `"]+' | sort | uniq)
     - error=0
     - |
       for url in $links; do
@@ -139,11 +136,13 @@ build:frontend:docker:
     - if: '$CI_COMMIT_REF_PROTECTED == "true"'
       when: always
   script:
-    - cd frontend
-    - if test -z "${DOCKER_PLATFORM}"; then
-    - docker context create ci;
-    - fi
-    - docker buildx create ${DOCKER_BUILDKITARGS} --name gui-builder --driver=docker-container ci
+    - |
+      if test -z "${DOCKER_PLATFORM}"; then
+        docker context create ci;
+      fi
+    - docker buildx create ${DOCKER_BUILDKITARGS}
+      --name gui-builder
+      --driver=docker-container ci
     # build production target
     - docker build
       --tag ${MENDER_IMAGE_GUI}
@@ -156,13 +155,15 @@ build:frontend:docker:
       --platform ${DOCKER_PLATFORM:-linux/amd64}
       --provenance false
       --push
-      .
+      ${FRONTEND_DIR}
     - docker context use ci
-    - docker run --rm --entrypoint "/bin/sh" -v $(pwd):/extract ${MENDER_IMAGE_GUI} -c "cp licenses.json /extract/"
+    - docker run --rm --entrypoint "/bin/sh"
+      -v ${FRONTEND_DIR}:/extract ${MENDER_IMAGE_GUI}
+      -c "cp licenses.json /extract/"
   artifacts:
     expire_in: 1w
     paths:
-      - frontend/licenses.json
+      - ${FRONTEND_DIR}/licenses.json
 
 .template:test:frontend:acceptance:
   stage: test
@@ -199,12 +200,12 @@ build:frontend:docker:
   artifacts:
     expire_in: 2w
     paths:
-      - frontend/coverage
-      - frontend/screenshots
-      - frontend/junit
+      - ${FRONTEND_DIR}/coverage
+      - ${FRONTEND_DIR}/screenshots
+      - ${FRONTEND_DIR}/junit
     reports:
       junit:
-        - frontend/junit/results.xml
+        - ${FRONTEND_DIR}/junit/results.xml
     when: always
   tags:
     - hetzner-amd-ax42
@@ -212,7 +213,7 @@ build:frontend:docker:
 test:frontend:acceptance:
   extends: .template:test:frontend:acceptance
   script:
-    - ./frontend/tests/e2e_tests/run
+    - ${FRONTEND_DIR}/tests/e2e_tests/run
 
 test:frontend:acceptance:enterprise:
   extends: .template:test:frontend:acceptance
@@ -225,7 +226,7 @@ test:frontend:acceptance:enterprise:
     - unset MENDER_IMAGE_REGISTRY MENDER_IMAGE_REPOSITORY
     - export MENDER_IMAGE_TAG=main
     - docker login -u $REGISTRY_MENDER_IO_USERNAME -p $REGISTRY_MENDER_IO_PASSWORD registry.mender.io
-    - ./frontend/tests/e2e_tests/run --enterprise
+    - ${FRONTEND_DIR}/tests/e2e_tests/run --enterprise
 
 .template:publish:frontend:tests:
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/${NODE_IMAGE}
@@ -256,7 +257,7 @@ publish:frontend:tests:
   script:
     - export COVERALLS_SERVICE_JOB_NUMBER=frontend-unit
     - export COVERALLS_FLAG_NAME=frontend-unit
-    - coveralls < frontend/coverage/lcov.info
+    - coveralls < ${FRONTEND_DIR}/coverage/lcov.info
 
 publish:frontend:e2e-tests:
   extends: .template:publish:frontend:tests
@@ -275,8 +276,8 @@ publish:frontend:e2e-tests:
     COVERALLS_SERVICE_JOB_NUMBER: frontend-e2e
     COVERALLS_FLAG_NAME: frontend-e2e
   script:
-    - sed -i -re 's/(^[SF:]+[../]+)(.*)$/SF:\2/' frontend/coverage/lcov.info
-    - coveralls < frontend/coverage/lcov.info
+    - sed -i -re 's/(^[SF:]+[../]+)(.*)$/SF:\2/' ${FRONTEND_DIR}/coverage/lcov.info
+    - coveralls < ${FRONTEND_DIR}/coverage/lcov.info
 
 publish:frontend:e2e-tests:enterprise:
   extends: publish:frontend:e2e-tests
@@ -346,12 +347,12 @@ publish:frontend:licenses:
     - job: build:frontend:docker
       artifacts: true
   script:
-    - deno task --cwd frontend/scripts licenseFormatting --rootDir $(pwd)
+    - deno task --cwd ${FRONTEND_DIR}/scripts licenseFormatting --rootDir ${CI_PROJECT_DIR}
   artifacts:
     when: on_success
     expire_in: 1w
     paths:
-      - frontend/licenses.md
+      - ${FRONTEND_DIR}/licenses.md
 
 publish:frontend:sentry:finalize:
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/getsentry/sentry-cli
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -212,7 +212,7 @@ build:frontend:docker:
 test:frontend:acceptance:
   extends: .template:test:frontend:acceptance
   script:
-    - GUI_REPOSITORY=$(pwd)/frontend SERVER_ROOT=$(pwd) ./frontend/tests/e2e_tests/run
+    - ./frontend/tests/e2e_tests/run
 
 test:frontend:acceptance:enterprise:
   extends: .template:test:frontend:acceptance
@@ -225,7 +225,7 @@ test:frontend:acceptance:enterprise:
     - unset MENDER_IMAGE_REGISTRY MENDER_IMAGE_REPOSITORY
     - export MENDER_IMAGE_TAG=main
     - docker login -u $REGISTRY_MENDER_IO_USERNAME -p $REGISTRY_MENDER_IO_PASSWORD registry.mender.io
-    - GUI_REPOSITORY=$(pwd)/frontend SERVER_ROOT=$(pwd) ./frontend/tests/e2e_tests/run --enterprise
+    - ./frontend/tests/e2e_tests/run --enterprise
 
 .template:publish:frontend:tests:
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/${NODE_IMAGE}
--- a/frontend/scripts/deno.json
+++ b/frontend/scripts/deno.json
@@ -1,5 +1,8 @@
 {
   "nodeModulesDir": "auto",
+  "imports": {
+    "license-checker-rseidelsohn": "npm:license-checker-rseidelsohn@4.2.10"
+  },
   "tasks": {
     "licenseCheck": "deno run --allow-read --allow-write --allow-run addLicense.js",
     "licenseFormatting": "deno run --allow-read --allow-write --allow-net formatLicenses.ts",
--- a/frontend/tests/licenses/licenseCheck.ts
+++ b/frontend/tests/licenses/licenseCheck.ts
@@ -15,7 +15,7 @@ import { DiffTerm, diff } from 'https://deno.land/x/diff_kit@v2.0.4/mod.ts';
 import { parseArgs } from 'jsr:@std/cli/parse-args';
 import { dirname, fromFileUrl, join, resolve as resolvePath } from 'jsr:@std/path';
 // import { dirname, fromFileUrl, resolve as resolvePath } from 'jsr:@std/path/mod.ts';
-import { asCSV, asSummary, init } from 'npm:license-checker-rseidelsohn@4.2.10';
+import { asCSV, asSummary, init } from 'license-checker-rseidelsohn';
 
 const licenseFile = 'directDependencies.csv';
 const licenseFileLocation = resolvePath(dirname(fromFileUrl(Deno.mainModule)), licenseFile);
--- a/frontend/tests/licenses/licenseCheck.ts
+++ b/frontend/tests/licenses/licenseCheck.ts
@@ -31,7 +31,7 @@ const usageMessage = [
   ''
 ].join('\n');
 
-const collectUsedLicenses = () => new Promise((resolve, reject) => init({ direct: 0, start: '.' }, (err, packages) => (err ? reject(err) : resolve(packages))));
+const collectUsedLicenses = (root: string) => new Promise((resolve, reject) => init({ direct: 0, start: root }, (err, packages) => (err ? reject(err) : resolve(packages))));
 
 const createPackageData = packageData => ({ licenses: packageData.licenses ?? 'unknown', repository: packageData.repository || packageData.url || '' });
 
@@ -49,8 +49,8 @@ const simplifyPackageData = packages =>
     return accu;
   }, {});
 
-const checkLicenses = () =>
-  collectUsedLicenses()
+const checkLicenses = (root: string = '.') =>
+  collectUsedLicenses(root)
     .then(simplifyPackageData)
     .then(async packages => {
       const existing = await Deno.readTextFile(licenseFileLocation);
@@ -59,8 +59,8 @@ const checkLicenses = () =>
       Deno.exit(existing === current ? 0 : 1);
     });
 
-const writeLicenseFile = () =>
-  collectUsedLicenses()
+const writeLicenseFile = (root: string = '.') =>
+  collectUsedLicenses(root)
     .then(simplifyPackageData)
     .then(async packages => {
       const newPackageData = `${asCSV(packages)}\n`;
@@ -82,7 +82,7 @@ await checkGeneratedLicenses(passedRoot);
 
 const hasUnknownFlag = Object.keys(flags).some(option => !knownOptions.find(({ key }) => option === key));
 if (flags.update) {
-  await writeLicenseFile();
+  await writeLicenseFile(passedRoot);
   Deno.exit(0);
 }
 if (flags.help || hasUnknownFlag) {
@@ -90,8 +90,8 @@ if (flags.help || hasUnknownFlag) {
   Deno.exit(0);
 }
 if (flags.summary) {
-  const result = await collectUsedLicenses();
+  const result = await collectUsedLicenses(passedRoot);
   console.log(asSummary(result));
   Deno.exit(0);
 }
-checkLicenses();
+checkLicenses(passedRoot);
--- a/frontend/tests/e2e_tests/run
+++ b/frontend/tests/e2e_tests/run
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-export SERVER_ROOT=${SERVIER_ROOT:-$(git rev-parse --show-toplevel)}
+export SERVER_ROOT=${SERVER_ROOT:-$(git rev-parse --show-toplevel)}
 export GUI_REPOSITORY=${GUI_REPOSITORY:-${SERVER_ROOT}/frontend}
 export COMPOSE_FILE="${SERVER_ROOT}/docker-compose.yml:${GUI_REPOSITORY}/tests/e2e_tests/docker-compose.e2e-tests.yml"
 LOCAL=""


frontend/pipeline.yml
```diff
@@ -38,12 +42,12 @@ test:frontend:license-headers:
...
-    - deno task --cwd frontend/scripts licenseCheck --rootDir $(pwd)
-    - git diff --exit-code frontend/src
-    - git diff --exit-code frontend/tests
+    - deno task --cwd ${FRONTEND_DIR}/scripts licenseCheck --rootDir ${CI_PROJECT_DIR}
+    - git diff --exit-code ${FRONTEND_DIR}/src
+    - git diff --exit-code ${FRONTEND_DIR}/tests
 
 test:frontend:licenses:
-  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-1.46.3
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/denoland/deno:debian-2.0.2
```
dhaustein, line 46: Hopefully this version works with the `licenseCheck` script, the other jobs here already use deno 2.0.2

frontend/pipeline.yml
```diff
@@ -77,19 +82,27 @@ test:frontend:unit:
...
   needs: []
   timeout: 10 minutes
+  variables:
+    IS_REACT_ACT_ENVIRONMENT: "true"
   script:
-    - cd frontend
-    - npm ci
-    - npm run test -- --reporter=default --reporter=junit --outputFile.junit=./junit.xml --coverage
+    - npm ci --prefix ${FRONTEND_DIR}
+    - npm run test --prefix ${FRONTEND_DIR} -- --reporter=default --reporter=junit --outputFile.junit=${JUNIT_DIR}/junit.xml --coverage --watchAll=false
```
dhaustein: throwing `--watchAll=false` in here, but I am not sure this is going to have any effect at all, probably not going to break anything though

frontend/pipeline.yml
```diff
@@ -142,35 +154,37 @@ build:frontend:docker:
...
     paths:
-      - frontend/licenses.json
+      - ${FRONTEND_DIR}/licenses.json
 
 .template:test:frontend:acceptance:
   stage: test
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-cli
   extends: .build:base
+  variables:
+    IS_REACT_ACT_ENVIRONMENT: "true"
```
dhaustein: an attempt at suppressing the mountain of warnings
```
Warning: The current testing environment is not configured to support act(...)
```

dhaustein:
@mender-test-bot sync

dhaustein:
trying to re-open the PR

dhaustein:
@mender-test-bot sync

frontend/pipeline.yml
dhaustein: judging by the pipeline it does!

frontend/pipeline.yml
dhaustein: I removed this, it didn't do anything, actually it was not recognized and the job failed on it

frontend/pipeline.yml
dhaustein: hm, didn't do anything for the logs, maybe there is something else I can do

frontend/scripts/deno.json
```diff
@@ -1,5 +1,8 @@
 {
   "nodeModulesDir": "auto",
+  "imports": {
+    "license-checker-rseidelsohn": "npm:license-checker-rseidelsohn@4.2.10"
+  },
```
dhaustein, line 5: Note: I wonder if this was actually needed since deno scans for node packages too 🤔

frontend/pipeline.yml
dhaustein: removing this, to truly get rid of the warnings we would need to address the tests, will create a ticket later

frontend/pipeline.yml
```diff
@@ -142,30 +150,30 @@ build:frontend:docker:
...
+        --build-arg GIT_COMMIT_SHA="${CI_COMMIT_SHA}" \
+        --build-arg SENTRY_ORG="${SENTRY_ORG}" \
+        --build-arg SENTRY_URL="${SENTRY_URL}" \
+        --secret id=sentryAuthToken,env=SENTRY_AUTH_TOKEN \
+        --platform ${DOCKER_PLATFORM:-linux/amd64} \
+        --provenance false \
+        --push \
+        ${FRONTEND_DIR}
+      docker context use ci
+      docker run --rm --entrypoint "/bin/sh" -v ${FRONTEND_DIR}:/extract ${MENDER_IMAGE_GUI} -c "cp licenses.json /extract/"
```
alfrunes: FYI; this will collapse the whole string in the execution log making it difficult to debug if this fails.

frontend/tests/licenses/licenseCheck.ts
```diff
@@ -82,16 +82,16 @@ await checkGeneratedLicenses(passedRoot);
...
   Deno.exit(0);
 }
 if (flags.summary) {
-  const result = await collectUsedLicenses();
+  const result = await collectUsedLicenses(passedRoot);
   console.log(asSummary(result));
   Deno.exit(0);
 }
-checkLicenses();
+checkLicenses(passedRoot);
```
alfrunes: Missing newline at the end of the file :monocle_face:

alfrunes (commented):
CI change looks reasonable to me. However, I'm not too familiar with the npm setup here and would like to defer the approval to @mendersoftware/frontend-dependabot-reviewers

frontend/tests/licenses/licenseCheck.ts
dhaustein: will add

frontend/pipeline.yml
dhaustein: yeah I was debating with myself if I should go back to the line by line version like the rest, I'll change it

frontend/tests/licenses/licenseCheck.ts
dhaustein: done

frontend/pipeline.yml
dhaustein: done

frontend/pipeline.yml
```diff
@@ -6,6 +6,11 @@ variables:
   DOCS_VERSION: development
   MENDER_IMAGE_GUI: ${MENDER_IMAGE_REGISTRY}/${MENDER_IMAGE_REPOSITORY}/gui:${MENDER_IMAGE_TAG}
   FF_TIMESTAMPS: true
+  FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
```
mzedel, line 9: I guess I can see some intention of doing this, but the hardcoded nature of the location is dictated by the repository structure, so making it configurable is IMO an invitation to not set this locally/ in new jobs. Based on experience I know that other people use the pipeline definition as a reference for local execution - as such I think this is an extra hurdle when seen in combination with the fixed repo structure.

frontend/pipeline.yml
```diff
@@ -78,18 +82,23 @@ test:frontend:unit:
...
+    - npm run test --prefix ${FRONTEND_DIR} -- --reporter=default --reporter=junit --outputFile.junit=${JUNIT_DIR}/junit.xml --coverage
   after_script:
-    - mkdir logs && cp frontend/logs/* logs/
+    - mkdir -p ${LOGS_DIR}
+    - |
+      if [ -d "${FRONTEND_DIR}/logs" ] && [ "$(ls -A ${FRONTEND_DIR}/logs 2>/dev/null)" ]; then
+        cp ${FRONTEND_DIR}/logs/* ${LOGS_DIR}/
+      else
+        echo "No ${FRONTEND_DIR}/logs directory found or it's empty"
+      fi
```
mzedel: This might well be broken since the switch to vitest several months ago or even earlier.
Instead of this it might be better to either drop the log storage as it's not being consumed as an artifact (+ unlike artifacts the availability of the entire job log from gitlab is much higher)

frontend/pipeline.yml
```diff
@@ -54,14 +58,14 @@ test:frontend:licenses:
...
   before_script:
-    - cd frontend
     - apt-get update && apt-get install -yq curl
     - curl -fsSL https://deb.nodesource.com/setup_23.x | bash
     - apt-get install -yq nodejs
-    - npm ci
+    - npm ci --prefix ${FRONTEND_DIR}
   script:
-    - deno run --allow-env --allow-read --allow-sys tests/licenses/licenseCheck.ts --rootDir $(pwd)
+    - deno run --allow-env --allow-read --allow-sys --config ${FRONTEND_DIR}/scripts/deno.json ${FRONTEND_DIR}/tests/licenses/licenseCheck.ts --rootDir ${FRONTEND_DIR}
```
mzedel, line 63: Instead of using the independent deno config like this (outside of the licenseCheck script hierarchy), you could move the `licenseCheck` to the scripts, define the execution as a new task in there and make sure the comparison in the file points to the license csv file. The change in versioning requires the config file, as otherwise the missing `"nodeModulesDir": "auto",` would prevent retrieving the package in CI - this also causes the import map entry in there that you noted about 👌.

This wasn't intended to be in there, but would make a nice consolidation 👍.

frontend/pipeline.yml
```diff
@@ -202,21 +214,21 @@ build:frontend:docker:
...
-      - frontend/junit
+      - ${COVERAGE_DIR}
+      - ${SCREENSHOTS_DIR}
+      - ${LOGS_DIR}
+      - ${JUNIT_DIR}
     reports:
       junit:
-        - frontend/junit/results.xml
+        - ${JUNIT_DIR}/results.xml
     when: always
```
mzedel: This is rather risky, as this disconnects the job from the location definition, but unlike the unit tests the location here is also set in the e2e compose files - which OTH won't have the env vars set, as they rely on the repo structure... e.g. changing the vars here would then require additional adjustments in the compose files.

frontend/pipeline.yml
mzedel: We don't use jest + even in jest watching is disabled once the `CI` env var is truthy.

frontend/pipeline.yml
mzedel: We don't use jest, thus the env var doesn't apply + quite a few of the warnings were caused by the msw/ rtl setup iirc (@mineralsfree might know best here as he spent time on this more recently)

mzedel (changes requested):
Quite a few changes in a single commit that don't align with the commit message...

The change in the commit title is IMO a personal preference over the repo structure, which is indeed hardcoded. Can you give an explanation as to the improvements this would bring over the explicit reliance on the repo structure - also in light of the added indirection?

More general:
Pure formatting changes are IMO also mostly a personal preference and should go at least into a separate commit if they are included at all as they will usually just cause noise for the people editing the files frequently.

frontend/pipeline.yml
dhaustein: It was never about making it configurable. Unfortunately the way GL does variables is that these will show up when you want to run the pipeline manually, but that's something I cannot change.

This is about addressing two things:
- mitigating bad practice of hardcoding the same values all over and extracting them into one single source of truth, this is programming 101
- brittleness and indirection caused by cd-ing around, I have seen this to be a huge pain in the ass in my years of building CI/CDs, granted it was mostly Jenkins, but the technology does not matter in this case

frontend/pipeline.yml
dhaustein: I don't follow, the logs are being stored as an artifact. Do you want me to get rid of this part then?

frontend/pipeline.yml
dhaustein: Don't the e2e compose files use `GUI_REPOSITORY`? That is being constructed here
```
frontend/pipeline.yml
231:    - GUI_REPOSITORY=${FRONTEND_DIR} SERVER_ROOT=${CI_PROJECT_DIR} ${FRONTEND_DIR}/tests/e2e_tests/run
244:    - GUI_REPOSITORY=${FRONTEND_DIR} SERVER_ROOT=${CI_PROJECT_DIR} ${FRONTEND_DIR}/tests/e2e_tests/run --enterprise

frontend/tests/e2e_tests/run
5:export GUI_REPOSITORY=${GUI_REPOSITORY:-${SERVER_ROOT}/frontend}
```
Then the usage of it in `frontend/tests/e2e_tests/docker-compose.e2e-tests.yml` is all `GUI_REPOSITORY`
I don't see why this shouldn't work. Or am I missing something?

frontend/pipeline.yml
dhaustein: We should address this in a followup PR, I am lumping too much stuff into this one PR as it is

frontend/pipeline.yml
mzedel: IMO you calling this as hardcoded values misses the point as the value is given by the frontend path in the repo.
We can avoid the cd-ing (and you may have noticed I haven't commented on any of it), but that's once again independent and would have to be in a separate commit as the commit title doesn't match the change (to rephrase your words: git 101).
Indeed having this in a variable is what's the risk and we can avoid them from showing up as configuration by not introducing the option - especially as it's not a nested directory as the e2e tests dir would be. My bigger gripe is that the value isn't debatable - even if you set `FRONTEND_DIR` instead of `frontend` - it will have to be correctly spelled and that's it. If the job for some reason should not start in our repo root - that would be an issue and will have to be addressed in any case.
Considering my own _"years of building CI/CDs"_ I can however confirm that this was never an issue causing brittleness in this part of the CI - but OTOH we *did* have several instances where CI variable definitions got stale (e.g. job image tags get modified less frequently, as their definition is no longer close to the job) or disconnected from their usage. Of course there are trade offs to these things, but it's also in no way a clear benefit (also something we experienced e.g. with job images).
I would be curious to hear if you are aware of information that shows something different - otherwise I would consider this to be a non-issue in our (non-Jenkins, non-prior working place) case and thus consider this argument not applicable.

frontend/pipeline.yml
mzedel: There are no logs stored (anymore since a change to the testing setup), thus the error message you seem to address here - as the change has been a while ago and no one seemed to have relied on the logs => we can get rid of treating the logs as artifacts.

frontend/pipeline.yml
mzedel: Sure, that would be a good practice. So remove the related changes from this PR as it conflates both and create a followup PR or task. Or separate the different aspects in separate commits as I requested.

frontend/pipeline.yml
mzedel: If you would only have the `FRONTEND_DIR` this would be the case, but here I refer to the redefinition of the coverage, screenshots, ... - adding the indirection to them is what's risky here and I would very much avoid it.
Contrary to your noted "hardcoding the same values all over", these are set only here - and are directly related to the job, so local values - to rephrase your words: putting local only values in a global variable is just poor programming practice 101. 🙂

frontend/pipeline.yml
mineralsfree: The warnings Dusan tried to resolve are coming from `react-testing-library`.  
I've tried the workarounds listed in the issue, but none of them worked consistently. Maybe we can revisit this if it starts causing problems.  
Context: https://github.com/testing-library/react-testing-library/issues/1061

frontend/pipeline.yml
dhaustein: I disagree, I still insist this is a good practice and there is no point in having the value copy pasted multiple times.

As for the risk of someone changing it in a manually run pipeline, I think that risk is minimal. And if anything, we can do https://docs.gitlab.com/ci/pipelines/#configure-a-list-of-selectable-prefilled-variable-values But again, I don't think people here will be so reckless as to change the variable when they run the pipeline manually.

frontend/pipeline.yml
dhaustein: So to rephrase what you said - remove it, because we don't generate logs, and we don't want the file stored among the artifacts?

frontend/pipeline.yml
dhaustein: I don't understand. How am I re-defining the file paths for screenshots, coverage etc?

frontend/pipeline.yml
mzedel: Again, the past has shown the problem to be different. If you want to as claimed "heavily rely on the opinion of the team" - this would be a place to do so.

frontend/pipeline.yml
mzedel: Yes. There exists no file to be stored + unlike the job log from within Gitlab, any log stored as artifact would be removed after the `expire_in: 2w` time.

frontend/pipeline.yml
mzedel: You set it at the top of the CI file.

frontend/pipeline.yml
dhaustein: Alright, thank, I'll remove this logging

frontend/pipeline.yml
dhaustein: But the file path remain the same?

mzedel: It is the added level of indirection - it doesn't solve anything and adds an extra location where it might diverge. The main point being that such indirections have diverged in the past - contrary the paths here stayed aligned.

frontend/pipeline.yml
```diff
@@ -6,6 +6,11 @@ variables:
   DOCS_VERSION: development
   MENDER_IMAGE_GUI: ${MENDER_IMAGE_REGISTRY}/${MENDER_IMAGE_REPOSITORY}/gui:${MENDER_IMAGE_TAG}
   FF_TIMESTAMPS: true
+  FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
+  LOGS_DIR: ${CI_PROJECT_DIR}/logs
+  COVERAGE_DIR: ${FRONTEND_DIR}/coverage
+  SCREENSHOTS_DIR: ${FRONTEND_DIR}/screenshots
+  JUNIT_DIR: ${FRONTEND_DIR}
```
alfrunes: How about instead of defining 5 different variables that points to frontend artifacts, consolidate everything into a `FRONTEND_ARTIFACTS_DIR` pointing to the root of all artifacts and update the jobs the scripts that generate these artifacts to write to sub-directories inside this path?
In that case it doesn't matter what you define the variable as and there will be less boilerplate dealing with the correct environment variables.

```suggestion
  FRONTEND_ARTIFACTS_ROOT: "${CI_PROJECT_DIR:-$(git rev-parse --show-toplevel)}/frontend/artifacts"
```

frontend/pipeline.yml
alfrunes: I wouldn't expect globals defined in this way as intended to be "configurable" unless it has a description. However, I can also see that defining so many variables based on a fixed repository structure could lead to problems if trying to execute the pipeline scripts locally (`$FRONTEND_DIR` would expand to `/frontend` if not defining built-in CI variables). OTOH changing directory inside pipeline scripts or using local paths is a similar hurdle as it makes the scripts non-repeatable and dependent on current working directory.

frontend/pipeline.yml
```diff
@@ -202,21 +214,21 @@ build:frontend:docker:
...
+        - ${JUNIT_DIR}/results.xml
     when: always
   tags:
     - hetzner-amd-ax42
 
 test:frontend:acceptance:
   extends: .template:test:frontend:acceptance
   script:
-    - GUI_REPOSITORY=$(pwd)/frontend SERVER_ROOT=$(pwd) ./frontend/tests/e2e_tests/run
+    - GUI_REPOSITORY=${FRONTEND_DIR} SERVER_ROOT=${CI_PROJECT_DIR} ${FRONTEND_DIR}/tests/e2e_tests/run
```
alfrunes: I think the script itself actually defines these variables as long as it's executed within the git repository.
```suggestion
    - tests/e2e_tests/run
```

alfrunes (commented):
Just a few observation. Sorry I unsubscribed from this issue due to lack of frontend expertise.
+1 for removing the job-local directories
I'm not sure what would be the best approach for the number of new variables introduced, but I would personally prefer keeping it to a minimum.
It would also be great if you could split the different changes into separate commits that more accurately describe the changes.

frontend/pipeline.yml
dhaustein: I like this approach!

It does rely on git, and I am not sure how it's going to work for things like shallow clones, but I also cannot see the $CI_PROJECT_DIR to not have a value. Something must've gone very wrong if that happened. Either way, a cool fallback.

frontend/pipeline.yml
dhaustein: It does, I missed that, thx. But looking at it now, they are default/fallback values:

```bash
# frontend/tests/e2e_tests/run
export SERVER_ROOT=${SERVIER_ROOT:-$(git rev-parse --show-toplevel)}  # <-- I'll fix the typo in another commit
export GUI_REPOSITORY=${GUI_REPOSITORY:-${SERVER_ROOT}/frontend}
export COMPOSE_FILE="${SERVER_ROOT}/docker-compose.yml:${GUI_REPOSITORY}/tests/e2e_tests/docker-compose.e2e-tests.yml"
```

I will remove setting the vars again from the CI, since they will be the same like you said. But I guess we do want to keep the option to override the ones in the `run` script for local running?

frontend/pipeline.yml
dhaustein: I looked into the artifacts root path changes and it won't be as straightforward. 

I'd need to add something akin to `path.join(frontendDir, 'artifacts');` into `vitest.config.ts` and `playwright.config.ts` because they are set to output to their root which would be `FRONTEND_DIR` and not `FRONTEND_DIR/artifacts` 

I'd rather not change it this way, unless we really really want it.

I can go back to a partial approach. Keep the ` FRONTEND_DIR` variable
```yaml
variables:
  FRONTEND_DIR: ${CI_PROJECT_DIR:-$(git rev-parse --show-toplevel)}/frontend
```
but don't do anything extra for the different artifacts dirs. And for the job artifacts just do:
```yaml
artifacts:
    paths:
      - ${FRONTEND_DIR}/coverage
      - ${FRONTEND_DIR}/screenshots
      - ${FRONTEND_DIR}/junit
```

What do you think?

frontend/pipeline.yml
dhaustein: done

frontend/pipeline.yml
dhaustein: done

dhaustein:
Hi @kjaskiewiczz @merlin-northern as discussed on the server meeting, I've broken down these changes into hopefully cleaner commit structure.

dhaustein:
Working on fixing the CI checks :skull:

frontend/pipeline.yml
dhaustein: Actually turns out that
```yaml
FRONTEND_DIR: ${CI_PROJECT_DIR:-$(git rev-parse --show-toplevel)}/frontend
```
won't work because [gitlab will not do command substitution](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/10904864530), so I changed it to simply 
```yaml
FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
```

vpodzime (commented):
The `feat` commits are misleading here -- the changes are not really features of **mender-server**. It would be OK if this was a tests repository, but we don't want to present these changes are new features in Mender Server, right? 

Looks good to me otherwise.

frontend/pipeline.yml
```diff
@@ -18,10 +19,9 @@ test:frontend:lint:
         compare_to: '${RULES_CHANGES_COMPARE_TO_REF}'
   needs: []
   script:
-    - cd frontend
-    - npm ci --cache .npm --prefer-offline
-    - cd tests/e2e_tests && npm ci && cd ../..
-    - npm run lint
+    - npm ci --prefix ${FRONTEND_DIR} --cache ${FRONTEND_DIR}/.npm --prefer-offline
+    - npm ci --prefix ${FRONTEND_DIR}/tests/e2e_tests
+    - npm run lint --prefix ${FRONTEND_DIR}
```
merlin-northern, line 24: okey so I do not want to attack (too much) you, but this couple of lines is my clipboard, so I come here to copy from when I forget how to run the ci for the Frontend. and if I have to remove that many chars from the command, I will have rather re-typed it. so all in all I do not really see the benefit of the FRONTEND_DIR variable. why did you introduce it?

dhaustein: Doing a `cd dir` in a ci is considered an anti-pattern, it's moving the pointer around and all the scripts assume a certain directory structure. Luckily all the tooling has some way to force a "workdir" so it can be defined once and re-used. Also when working on the CI you have to keep that in your head too. 

If you rely on the CI code for your local needs, the code will still work the same, you just need to ensure you have the new env var. 

And if you are copy pasting it every time, maybe it's time to create a script. :grin:

dhaustein:
> The `feat` commits are misleading here -- the changes are not really features of **mender-server**. It would be OK if this was a tests repository, but we don't want to present these changes are new features in Mender Server, right?
> 
> Looks good to me otherwise.

@vpodzime Yeah I wasn't sure how this team approaches features in and around testing. I will reword the commits.

frontend/pipeline.yml
dhaustein: addressed the vars, should be a bit more streamlined now, more details in the thread below

frontend/pipeline.yml
dhaustein: @alfrunes What's your take on he changes now?

frontend/pipeline.yml
mzedel: @dhaustein your question didn't answer the question to the _why_ of the variable, but a _why_ to the (undisputed) `--prefix`. Please come up with a reasonable explanation to the _why_ of the variable.

As to best practices, it is also a best practice to reduce complexity where possible - in comparison a single level folder definition that is not only shorter than the variable definition, but moreover won't ever be an issue getting inherited in child pipelines/ expanded in a script/ is clearly readable without knowing the full pipeline definition.

And since this is repeatedly ignored here: the CI instructions are not only used by us internally and if we want to add the extra indirection we should have a *good* reason - so far I haven't seen any.

dhaustein: I believe the why has been discussed at least thrice now. 

This is decreasing complexity, like I said to a comment to Peter, please read it.

I agree that in an ideal world the CI and the local would work the same. Alas we are not there yet and probably never will be 100%. Copying over lines from a CI script so you can run stuff locally, instead of relying on, I dunno, makefiles? is also a symptom of bad design.

Additionally, to this. Since I am in charge of the test automation, I have the final say. If you have a problem with this, talk to Marcin or Thomas.

mzedel: @dhaustein you should re-read the company handbook on the decision making process and I'm talking with both to hopefully prevent more of these discussions with you.

So how can an added layer of indirection decrease complexity? Now you have to do a lookup of: script => line => variable => line -- in comparison to: script => line ... we clearly must have a diverging counting system. So no, your why has not been discussed in a convincing way.
And again it's to avoid adding steps also to external consumers of the CI file - adding additional dependencies (such as make) for them to build parts of the application is in contrast to this.

kjaskiewiczz:
I'm not an expert here and I don't have a strong opinion.
For sure I don't like the temperature of the discussion. Is the change worth it? Are we only discussing `frontend` vs `${FRONTEND_DIR}`?

If I understand correctly @dhaustein wants to standardize things, which is good, and @mzedel is trying to avoid issues, which is also good.

As for the arguments:

> mitigating bad practice of hardcoding the same values all over and extracting them into one single source of truth, this is programming 101

In programming I think this would be a good case for using const, but there are no consts in gitlab. Still I get the idea of having it organized this way.

> Indeed having this in a variable is what's the risk and we can avoid them from showing up as configuration by not introducing the option - especially as it's not a nested directory as the e2e tests dir would be. My bigger gripe is that the value isn't debatable - even if you set FRONTEND_DIR instead of frontend - it will have to be correctly spelled and that's it.

I agree that if we see the risk in the change and we've seen some issues caused by similar approach in the past I would skip this change.

So as for this particular change/commit +1 for replacing cd with prefix and -1 for replacing `fronted` with `${FRONTEND_DIR}`, but as I said I'm not an expert here and I wouldn't relay on my opinion when making final decision.

frontend/pipeline.yml
dhaustein: @merlin-northern I could add the new env var to https://github.com/mendersoftware/mender-server/blob/main/.env And if you use the likes of direnv, you would be able to copy and use the CI commands withouth having to think about the env var. Would that work dor you?

oldgiova:
My two cents here; as for my experience when testing CI pipelines locally, I found very valuable the [direnv project](https://direnv.net/): by having a `.envrc` with: 
```
export FRONTEND=frontend
```
would solve both issues: you can run the same code from your terminal as in CI, and you can keep Gitlab best practices.

This makes direnv as a requirements for the project, though.. Fell free to ignore my 2 cents

mzedel:
@oldgiova s mention of 

>  Gitlab best practices.

made me think, maybe there is guidance from them on their CI - and turns out it's even better: their CI is of course open source. So you can go straight to their frontend file for an example on how they consider their CI to be used and they even share the same _frontend_ name for their frontend things: https://gitlab.com/gitlab-org/gitlab/-/blob/master/.gitlab/ci/frontend.gitlab-ci.yml?ref_type=heads#L347 - turns out they use variables for things to be reassigned during the pipeline.
So keeping the paths seems to align with the tool standard and is arguably considered best practice there.

I highly recommend anyone interested, but @dhaustein in particular to take a look through the [main gitlab CI file](https://gitlab.com/gitlab-org/gitlab/-/blob/master/.gitlab-ci.yml?ref_type=heads) and the corresponding [includes folder](https://gitlab.com/gitlab-org/gitlab/-/tree/master/.gitlab/ci?ref_type=heads).

frontend/pipeline.yml
mzedel: @dhaustein applying your own logic here: the contents of the script are frontend team code and as such the final say is: that change is poor, please adjust or create a follow up task.

dhaustein: we will be defining a QA role and who is responsible for what exactly, hopefully next week, until then we have the same role

dhaustein:
> @oldgiova s mention of
> 
> > Gitlab best practices.
> 
> made me think, maybe there is guidance from them on their CI - and turns out it's even better: ...

But that is exactly what I am trying to do with
```
FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
```

mzedel:
putting it into a picture to avoid confusion, but how is this:

Screenshot 2025-08-15 at 12 34 15

relying on an env var for the path in the way that this is:

> ```
> FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
> ```

frontend/pipeline.yml
mzedel: Interesting...
Until then we still have this PR opened before the fact and as the person mainly responsible for the code in here, I still insist on this being addressed before this can get merged or put into a follow up task, since the change as is, is considered bad practice and thus nothing we want in the codebase.

dhaustein:
@mzedel 
> relying on an env var for the path in the way that this is:
> 
> > ```
> > FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
> > ```

I thought you were referring to the likes of
```
RSPEC_PROFILING_FOLDER_PATH: rspec/profiling
```
or the other `*_PATH`  vars which are then re-used throughout.

The primary change in this PR is me trying to extract the repeated pattern of `${CI_PROJECT_DIR}/frontend` and in doing so get rid of `cd frontend/` I actually don't care if this is an env var in the end or not, but Gitlab doesn't let me have too many choices here in how to do it cleanly.

If your problem is that this would create a new variable, we can go with something like:
```
.frontend-dir-var: &frontend-dir-var
  FRONTEND_DIR: ${CI_PROJECT_DIR}/frontend
```
Which is a style Gitlab's CI code uses, too. However, this would mean having an instance of:
```
variables:
  <<: *frontend-dir-var
```
in every job that wants to use this var again.

mzedel:
@dhaustein look again please.
The paths defined in variables are used in CI and not ones reliant on the project structure. The `RSPEC_PROFILING_FOLDER_PATH` is where profiling results gathered from CI runs go, something like `GLCI_ASSETS_TAG_ENV_FILE_PATH` that is generated during a CI job, temp paths like `WEBPACK_COMPILE_LOG_PATH` - OTOH static paths that exist in the repo are referenced in full.

It's quite consistent among the files (e.g. https://gitlab.com/gitlab-org/gitlab/-/blob/master/.gitlab/ci/gitlab-gems.gitlab-ci.yml and the 15 repetitions of the `.gitlab/ci/templates/` folder, or the _countless_ references to the `script` folder in all of the files, ...) so I would assume it's a common practice used by the people actually developing the tool - so likely recommended to stick to it.

dhaustein:
@mzedel Yes, however the usage of variables in includes is limited https://docs.gitlab.com/ci/yaml/includes/#use-variables-with-include 

They cannot replace the common filepath with a var, not within the job at least, they'd have to define it in the global CI/CD settings which would hide the value to a completely different place.

On our end for example - our yocto pipeline hinges on `$WORKSPACE` for similar reasons. I went over the history in mender-qa and clearly a lot of thought went into its design even before it was ported from Jenkins.

The thing is complex sure, but for one the usage of `$WORKSPACE` makes it, at least, easier to reason about where the commands are running. I want something like that here too.

I know DRY is not a dogma, but generally a good idea. What should I change in this regard so that you are more receptive to these changes?

mzedel:
To the handling in the yocto pipeline... I think that's a very, very different setup, as the client pipeline relies on multiple repositories (getting dependencies as submodules or cloning them, building them inside and going back again) and at least I haven't touched the build steps of any of this (as personally I found both the Jenkins setup & the gitlab setup not at all clear to work in + I lack the client knowledge to make informed decisions there). I'd be happy to have a chat about the reasons for the switch from Jenkins in the first place & things we considered - maybe we can do this once Lluis is back..?
The server side OTOH was adjusted quite a bit to be contained in the repositories it used to live in, so the `$WORKSPACE` situation is intentionally not happening. Any of the consumed dependencies should be built externally - if we still have some, these need to be addressed and externalized IMO.

So my recommendation would rather be to not make steps to the (afaict more Jenkins-aligned) mender-qa setup, but instead decouple mender-qa if possible & align with Gitlab built in features to no longer need the `$WORKSPACE` workaround (IMO that should be handled differently to rely on `CI_PROJECT_DIR`, but again I never dared to touch this, so I'm judging this from a weak knowledge base).

As to the PR here: I think the removal of `cd`s + fixes to existing vars are great & I'd be happy to provide the changes to the `licenseCheck` to make this run under the deno tasks file (+ I'll adjust the env reliance for the e2e tests separately as the `GUI_REPOSITORY` really looks off, coming from the pre-monorepo days). The indirection over the dir variable OTOH is however something I think is risky without an added benefit due to the nature of the value the var replaces.
As to the repo here: the feature based/ PR based deployments would be amazing to have, improving parallelization of the jobs especially in the full pipeline as well, caching!, aligning rules etc. also - I think there are a lot of possibilities...


Regarding what's in the gitlab repo:

> @mzedel Yes, however the usage of variables in includes is limited https://docs.gitlab.com/ci/yaml/includes/#use-variables-with-include

That would address the `gitlab-gems` case, but a look through the gitlab codebase shows that there is _no_ repository path mapped through a variable (ok... there is *one* exception I found not in CI, but in a template: https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/templates/Pages/SwaggerUI.gitlab-ci.yml?plain=1#L29).

pasinskim:
Thanks all. @mzedel @dhaustein let's land the clear wins and pause the contentious part. 

@dhaustein the DRY principle is a guideline, not an absolute law. It is most useful when abstracting values that might change. The name of the main `frontend` source directory is highly unlikely to change. In this context, the explicit path `frontend/` is clearer and more direct than the variable `$FRONTEND_DIR`. Adding the variable makes the configuration more complex to read and debug for no tangible gain. It also make sense to follow the tool's own developers (Gitlab) practices. Other changes, like replacing `cd` with `npm --prefix`, are IMO legitimate improvements.

What I propose is:
- keep switching from cd to --prefix
- keep the Deno/license tweaks and logs cleanup
- revert FRONTEND_DIR; use literal frontend/… paths

Also, please keep the discussion to the technical arguments here and other things (process/ownership) let's discuss elsewhere.

PR #797, opened 2025-07-15
fix(deviceconnect): Send pings to keep the websocket connection alive

bahaa-ghazal:
Ticket: MEN-8439
Changelog: None

### DIFF
--- a/backend/services/deviceconnect/api/http/device.go
+++ b/backend/services/deviceconnect/api/http/device.go
@@ -19,6 +19,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/gin-gonic/gin"
@@ -39,6 +40,8 @@ import (
 )
 
 var (
+	// Time allowed to read the next pong message from the peer.
+	pongWait = time.Minute
 	// Seconds allowed to write a message to the peer.
 	writeWait = time.Second * 10
 )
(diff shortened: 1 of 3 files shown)


backend/services/deviceconnect/api/http/device.go
```diff
@@ -258,6 +264,11 @@ Loop:
 		case <-ctx.Done():
 			err = errors.New("connection closed by the server")
 			break Loop
+		case <-ticker.C:
```
alfrunes, line 266: I think we can also reset the ticket whenever we receive a message from the peer too as we only want to send ping when the connection is idle.

backend/services/deviceconnect/api/http/device.go
```diff
@@ -230,11 +233,14 @@ func (h DeviceController) connectWSWriter(
...
+	pingPeriod := (pongWait * 9) / 10
+	ticker := time.NewTicker(pingPeriod)
+	defer ticker.Stop()
+	conn.SetPongHandler(func(string) error {
+		ticker.Reset(pingPeriod)
+		return conn.SetReadDeadline(time.Now().Add(pongWait))
+	})
 
 	conn.SetPingHandler(func(msg string) error {
 		if err != nil {
```
alfrunes: This condition is always false: `nil != nil`.
`err` is cumming from the parent scope.

backend/services/deviceconnect/api/http/device.go
```diff
@@ -230,11 +233,14 @@ func (h DeviceController) connectWSWriter(
...
+	// send periodic ping messages to keep the connection alive
+	pingPeriod := (pongWait * 9) / 10
+	ticker := time.NewTicker(pingPeriod)
+	defer ticker.Stop()
+	conn.SetPongHandler(func(string) error {
+		ticker.Reset(pingPeriod)
+		return conn.SetReadDeadline(time.Now().Add(pongWait))
+	})
 
 	conn.SetPingHandler(func(msg string) error {
```
alfrunes, line 245: Also on ping we can defer sending ping when the connection is idle.
```suggestion
	conn.SetPingHandler(func(msg string) error {
		ticker.Reset(pingPeriod)
```

alfrunes (approved):

PR #798, opened 2025-07-15
MEN-8412, MEN-8303: new self-service subscription page + e2e tests

mineralsfree:


frontend/src/js/components/settings/CardSection.tsx
```diff
@@ -105,16 +107,16 @@ export const CardSection = ({ isSignUp, onClose, organization, onSubmit, onCardC
...
         {!isSignUp && (
           <Button type="reset" disabled={loading} style={{ marginRight: 15 }}>
             Cancel
           </Button>
         )}
         <Button variant="contained" color="secondary" type="submit" disabled={errors || loading || empty || !isValid}>
-          {isSignUp ? 'Sign up' : 'Save'}
+          {isSignUp ? 'Confirm Subscription' : 'Save'}
         </Button>
       </div>
```
mzedel: to align with the design I would suggest 
```suggestion
<div className="flexbox center-aligned margin-top-small">
<Button type="reset" disabled={loading} style={{ marginRight: 15 }} onClick={onCancel}>
Cancel
</Button>
<Button variant="contained" color="secondary" type="submit" disabled={errors || loading || empty || !isValid}>
          {isSignUp ? 'Confirm Subscription' : 'Save'}
</Button>
</div>
```
as the components current uses will be phased out soon anyway + the missing click handler on the button wasn't raised so far 🙈

frontend/src/js/components/subscription/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+    idMap[line.price_id].total += line.amount;
+  });
+
+  const items = Object.values(idMap);
+  const planPrice = items.find(p => p.isPlan)?.total || 0;
+  const addons = items.filter(p => !p.isPlan).reduce((acc, curr) => ({ ...acc, [curr.title.toLowerCase()]: curr.total }), {});
+  return { plan: planPrice, addons };
+};
+export const currencyFormatter = new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD', minimumFractionDigits: 0 });
```
mzedel, line 30: considering that this is called with `toBeFormatted / 100` every time - what would you think about having the formatter call wrapped?

frontend/src/js/config/routes.tsx
```diff
@@ -42,7 +43,8 @@ export const routeConfigs: RouteConfigs = {
   releases: { path: 'releases', element: <Releases />, title: 'Releases' },
   settings: { path: 'settings', element: <Settings />, title: 'Settings' },
   signup: { path: 'signup', element: <Signup />, title: 'Tenants', isPublic: true },
-  tenants: { path: 'tenants', element: <TenantPage />, title: 'Tenants' }
+  tenants: { path: 'tenants', element: <TenantPage />, title: 'Tenants' },
+  subscription: { path: 'subscription', element: <SubscriptionPage />, title: 'Upgrade your subscription' }
```
mzedel:
```suggestion
  subscription: { path: 'subscription', element: <SubscriptionPage />, title: 'Upgrade your subscription' },
  tenants: { path: 'tenants', element: <TenantPage />, title: 'Tenants' }
```
just to keep things sorted there.

and: 🤫 the commit order isn't fully supportive of the `<SubscriptionPage />` introduction here 😉

frontend/src/js/store/organizationSlice/index.ts
```diff
@@ -35,6 +35,7 @@ export const initialState: OrganizationState = {
     }
   },
   organization: {
+    isLoaded: false
     // id, name, status, tenant_token, plan
```
mzedel: would this not be covered by checking for a set `id` or `name` too? I would assume in case of a successful response, but wrong data the `isLoaded` indicator won't be of much help either...?

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -284,6 +292,26 @@ export const getUserBilling = createAsyncThunk(`${sliceName}/getUserBilling`, (_
   Api.get(`${tenantadmApiUrlv2}/billing/profile`).then(res => dispatch(actions.setBillingProfile(res.data)))
 );
 
+export const getUserSubscription = createAsyncThunk(`${sliceName}/getUserSubscription`, (_, { dispatch }) => {
+  const tasks = [dispatch(getBillingPreview({ preview_mode: 'next' })).unwrap(), dispatch(getCurrentSubscription()).unwrap()];
+  Promise.all(tasks).then(([currentPreview, currentSubscription]) => dispatch(actions.setSubscription({ ...currentPreview, ...currentSubscription })));
+});
```
mzedel, line 300: can you explain why you have to get both of these in the initialization hook? especially if you have not signed up yet, the 'next' preview request seems to fail and afaict there's no immediate use of the data in the UI
If it's not really needed I would otherwise fall back to fetch the information when showing the page/ when needed

frontend/src/js/store/organizationSlice/index.ts
```diff
@@ -85,7 +86,10 @@ export const organizationSlice = createSlice({
       state.intentId = action.payload;
     },
     setOrganization: (state, action) => {
-      state.organization = { ...state.organization, ...action.payload };
+      state.organization = { ...state.organization, ...action.payload, isLoaded: true };
+    },
+    setSubscription: (state, action) => {
+      state.organization.subscription = { ...state.organization.subscription, ...action.payload };
```
mzedel, line 91: this will have to go into the `Organization` interface too

frontend/src/js/store/storehooks.ts
```diff
@@ -160,6 +161,7 @@ export const useAppInit = userId => {
     const multitenancy = hasMultitenancy || isHosted || isEnterprise;
     if (multitenancy) {
       tasks.push(dispatch(getUserOrganization()));
+      tasks.push(dispatch(getUserSubscription()));
     }
```
mzedel: since anything but hosted won't have the information accessible:
```suggestion
const multitenancy = hasMultitenancy || isHosted || isEnterprise;
if (multitenancy) {
tasks.push(dispatch(getUserOrganization()));
}
if (isHosted) {
      tasks.push(dispatch(getUserSubscription()));
}
```
but apart from this, I don't think we have to get this info here...

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,247 @@
...
+      await dispatch(createBillingProfile({ billingProfile: billing_profile }));
+    }
+    setIsEdit(false);
+  };
+  const upgradePlanSubmit = async () => {
+    try {
+      setLoading(true);
+      setError('');
+      await dispatch(requestPlanUpgrade(order));
+      dispatch(setSnackbar(successMessage((selectedPlan as Plan).name)));
```
mzedel: I would assume the `successMessage` handling can be moved into the `requestPlanUpgrade` too (with the `selectedPlan.id` part of the `order), to have it easier testable in the thunk

frontend/src/js/components/subscription/SubscriptionSummary.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  {isPreviewLoading || !previewPrice.addons[addon] ? <NumberSkeleton /> : currencyFormatter.format(previewPrice.addons[addon] / 100)}
+                </Typography>
+              </div>
+            ))}
+          </div>
+        )}
+        <Divider variant="middle" className="margin-none" />
+        <div className="flexbox space-between margin-top-small">
+          <Typography variant="subtitle1">Monthly price</Typography>
+          <Typography variant="subtitle1">{isPreviewLoading ? <NumberSkeleton /> : currencyFormatter.format(previewPrice.total / 100)} </Typography>
```
mzedel:
```suggestion
          <Typography variant="h5">{isPreviewLoading ? <NumberSkeleton /> : currencyFormatter.format(previewPrice.total / 100)} </Typography>
```

frontend/src/js/components/subscription/SubscriptionSummary.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </div>
+      </CardContent>
+      {!readOnly && (
+        <Button className="margin-top-small" sx={{ textTransform: 'none' }} disabled={!isNew} variant="contained" onClick={onAction} fullWidth>
```
mzedel: The `textTransform` should be handled on the theme level (as part of https://northerntech.atlassian.net/browse/MEN-8530) 👌 - thus we don't want to have custom handling spread out across the codebase...

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+      preview_mode: 'recurring',
+      plan: selectedPlan.id,
+      products: [{ name: 'mender_standard', quantity: limit, addons }]
+    };
+    setOrder({ plan: order.plan, products: order.products });
+
+    dispatch(getBillingPreview(order))
+      .unwrap()
+      .then(invoice => {
+        const previewPriceParsed = parseSubscriptionPreview(invoice.lines, selectedPlan.name.toLowerCase());
```
mzedel: To ensure the parsing is tied to the (potentially changing) backend response, I would rather see this in the `getBillingPreview` thunk. You could either use the plan names from the constants to check or maybe add the stripe `price_id` to the plans and match with this?

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+  }, [dispatch, selectedPlan, debouncedLimit, JSON.stringify(selectedAddons), isOrgLoaded]);
+
+  const onChangePlan = planId => {
+    //we need to reset unavailable addons from selection
+    const unavailableAddons = Object.keys(selectedAddons).filter(addonId => !ADDONS[addonId].eligible.includes(planId));
+    const newAddons = { ...selectedAddons, ...unavailableAddons.reduce((acc, addon) => ({ ...acc, [addon]: false }), {}) };
+    setSelectedAddons(newAddons);
+    setSelectedPlan(PLANS[planId]);
+    setContactReason('');
+    setLimit(PLANS[planId].minimalDeviceCount);
```
mzedel: This will cause the form to remove any previously entered device count values - maybe you can check if the current limit is too low and only then set it?

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+            <RadioGroup
+              row
+              aria-labelledby="plan-selection"
+              name="plan-selection-radio-group"
+              value={selectedPlan ? selectedPlan.id : null}
+              onChange={(_, value) => onChangePlan(value)}
+            >
+              <FormControlLabel disabled={isBasicDisabled} value="os" control={<Radio />} label="Basic" />
+              <FormControlLabel disabled={isProfessionalDisabled} value="professional" control={<Radio />} label="Professional" />
+              <FormControlLabel value="enterprise" control={<Radio />} label="Enterprise" />
```
mzedel: the values and labels shoudl be taken from the `PLANS` constant and I would imagine this to be a reasonable spot for a loop over the values in there (then you could also get rid of the `isBasicDisabled` & `isProfessionalDisabled` as you can use the order of the plans + `isTrial` as an indicator

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const stripeAPIKey = useSelector(getStripeKey);
+  const [isPreviewLoading, setIsPreviewLoading] = useState(false);
+  const [showUpgradeDrawer, setShowUpgradeDrawer] = useState(false);
+  const [loadingFinished, setLoadingFinished] = useState(!stripeAPIKey);
+
+  const dispatch = useAppDispatch();
+  const currentDeviceLimit = useSelector(getDeviceLimit);
+  const org = useSelector(getOrganization);
+  const { addons: orgAddOns = [], plan: currentPlan = PLANS.os.id as AvailablePlans, trial: isTrial = true, isLoaded: isOrgLoaded } = org;
+  const plan = Object.values(PLANS).find(plan => plan.id === (isTrial ? PLANS.os.id : currentPlan));
```
mzedel:
```suggestion
  const plan = Object.values(PLANS).find(plan => plan.id === (isTrial ? PLANS.os.id : currentPlan)) || PLANS.os;
```
to be in line with the `currentPlan` assignment above and to avoid the need for the repeated checks for a set plan further down?

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+    enabledAddons.forEach(addon => {
+      newSelectedAddons[addon.name] = addon.enabled && !isTrial;
+    });
+    setSelectedAddons(newSelectedAddons);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [enabledAddons]);
+
+  useEffect(() => {
+    if (debouncedLimit >= enterpriseDeviceCount) {
+      setContactReason(`For over ${enterpriseDeviceCount} devices, please contact for pricing.`);
```
mzedel: I assume the contact should also point to support here 😉

more general: the contact method should be clickable, so I guess you could put the different `contactReason`s in a tiny component with a `SupportLink` where applicable and either have this directly tracked as `contactReason` or just the key and pick the proper message in the component...

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+      setLimit(currentDeviceLimit);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [plan, currentDeviceLimit]);
+
+  useEffect(() => {
+    const newSelectedAddons: Record<AvailableAddon, boolean> = {} as Record<AvailableAddon, boolean>;
+    enabledAddons.forEach(addon => {
+      newSelectedAddons[addon.name] = addon.enabled && !isTrial;
+    });
```
mzedel: this looks a lot like a `reduce` to me 🙈

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+      return;
+    }
+    const addons = Object.entries(selectedAddons)
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .filter(([_, value]) => value)
+      .map(([key]) => ({ name: key }));
+    setIsPreviewLoading(true);
+    const order = {
+      preview_mode: 'recurring',
+      plan: selectedPlan.id,
```
mzedel, line 189: unfortunately I don't get a successful preview response when trying this locally (starting from os/basic => to professional => back to basic), unless I add the following:
```suggestion
    plan: selectedPlan.id === PLANS.os.id ? 'basic' : selectedPlan.id,
```
I don't think that's intentional on the backend side... will ping someone there 👌

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+                    onBlur={handleDeviceLimitBlur}
+                    slotProps={{ htmlInput: { min: selectedPlan.minimalDeviceCount, step: DIVISIBILITY_STEP } }}
+                    value={limit}
+                    fullWidth
+                  />
+                </div>
+                <FormHelperText className="info margin-top-none">{inputHelperText}</FormHelperText>
+              </FormControl>
+              {contactReason && (
+                <Alert severity="info" className="margin-bottom-x-small">
```
mzedel: looking at the design I think a `x-small` margin at the top too

frontend/src/js/components/subscription/SubscriptionSummary.tsx
```diff
@@ -0,0 +1,90 @@
...
+const NumberSkeleton = () => <Skeleton width={35} height={26} />;
+
+export const SubscriptionSummary = (props: SubscriptionSummaryProps) => {
+  const { plan, deviceLimit, addons, title, isNew, isPreviewLoading, readOnly, onAction, previewPrice } = props;
+  const enabledAddons = Object.entries(addons)
+    .filter(([addon, enabled]) => enabled && addon)
+    .map(([addon]) => addon);
+
+  return (
+    <Card variant="outlined" className="padding" sx={{ minWidth: '320px' }}>
```
mzedel: looking at the design, in the `readOnly` variant this shouldn't have padding or an outline

mzedel (changes requested):
Looks quite nice already 👍.
I came across some misalignments with the design, but considering these have changed in the past I would rely on you having the proximity to the design people involved 😉.
As to how to reach the page, looking at the designs I assume this is the new target for the `Upgrade Now` button in the page header + any links from the billing section in the settings to a plan upgrade would also point to this new page for "new" tenants (there's a distinction made between old and new in the `Billing/OldTenantPriceIncreaseNote` component).

As a general observation I would prefer for you to stick to the style prop if possible or at least aim for consistency in the way you pass it - if you really want to use the sx prop instead. And the second one would be the absence of tests - especially for the parsing, but also for the interaction within the page... and considering that this will be used to upgrade from trial to a plan in the staging e2e tests, these need to be adjusted as well.
Let me know if I can help you with any of this!

Oh and another general observation: the multiple loops instead of a reducer function look disorienting - but I guess that's a me problem 😅

frontend/src/js/components/subscription/SubscriptionPage.tsx
mineralsfree: no need to ping, Alf is aware of the issue, it is solved with stripe compose config I think :)  I forgot to warn you, sorry

mineralsfree:
There are in-progress follow-up design changes to the subscription flow :sauna_man:. Therefore, I would wait for those to be complete before implementing interaction tests to avoid double work.

I've added a test for the parsing function.

Regarding e2e tests, I think those changes could be made within the scope of MEN-8303?

For the reduce function, I think we should value readability here. While many things are achievable with reduce, it can be cognitively more taxing to read the code afterward.

I would also link to this new page as soon as we follow up on design changes and add tests, so for now, I didn't change it from the old flow

mzedel:
> There are in-progress follow-up design changes to the subscription flow 🧖‍♂️. Therefore, I would wait for those to be complete before implementing interaction tests to avoid double work.
> Regarding e2e tests, I think those changes could be made within the scope of MEN-8303?

We don't know when the follow ups will be scheduled for work, nor whether they will be followed up by more, but we do know that the current task is to be worked on. Additionally the design shows the new subscription page as a replacement for the "status quo" current upgrade page - that means the new page is supposed to take over the duty of the current page (with the caveat of the existing/ non-migrated tenants). Combined this then means we should have staging using the new page once this is merged, which requires e2e adjustments - otherwise we won't be able to deploy this.
As to MEN-8303 - that's to extend the existing tests (as the existing tests are expected to run with the new page already).

> For the reduce function, I think we should value readability here. While many things are achievable with reduce, it can be cognitively more taxing to read the code afterward.

That's what I commented on earlier - to me multiple loops look very much less readable (IOW readability can be highly subjective) and are prone to invite state modifications as an unwanted side effect.

frontend/src/js/components/subscription/SubscriptionPage.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const ui = <SubscriptionPage />;
+    render(ui, {
+      preloadedState: {
+        ...defaultState,
+        devices: { ...defaultState.devices, limit: 10 },
+        organization: { ...defaultState.organization, organization: { ...defaultState.organization.organization, trial: true } }
+      }
+    });
+    expect(screen.getByText('Upgrade your subscription')).toBeInTheDocument();
+    const professionalRadioButton = screen.getByRole('radio', { name: new RegExp(PLANS.professional.id, 'i') });
```
mzedel: The accessible `name` here is derived from the `label` of the buttons afaik, which OTOH is set with the `name`:
```suggestion
    const professionalRadioButton = screen.getByRole('radio', { name: new RegExp(PLANS.professional.name, 'i') });
```
works for anything, but 'os'/ 'basic' - but might stop working once the tinier compute units get introduced...

frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+    await user.type(cityInput, 'Oslo');
+    await act(async () => await user.type(zipInput, '5678'));
+
+    const countryAutoComplete = screen.getByRole('combobox', { name: /country/i });
+    await act(async () => {
+      await user.type(countryAutoComplete, 'Polan');
+      await user.keyboard('[ArrowUp]');
+      await user.keyboard('[Enter]');
+    });
+    await act(async () => await user.click(screen.getByRole('button', { name: /Save Billing profile/i })));
```
mzedel: is the `click` not already wrapped in an `act`?

frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+    expect(input.value).toEqual('Poland');
```
mzedel, line 167: is there a reason this is after the `click` on save? I could imagine the input being gone after saving the profile

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { BillingDetails } from '../settings/organization/BillingDetails';
+import OrganizationPaymentSettings from '../settings/organization/OrganizationPaymentSettings';
+import { PreviewPrice } from './SubscriptionPage';
+import { SubscriptionSummary } from './SubscriptionSummary';
+import { currencyFormatter } from './utils';
+
+interface SubscriptionDrawerProps {
+  addons: Record<AvailableAddon, boolean>;
+  isTrial?: boolean;
+  onCloseClick: () => void;
```
mzedel: maybe align with MUI to `onClose`?

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+  const currentSubscription = useSelector(getSubscription);
+  const initialValues = { email, name: organization?.name || '', line1: '', state: '', city: '', postal_code: '', country: '' };
+  const [formInitialValues, setFormInitialValues] = useState(initialValues);
+  const [isValid, setIsValid] = useState(false);
+  const [isEdit, setIsEdit] = useState<boolean>(false);
+  const [nextPayment, setNextPayment] = useState(0);
+  const [updatingCard, setUpdatingCard] = useState(false);
+  const dispatch = useAppDispatch();
+  const formSubmitRef = useRef<() => void | null>(null);
+  const [error, setError] = useState('');
```
mzedel: considering that the error is defined through the `Alert` I think the error should rather be a boolean

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+              isSignUp
+            />
+          </div>
+        )
+      )}
+      {(!isTrial || (isTrial && !billing)) && summary}
+      {nextPayment > 0 && currentSubscription ? (
+        <>
+          <p>
+            You’re currently subscribed to Mender Basic at {currencyFormatter.format(currentSubscription.total / 100)}/month. <br />
```
mzedel:
```suggestion
            You’re currently subscribed to {PLANS[organization.plan].name} at {formatPrice(currentSubscription.total)}/month. <br />
```

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+            On your next payment, you&#39;ll be charged for any days used under your current <br /> rate, and the rest will be billed at your new subscription
+            rate. The total amount <br /> for your next payment will be {currencyFormatter.format(nextPayment / 100)}.
```
mzedel: instead of the forced line breaks, this should rely on having this wrapped based on the `maxWidth` set for the other parts (there through the `formWrapper`)

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+type ContactReason = 'reduceLimit' | 'overLimit';
+interface ContactReasonProps {
+  reason: ContactReason;
+}
+const ContactReasonAlert = ({ reason }: ContactReasonProps) => {
+  const reasonMap: { [key in ContactReason]: ReactNode | string } = {
+    reduceLimit: `For over ${enterpriseDeviceCount} devices, please contact for pricing.`,
+    overLimit: (
+      <div>
+        If you want to reduce your device limit, please contact <SupportLink variant="email " />
```
mzedel:
```suggestion
        If you want to reduce your device limit, please contact <SupportLink variant="email" />
```
I guess allowing a free form variant there prevents this from getting caught 😦

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+      </div>
+    )
+  };
+  return (
+    <Alert severity="info" className="margin-bottom-x-small margin-top-x-small">
+      {reasonMap[reason]}
+    </Alert>
+  );
+};
```
mzedel: or as an alternative to have the reason map tied to the notification:
```suggestion
const contactReasons = {
  reduceLimit: {
    id: 'reduceLimit',
    alert: (
      <div>
        For over {enterpriseDeviceCount} devices, please contact <SupportLink variant="email" /> for pricing.
      </div>
    )
  },
  overLimit: {
    id: 'overLimit',
    alert: (
      <div>
        If you want to reduce your device limit, please contact <SupportLink variant="email" />.
      </div>
    )
  }
} as const;
interface ContactReasonProps {
  reason: keyof typeof contactReasons;
}
const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
  <Alert severity="info" className="margin-bottom-x-small margin-top-x-small">
    {contactReasons[reason].alert}
  </Alert>
);
```
+ adjusting the `setContactReason` type + calls?

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+      navigate('/settings/billing');
+    }
+  }, [isTrial, navigate, hasCurrentPricing, org]);
+
+  //Fetch Billing profile & subscription
+  useEffect(() => {
+    dispatch(getCurrentCard());
+    dispatch(getUserBilling());
+    dispatch(getUserSubscription());
+  }, [dispatch]);
```
mzedel:
```suggestion
  useEffect(() => {
    if (isTrial) {
      return;
    }
    dispatch(getCurrentCard());
    dispatch(getUserBilling());
    dispatch(getUserSubscription());
  }, [dispatch, isTrial]);
```
as these calls will fail for trial users

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+  useEffect(() => {
+    if (plan) {
+      setSelectedPlan(plan);
+    }
+
+    if (currentDeviceLimit && limit < currentDeviceLimit) {
+      setLimit(currentDeviceLimit);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [plan, currentDeviceLimit]);
```
mzedel:
```suggestion
    setLimit(current => (currentDeviceLimit && current < currentDeviceLimit ? currentDeviceLimit : current));
  }, [plan, currentDeviceLimit]);
```
just to get around the need for the eslint escape

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+                  control={<Radio />}
+                  label={plan.name}
+                />
+              ))}
+            </RadioGroup>
+          </FormControl>
+          <Typography variant="body2" style={{ minHeight: '56px' }}>
+            {selectedPlan.description}
+          </Typography>
+          {selectedPlan.id !== 'enterprise' && (
```
mzedel:
```suggestion
          {selectedPlan.id !== PLANS.enterprise.id && (
```

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [debouncedLimit]);
+
+  useEffect(() => {
+    setInputHelperText(`The minimum limit for ${selectedPlan.name} is ${selectedPlan.minimalDeviceCount}`);
+  }, [selectedPlan]);
+
+  useEffect(() => {
+    if (!isOrgLoaded || selectedPlan.id === 'enterprise' || limit % 50 !== 0) {
```
mzedel:
```suggestion
    if (!isOrgLoaded || selectedPlan.id === PLANS.enterprise.id || limit % DIVISIBILITY_STEP !== 0) {
```

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+                  onChange={onSelectAddon}
+                />
+              ))}
+          </div>
+          {enabledAddons.length > 0 && !isTrial && (
+            <Typography variant="body2" className="margin-bottom">
+              To remove active Add-ons from your plan, please contact <SupportLink variant="email" />
+            </Typography>
+          )}
+          {selectedPlan.id === 'enterprise' && (
```
mzedel:
```suggestion
          {selectedPlan.id === PLANS.enterprise.id && (
```

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+                    fullWidth
+                  />
+                </div>
+                <FormHelperText className="info margin-top-none">{inputHelperText}</FormHelperText>
+              </FormControl>
+              {contactReason && <ContactReasonAlert reason={contactReason} />}
+            </>
+          )}
+          <Typography variant="subtitle1" className="margin-top">
+            {selectedPlan.id === 'enterprise' ? 2 : 3}. Choose Add-ons
```
mzedel:
```suggestion
            {selectedPlan.id === PLANS.enterprise.id ? 2 : 3}. Choose Add-ons
```

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+          </Typography>
+          <div className="margin-top-x-small">
+            {Object.values(ADDONS)
+              .sort((a, b) => a.id.localeCompare(b.id))
```
mzedel: should we maybe reorder the ADDONS definition instead? I see we have the same in `AddOnSelection` (although that will soon be removed)

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+                onClick={() => onEnterpriseRequest({ message: enterpriseMessage })}
+                variant="contained"
+              >
+                Submit request
+              </Button>
+            </>
+          )}
+        </div>
+        <div>
+          {selectedPlan.id !== 'enterprise' && previewPrice && (
```
mzedel:
```suggestion
          {selectedPlan.id !== PLANS.enterprise.id && previewPrice && (
```

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
mzedel: _+_ switch the `currencyFormatter.format` to `formatPrice`

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+    </div>
+  );
+
+  return (
+    <Drawer anchor="right" open={true} PaperProps={{ style: { minWidth: '50vw' } }}>
+      <DrawerTitle title={currentSubscription ? `Upgrade your subscription` : `Subscribe to Mender ${selectedPlan.name}`} onClose={onCloseClick} />
+      <Divider className="margin-bottom" />
+      {selectedPlan && (
+        <div className="margin-bottom">
+          Complete checkout to subscribe to <b>{selectedPlan.name}</b> at <b> {previewPrice ? currencyFormatter.format(previewPrice.total / 100) : ''}</b>
```
mzedel:
```suggestion
          Complete checkout to subscribe to <b>{selectedPlan.name}</b> at <b> {previewPrice ? formatPrice(previewPrice.total) : ''}</b>
```

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+      {nextPayment > 0 && currentSubscription ? (
+        <>
+          <p>
+            You’re currently subscribed to Mender Basic at {currencyFormatter.format(currentSubscription.total / 100)}/month. <br />
+            On your next payment, you&#39;ll be charged for any days used under your current <br /> rate, and the rest will be billed at your new subscription
+            rate. The total amount <br /> for your next payment will be {currencyFormatter.format(nextPayment / 100)}.
+          </p>
+          {error && (
+            <Alert icon={<ErrorOutlineIcon />} severity="error">
+              There was an error while creating the tenant. Please try again, or contact <SupportLink variant="email" />
```
mzedel:
```suggestion
              There was an issue while processing your order. Please try again, or contact <SupportLink variant="email" />.
```

mzedel (changes requested):
Thank you for adding the tests and the adjustments so far @mineralsfree!
IMO the other changes from https://northerntech.atlassian.net/browse/MEN-8644 can still be handled once we have this on staging (+ at least for me locally the backend won't fully change plans after I upgrade). We risk getting further delayed in the feedback cycle otherwise. Feel free to pick up things from there if you like, but IMO there's nothing preventing us from doing this in a next iteration.
A few more comments, nitpicks and questions and two more things needed:
- please push the staging e2e test adjustments for this (IIRC you mentioned you had them already)
- in the `Settings.tsx` the upgrade section needs to depend on `getHasCurrentPricing` too

frontend/src/js/store/organizationSlice/index.ts
mzedel: 🔔 @mineralsfree

frontend/src/js/components/subscription/SubscriptionPage.tsx
mineralsfree: It will remove the eslint issue, but will make it less readable IMO, I'll at least name it `updatedLimit` to make it a bit more clear

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
mineralsfree: Honestly, I don't see big value in it, moreover that would conflict with[ DrawerTitle interface](https://github.com/mendersoftware/mender-server/blob/c0ce47e707e549d3682c8c920db9bd58125ec042/frontend/src/js/common-ui/DrawerTitle.tsx#L20)

frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
mineralsfree: It should be, but removing it triggers missing act warnings in the console. :man_shrugging:

frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
mineralsfree: Tbh, this wasn't ready for review yet :sweat_smile: but thanks for pointing this out :+1:

frontend/src/js/components/settings/organization/Billing.tsx
```diff
@@ -131,7 +133,7 @@ export const DeviceLimitExpansionNotification = ({ isTrial }: { isTrial: boolean
     <div className="muted" style={{ marginRight: 4 }}>
       To increase your device limit,{' '}
     </div>
-    {isTrial ? <Link to="/settings/upgrade">upgrade to a paid plan</Link> : <SupportLink variant="salesTeam" />}
+    {isTrial ? <Link to="/subscription">upgrade to a paid plan</Link> : <SupportLink variant="salesTeam" />}
```
mzedel, line 130: testing this now I noticed how the redirect in the subscription page makes the old upgrade page unreachable for customers still on the old pricing 😕 - I'm afraid the link target needs to be dependent on the `hasCurrentPricing` too.

mzedel (changes requested):
Almost there I think! One more navigation issue with the old pricing tenants and can you add a bit of space at the end of the subscription page (to not have the add on selection that close to the window bottom)?

+ there are again several one line body arrow functions - but by now I think I should either accept it or have the lint rule here as well 🤷🙈 - so no need to spend time on these

mzedel (approved):
👍 thank you! Now let's see what staging says!

PR #799, opened 2025-07-15
ci: switch runners to k8s

oldgiova:
Move most of the gitlab runners to k8s to avoid collision issues with multiple pipelines running at the same time

Ticket: QA-1111


dhaustein (approved):
very cool, looks like it has been fairly painless, looking at the pipelines

oldgiova:
> very cool, looks like it has been fairly painless, looking at the pipelines

One of them retried automatically, because of a runner system failure, though........

PR #802, opened 2025-07-16
fix: Ensure that middlewares are set in the correct order

bahaa-ghazal:
Ticket: MEN-8587

### DIFF
--- a/backend/pkg/routing/routing.go
+++ b/backend/pkg/routing/routing.go
@@ -106,3 +106,16 @@ func NewGinRouter() *gin.Engine {
 
 	return router
 }
+
+// New Gin router without any middlewares
+func NewMinimalGinRouter() *gin.Engine {
+	SwitchToReleaseMode()
+
+	router := gin.New()
+
+	router.HandleMethodNotAllowed = true
+	router.NoMethod(handleNoMethod)
+	router.NoRoute(handleNoRoute)
+
+	return router
+}
--- a/backend/services/deployments/api/http/routing.go
+++ b/backend/services/deployments/api/http/routing.go
@@ -27,6 +27,7 @@ import (
 	"github.com/mendersoftware/mender-server/pkg/contenttype"
 	"github.com/mendersoftware/mender-server/pkg/identity"
 	"github.com/mendersoftware/mender-server/pkg/log"
+	"github.com/mendersoftware/mender-server/pkg/requestid"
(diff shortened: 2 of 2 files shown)


backend/pkg/routing/routing.go
```diff
@@ -106,3 +106,16 @@ func NewGinRouter() *gin.Engine {
...
+	SwitchToReleaseMode()
+
+	router := gin.New()
+
+	router.HandleMethodNotAllowed = true
+	router.NoMethod(handleNoMethod)
+	router.NoRoute(handleNoRoute)
+
+	return router
+}
```
alfrunes, line 121: All endpoints (public or internal) should have accesslogs and requestid.

bahaa-ghazal: Some endpoint have a custom accesslog : https://github.com/mendersoftware/mender-server/blob/ab478c3f25db9024617b3596c68df84922341082/backend/services/deployments/api/http/routing.go#L273

alfrunes: This PR is removing accesslogs entirely from internal endpoints which is not what we want. I don't think it's that important to suppress the from successful health checks.

backend/services/deployments/api/http/routing.go
```diff
@@ -272,11 +275,14 @@ func InternalRoutes(router *gin.RouterGroup, controller *DeploymentsApiHandlers)
 	// Health Check
 	// Skiping logging 2XX status code requests to decrease	noise
 	router.GET(ApiUrlInternalAlive, accesslogErrorsOnly.Middleware,
+		requestid.Middleware(),
 		controller.AliveHandler)
 	router.GET(ApiUrlInternalHealth, accesslogErrorsOnly.Middleware,
+		requestid.Middleware(),
 		controller.HealthHandler)
 
 	router.Use(accesslog.Middleware())
```
bahaa-ghazal, line 284: @alfrunes It still uses the both middlewares in the internal endpoint. If u don't care about skipping logging healthcheck then we need to remove this middleware anyways

alfrunes: Ahh, sorry. I missed that. Thanks for the clarification. Would be great if you left a comment in the code as well as the setup of the routes is now sensitive to the order of setup of the individual routes.

alfrunes (approved):

PR #805, opened 2025-07-17
Revert to absolute API paths and group router by middleware

alfrunes:


backend/services/deployments/api/http/api_deployments.go
```diff
@@ -1021,7 +1021,7 @@ func (d *DeploymentsApiHandlers) createDeployment(
 	id, err := d.app.CreateDeployment(ctx, constructor)
 	switch err {
 	case nil:
-		location := fmt.Sprintf("%s/%s", ApiUrlManagement+ApiUrlManagementDeployments, id)
+		location := fmt.Sprintf("%s/%s", ApiUrlManagementDeployments, id)
```
merlin-northern: `GetDeploymentLocationUrl` should be a function in the http package returning the URL. this is buried here and hard to remember about.

backend/services/deployments/api/http/routing.go
```diff
@@ -113,34 +116,37 @@ func NewRouter(
...
 	//
 	// Encode base64 secret in either std or URL encoding ignoring padding.
 	deploymentsHandlers := NewDeploymentsApiHandlers(
 		ds, new(view.RESTView), app, cfg,
 	)
+	router.Use(requestid.Middleware())
+	accesslogDefault := accesslog.Middleware()
+	accesslogSupressed := accesslog.AccessLogger{
+		DisableLog: func(c *gin.Context) bool {
+			if c.Writer.Status()/100 == 2 {
```
merlin-northern:
```suggestion
			if c.Writer.Status()-200 < 100 {
```
(or something like that I would suggest)
you want explicitly 200-299 range? or 1xx fits return true as well?
division always costs more than multiplication and both more than subtraction.

backend/services/deployments/api/http/routing.go
```diff
@@ -113,34 +116,37 @@ func NewRouter(
...
+				// 2XX
+				return true
+			}
+			return false
```
merlin-northern:
```suggestion
			} else {
				return false
			}
```
usually faster.

backend/services/deployments/api/http/routing.go
```diff
@@ -113,34 +116,37 @@ func NewRouter(
...
 
-	NewDeploymentsResourceRoutes(publicAPIs, deploymentsHandlers)
-	NewLimitsResourceRoutes(withAuth, deploymentsHandlers)
-	InternalRoutes(internalAPIs, deploymentsHandlers)
-	ReleasesRoutes(withAuth, deploymentsHandlers)
+	NewImagesResourceRoutes(routerAuth, deploymentsHandlers, cfg)
+	NewDeploymentsResourceRoutes(routerAuth, routerOpen, deploymentsHandlers)
+	NewLimitsResourceRoutes(routerAuth, deploymentsHandlers)
+	InternalRoutes(routerOpen, routerHealthz, deploymentsHandlers)
+	ReleasesRoutes(routerAuth, deploymentsHandlers)
```
merlin-northern: for some time I wanted to make an explicit distinction between the open source router and enterprise router. can we separate them? I am looking for you opinion on that.

backend/services/deployments/api/http/routing.go
```diff
@@ -155,37 +161,28 @@ func NewImagesResourceRoutes(router *gin.RouterGroup,
 	if controller == nil {
 		return
 	}
-	mgmtV1 := router.Group(ApiUrlManagement)
-	mgmtV1Artifacts := mgmtV1.Group(".")
-
-	artifactSizeLimit := requestsize.Middleware(cfg.MaxImageSize)
-	generateDataSizeLimit := requestsize.Middleware(cfg.MaxGenerateDataSize)
-
-	mgmtV1.Use(requestsize.Middleware(cfg.MaxRequestSize))
```
merlin-northern, line 166: where has the `MaxRequestSize` gone?

backend/services/deployments/api/http/routing.go
alfrunes: It compiles to the same thing. The suggestion has 1 more gocyclo keyword.

backend/services/deployments/api/http/routing.go
alfrunes: That sounds like a good idea. I'm noting it down for the PR to enterprise :memo:

backend/services/deployments/api/http/routing.go
alfrunes: Right! Indeed a 100 status code would never reach the access logs either! Let's drop even the subtraction and do `status < 300`.

backend/pkg/testing/rest/request.go
merlin-northern, line 1: this file lacks copyright header, why?

backend/services/deployments/api/http/api_deployments.go
merlin-northern, line 1: could we adjust slightly the copyright year?

merlin-northern (approved):
I really trust you you know

alfrunes:
Closing due to inactivity and there seems to be no decisive conclusion on this.

PR #806, opened 2025-07-17
fix(deviceconnect): Increase the file upload size limit

bahaa-ghazal:
Changelog: None

### DIFF
--- a/backend/services/deviceconnect/api/http/router.go
+++ b/backend/services/deviceconnect/api/http/router.go
@@ -61,6 +61,7 @@ const (
 type RouterConfig struct {
 	GracefulShutdownTimeout time.Duration
 	MaxRequestSize          int64
+	MaxFileSize             int64
 }
 
 // NewRouter returns the gin router
@@ -76,8 +77,11 @@ func NewRouter(
 			SetPathRegex(`^/api/(devices|management)/v[0-9]/`),
 	))
 
+	publicAPI := router.Group(".")
+	fileLimit := publicAPI.Group(".")
 	if config != nil {
-		router.Use(requestsize.Middleware(config.MaxRequestSize))
+		publicAPI.Use(requestsize.Middleware(config.MaxRequestSize))
+		fileLimit.Use(requestsize.Middleware(config.MaxFileSize))
 	}
 
 	gracefulShutdownTimeout := time.Duration(0)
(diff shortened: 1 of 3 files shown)


bahaa-ghazal:
I chose 1 GiB as limit, but I am not sure if we need to increase/decrease it

backend/services/deviceconnect/api/http/router.go
```diff
@@ -95,19 +99,19 @@ func NewRouter(
...
-	router.GET(APIURLManagementDeviceConnect, management.Connect)
-	router.GET(APIURLManagementDeviceDownload, management.DownloadFile)
-	router.HEAD(APIURLManagementDeviceDownload, management.DownloadFile)
-	router.POST(APIURLManagementDeviceCheckUpdate, management.CheckUpdate)
-	router.POST(APIURLManagementDeviceSendInventory, management.SendInventory)
-	router.PUT(APIURLManagementDeviceUpload, management.UploadFile)
-	router.GET(APIURLManagementPlayback, management.Playback)
+	publicAPI.GET(APIURLManagementDevice, management.GetDevice)
+	publicAPI.GET(APIURLManagementDeviceConnect, management.Connect)
+	publicAPI.GET(APIURLManagementDeviceDownload, management.DownloadFile)
```
merlin-northern, line 109: do you think we should also limit the download?

backend/services/deviceconnect/api/http/router.go
```diff
@@ -61,6 +61,7 @@ const (
 type RouterConfig struct {
 	GracefulShutdownTimeout time.Duration
 	MaxRequestSize          int64
+	MaxFileSize             int64
```
merlin-northern, line 64: could you add // comments here to explain what is the difference between these two limits? first look and one can think that PUT request is just a request so why do we need a separate limit?

merlin-northern (approved):

backend/services/deviceconnect/config/config.go
```diff
@@ -88,6 +88,10 @@ const (
 	// Max Request body size
 	SettingMaxRequestSize        = "request_size_limit"
 	SettingMaxRequestSizeDefault = 1024 * 1024 // 1 MiB
+
+	// Max Upload size
+	SettingMaxFileUploadSize        = "file_upload_limit"
+	SettingMaxFileUploadSizeDefault = 1024 * 1024 * 1024 // 1 GiB
```
merlin-northern, line 94: I have no idea if the customers were using larger files. it can be a breaking change. we should document it on docs.mender.io

merlin-northern:
@bahaa-ghazal do you think a change log is needed? it would be a nice idea to have it in the change log I would say, would you?

backend/services/deviceconnect/api/http/router.go
merlin-northern: btw there are also limits inside the mender-connect. (dont remember the defaults)

bahaa-ghazal:
> @bahaa-ghazal do you think a change log is needed? it would be a nice idea to have it in the change log I would say, would you?

Yeah sure. I felt that this PR was more of a quick revert than an actual fix, I wasn’t sure if it need a changelog entry. But yeah, it wouldn’t hurt to include it

backend/services/deviceconnect/config/config.go
bahaa-ghazal: yeah I'll wait for Alf's review too see what he think before merging these changes

backend/services/deviceconnect/api/http/router.go
bahaa-ghazal: Sure, that seems also like a good idea

backend/services/deviceconnect/config/config.go
alfrunes: Yeah, we should document these limits on mender docs before releasing.

alfrunes (approved):
LGTM. Thanks for fixing it so quickly! :rocket:

PR #810, opened 2025-07-18
fix(useradm): Always generate a unique user ID for newly created users

alfrunes:


kjaskiewiczz (changes requested):
`FAILED tests/test_cli.py::TestCli::test_create_user_with_id`
looks like we test this behavior
I'm wondering if there was a use-case for having possibility to create user with particular ID using useradm CLI? I don't think so.

alfrunes:
> `FAILED tests/test_cli.py::TestCli::test_create_user_with_id` looks like we test this behavior I'm wondering if there was a use-case for having possibility to create user with particular ID using useradm CLI? I don't think so.

I adjusted the PR to always generate a UUID on public APIs, leaving the internal and CLI commands as before.

kjaskiewiczz (approved):
looks good

PR #813, opened 2025-07-22
test(deployments): Integration test for v2 GET `/artifacts`

bahaa-ghazal:
Ticket: MEN-8182


backend/tests/integration/tests/test_deployments.py
```diff
@@ -2378,16 +2391,142 @@ def do_test_show_artifact_size(self, clean_mongo, user_token, devs):
...
+        assert len(artifacts) == 4
+        
+        # search for single exact name
+        r = api_dep_v2.with_auth(user_token).call(
+            "GET", deployments_v2.URL_ARTIFACTS +
+            "?name=test-artifact"
+        )
+        assert r.status_code == 200
+        artifacts = r.json()
+        assert len(artifacts) == 1
```
alfrunes: This assertion is quite weak. It's not checking that it's the expected artifact.

backend/tests/integration/tests/test_deployments.py
```diff
@@ -2378,16 +2391,142 @@ def do_test_show_artifact_size(self, clean_mongo, user_token, devs):
...
+    def do_list_artifacts_v2(self, clean_mongo, user_token):
+        api_dep_v2 = ApiClient(deployments_v2.URL_MGMT)
+        # we allready have the test-artifact.mender with name "deployments-phase-testing"
+        # and we add 3 more
+        create_test_artifacts(user_token, 2)
+        # create another artifact with the name "test-artifact" to test the exact search
+        with get_mender_artifact(
+            "test-artifact",
+        ) as artifact:
+            upload_image(artifact, user_token)
```
alfrunes: This could be a pretty nice [test fixture](https://docs.pytest.org/en/6.2.x/fixture.html#what-fixtures-are) for reuse in other tests as well.

alfrunes (dismissed):
Great coverage of different test scenarios. Would appreciate if you could make the assertions a little more rigid. Otherwise looks good.

backend/tests/integration/tests/test_deployments.py
bahaa-ghazal: I couldn't turn it into a fixture because it needed a user token 🥲

backend/tests/integration/tests/test_deployments.py
```diff
@@ -2378,16 +2413,157 @@ def do_test_show_artifact_size(self, clean_mongo, user_token, devs):
...
 @pytest.mark.storage_test
-class TestDeploymentShowArtifactSizeOpenSource(_TestDeploymentsShowArtifactSizeBase):
+class TestDeploymentArtifactOpenSource(_TestDeploymentsArtifactBase):
     def test_show_artifact_size(self, clean_mongo):
         _user, user_token, devs = setup_devices_and_management_st(10)
         self.do_test_show_artifact_size(clean_mongo, user_token, devs)
+    def test_list_artifacts_v2(self, clean_mongo):
+        _user, user_token, devs = setup_devices_and_management_st()
+        artifacts = create_test_artifacts(utoken=user_token)
+        artifacts.insert(0, "deployments-phase-testing")
```
alfrunes, line 2556: Not very obvious where this comes in, but figured out it comes as a side-effect of `setup_devices_and_management`...

backend/tests/integration/tests/test_deployments.py
```diff
@@ -2378,16 +2413,157 @@ def do_test_show_artifact_size(self, clean_mongo, user_token, devs):
...
+        # search for multiple exact names
+        r = api_dep_v2.with_auth(user_token).call(
+            "GET", deployments_v2.URL_ARTIFACTS +
+            "?name=test-artifact&name=test-artifact-0"
+        )
+        assert r.status_code == 200
+        artifacts = r.json()
+        assert compare_artifact_names(
+            artifacts, test_artifacts[1:3]
+        )
```
alfrunes, line 2464: This does obfuscate the expectations, but it's much better than what it was.

backend/tests/integration/tests/test_deployments.py
alfrunes: You could make the fixture request the fixture for setting up the user.
Anyway, treat as suggestion. This test file is also not really using fixtures for setup and teardown.

alfrunes (approved):

PR #816, opened 2025-07-22
Fixes and imporvements to a few integraiton tests

alfrunes:


kjaskiewiczz (approved):
Nice!
Nitpick - following https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#description
I wouldn't use capital letter in commit messages, so 
`refactor: move (...)` instead of `refactor: Move (...)`
`test: speed up` instead of `test: Speed up` etc.
and would probably use `speed up` everywhere (instead of `speedup`)
;)

PR #825, opened 2025-07-24
Fix deployments integration tests

alfrunes:


backend/tests/integration/tests/test_deployments.py
```diff
@@ -87,14 +98,33 @@ def create_test_artifacts(utoken: str,
...
+        .images.find(
+            {},
+            {
+                "meta_artifact.name": 1,
+                "meta_artifact.depends.device_type": 1,
+                "meta.description": 1,
+            },
+        )
+        .sort({"meta_artifact.name": 1})
+    ]
```
alfrunes, line 118: Not ideal, but I didn't want to use the deployments API as the source of truth since it is the system under test.

kjaskiewiczz (approved):

backend/tests/integration/tests/test_deployments.py
merlin-northern: yeah that is one of the points I wanted to bring up: if we tweak the database, it defies the purpose of a test. we can arrive at any state in that way.

merlin-northern (approved):

PR #827, opened 2025-07-25
QA-721: Change database cleanup strategy to preserve indexes

alfrunes:


backend/services/deployments/tests/common.py
```diff
@@ -376,21 +377,26 @@ def clean_minio(s3_bucket):
...
-        mongo.drop_database(d)
-    db = mongo["deployment_service"]
-    collections = db.list_collection_names()
-    for c in collections:
-        if c == "migration_info":
-            continue
-        db[c].delete_many({})
+def mongo_cleanup(mongo: MongoClient):
+    dbs = mongo.list_databases(
+        filter={"name": {"$nin": ["admin", "config", "local", "workflows"]}},
```
kjaskiewiczz, line 382: for my knowledge (I simply don't remember) - why do we skip `workflows`?

alfrunes: I believe it's because of the workflows being persisted in the database. If the collection is dropped it will forget about all the workflows definitions.

kjaskiewiczz (approved):
nice!

PR #832, opened 2025-07-28
test: add `None` arguments to `typing.Generator`

danielskinstad:
Python versions older than 3.13 raises:
`TypeError: Too few arguments for typing.Generator; actual 1, expected 3`

https://docs.python.org/3/library/typing.html#typing.Generator: `Changed in version 3.13: Default values for the send and return types were added.`

### DIFF
--- a/backend/tests/integration/testutils/common.py
+++ b/backend/tests/integration/testutils/common.py
@@ -647,7 +647,7 @@ class MockedHttp:
 def retry(
     period: timedelta = timedelta(milliseconds=100),
     timeout: timedelta = timedelta(minutes=2),
-) -> Generator[datetime]:
+) -> Generator[datetime, None, None]:
     if timeout < period:
         raise ValueError(f"retry period {period} is larger than timeout {timeout}")
 


backend/tests/integration/testutils/common.py
```diff
@@ -647,7 +647,7 @@ def request_get_all(self, path):
 def retry(
     period: timedelta = timedelta(milliseconds=100),
     timeout: timedelta = timedelta(minutes=2),
-) -> Generator[datetime]:
+) -> Generator[datetime, None, None]:
```
alfrunes, line 650: Shouldn't it be like this?
```suggestion
) -> Generator[datetime | None]:
```

alfrunes: Sorry, should've consulted the documentation first :see_no_evil:

alfrunes (approved):

backend/tests/integration/testutils/common.py
alfrunes: I probably should've just used:
```suggestion
) -> Iterator[datetime]:
```

PR #833, opened 2025-07-29
Update ratelimits configuration interface

alfrunes:
The behavior of the default rate limiting group can be confusing, so I thought it might be a good idea to remove it and replace it with an (almost) equivalent default config:
```yaml
# Replaces the `default` configuration block:
groups:
- name: default
  # ... (default paramters)
match:
- api_pattern: / # this pattern is a catch-all with lowest priority
  group_expression: "default"
```
The default is now to ignore ALL unmatched requests with a new option to reject them to make configuration more predictable.


kjaskiewiczz (commented):
Is this one ready for review? (it doesn't pass static checks)
Is this part of https://northerntech.atlassian.net/browse/MEN-7744? (sorry, I was off recently and lost track of what's going on with rate limits)

alfrunes:
> Is this one ready for review? (it doesn't pass static checks) Is this part of https://northerntech.atlassian.net/browse/MEN-7744? (sorry, I was off recently and lost track of what's going on with rate limits)

I fixed the static check errors.
About the rate limiting stuff, I have to admit this turned out to be a bit of a mess. In this PR I'm trying to simplify the configuration interface a little by removing the confusing "default" limits and instead define the limits using the api patterns instead. I also added the "reject_unmatched" option since the replacement of the default is to accept requests that don't match a ratelimiting group.

kjaskiewiczz (approved):
looks good

PR #839, opened 2025-07-31
Fix/unwrap session details

mzedel:
Stumbled across the missing unwrap in the test config in https://github.com/NorthernTechHQ/nt-gui/pull/259 ... as a different approach than adding the `unwrap()` as in https://github.com/mendersoftware/mender-server/pull/473 this is a bit more expansive in order to reduce the code duplication.

### DIFF
--- a/frontend/src/js/components/auditlogs/eventdetails/utils.tsx
+++ b/frontend/src/js/components/auditlogs/eventdetails/utils.tsx
@@ -23,6 +23,8 @@ import { getDeviceById, getSessionDetails } from '@northern.tech/store/thunks';
 import dayjs from 'dayjs';
 import duration from 'dayjs/plugin/duration';
 
+import DeviceDetails, { DetailInformation } from './DeviceDetails';
+
 dayjs.extend(duration);
 
 export interface SessionDetailsEventProps {
@@ -102,3 +104,15 @@ export const useSessionDetails = (auditLogItem: AuditLog): UseSessionDetailsRetu
     sessionMeta
   };
 };
+
+interface SessionInfoProps extends Omit<UseSessionDetailsReturn, 'sessionDetails' | 'isLoading'> {
+  onClose: () => void;
+  title: string;
+}
+
+export const SessionInfo = ({ device, idAttribute, sessionMeta, onClose, canReadDevices, title }: SessionInfoProps) => (
+  <div className="flexbox column margin-small" style={{ minWidth: 'min-content' }}>
+    {canReadDevices && <DeviceDetails device={device} idAttribute={idAttribute} onClose={onClose} />}
+    <DetailInformation title={title} details={sessionMeta} />
+  </div>
+);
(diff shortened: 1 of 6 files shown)


frontend/src/js/components/auditlogs/eventdetails/utils.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface UseSessionDetailsReturn {
+  canReadDevices: boolean;
+  device?: Device;
+  idAttribute: IdAttribute | string;
+  isLoading: boolean;
+  sessionDetails?: SessionDetails;
+  sessionMeta: SessionMeta | Record<string, never>;
+}
+
+export const useSessionDetails = (item: AuditLog): UseSessionDetailsReturn => {
```
mineralsfree: I see it's used everywhere, and with TypeScript it's not that bad, but to me, `item` is too generic. Maybe we could use `auditLogItem` or `auditLogEvent` instead?

frontend/src/js/components/auditlogs/eventdetails/utils.tsx
```diff
@@ -0,0 +1,100 @@
...
+        userId: actor.id,
+        startDate: action.startsWith('open') ? time : undefined,
+        endDate: action.startsWith('close') ? time : undefined
+      })
+    )
+      .unwrap()
+      .then(setSessionDetails);
+  }, [action, actor.id, canReadDevices, dispatch, meta.session_id, object.id, time]);
+
+  const isLoading = !sessionDetails || (canReadDevices && !device);
```
mineralsfree, line 86: It looks like `isLoading` will remain `true` if the request fails, which might be confusing for the user

mineralsfree (dismissed):
Few things to consider, looks good otherwise

frontend/src/js/components/auditlogs/eventdetails/utils.tsx
mzedel: Good point! I assume getting the duration isn't as crucial as showing the other info, so it falls back to use the one known time instead... good enough?

mineralsfree: looks good :)

frontend/src/js/components/auditlogs/eventdetails/utils.tsx
```diff
@@ -0,0 +1,118 @@
...
+        setSessionDetails({ end: time, start: time });
+      })
+      .then(setSessionDetails);
+  }, [action, actor.id, canReadDevices, dispatch, meta.session_id, object.id, time]);
+
+  const isLoading = !sessionDetails || (canReadDevices && !device);
+
+  const sessionMeta: SessionMeta | Record<string, never> = sessionDetails
+    ? {
+        'Session ID': item.meta.session_id[0],
```
mineralsfree: `item` should also be renamed to `auditLogItem`  here as well

mineralsfree (changes requested):
Not all of the `item` references where renamed

frontend/src/js/components/auditlogs/eventdetails/utils.tsx
mzedel: 🫣

mineralsfree (approved):

PR #842, opened 2025-08-01
docs(workflows): prepare API specification in OpenAPI format

kjaskiewiczz:
Changelog: Title
Ticket: QA-893

### DIFF
--- a/backend/docs/api/workflows.openapi.yaml
+++ b/backend/docs/api/workflows.openapi.yaml
@@ -1,0 +1,425 @@
+openapi: 3.0.1
+info:
+  title: Workflows
+  description: |
+    An API for workflow management.
+    Intended for use by other services.
+  version: "1"
+servers:
+- url: http://mender-workflows:8080/
+paths:
+  /status:
+    get:
+      summary: Check if service API is alive and serving requests
+      operationId: Check Liveliness
+      responses:
+        "200":
+          description: Successful response.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  id:
+                    type: string
+                  name:
+                    type: string
+              example:
+                status: ok
+        "500":
+          $ref: '#/components/responses/InternalServerError'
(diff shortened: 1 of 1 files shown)


backend/docs/api/workflows.openapi.yaml
```diff
@@ -0,0 +1,425 @@
...
+          type: string
+      responses:
+        "200":
+          description: Successful query
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/JobStatus'
+        "404":
+          $ref: '#/components/responses/NotFoundError'
```
alfrunes, line 157: Is this a duplicate endpoint of `/api/v1/jobs/{id}`?

backend/docs/api/workflows.openapi.yaml
```diff
@@ -0,0 +1,425 @@
...
+        inputParameters:
+          type: array
+          items:
+            $ref: '#/components/schemas/InputParameter'
+        status:
+          type: string
+        results:
+          type: array
+          items:
+            $ref: '#/components/schemas/TaskResult'
```
alfrunes, line 377: JobStatus is the same as JobObject minus insert_time and version :monocle_face:

alfrunes (approved):
API conversion looks good to me.
Just found a minor unrelated observation.

backend/docs/api/workflows.openapi.yaml
kjaskiewiczz: @merlin-northern maybe you know more about this one (I think you introduced `api/v1/jobs/{id}`)?

kjaskiewiczz:
@merlin-northern @alfrunes I don't want to change the API as part of this PR, but it looks like there is some duplication and we should take a look at some point, so I created follow-up QA ticket - https://northerntech.atlassian.net/browse/QA-1145 and I'm merging this PR without changes.

PR #854, opened 2025-08-06
MEN-8326: Migrate consumers from push to pull mode

bahaa-ghazal:


backend/services/workflows/client/nats/client.go
```diff
@@ -263,19 +276,53 @@ func (c *client) CreateConsumer(name string, upsert bool, config ConsumerConfig)
...
+				continue
+			}
+			return
+		}
+		if msgs == nil || len(msgs) != 1 {
+			return
+		}
+		q <- msgs[0]
+	}
+}
```
alfrunes: When this go routine terminates, the application is going to hang. Perhaps `defer close(q)`?

backend/services/workflows/client/nats/client.go
```diff
@@ -263,19 +276,53 @@ func (c *client) CreateConsumer(name string, upsert bool, config ConsumerConfig)
...
+		if err != nil {
+			if errors.Is(err, natsio.ErrTimeout) {
+				continue
+			}
+			return
+		}
+		if msgs == nil || len(msgs) != 1 {
+			return
+		}
+		q <- msgs[0]
```
alfrunes: There is a congestion issue here: if all workers are busy, this message handler could timeout prematurely or even be processed after the deadline which could lead to strange behavior. How about sending a nil Msg before calling fetch (needs appropriate handling in worker_group.go) to make sure there is a listener?

backend/services/workflows/client/nats/client.go
```diff
@@ -190,10 +189,24 @@ func (cfg ConsumerConfig) Validate() error {
...
+	deliverSubject string,
+	mode ConsumerMode) *natsio.ConsumerConfig {
+	switch mode {
+	case PushMode:
+		if deliverSubject == "" {
+			deliverSubject = natsio.NewInbox()
+		}
+	case PullMode:
+		deliverSubject = ""
+		cfg.MaxDeliver = 0
```
alfrunes:
```suggestion
		fallthrough
```

backend/services/workflows/client/nats/client.go
```diff
@@ -286,3 +333,36 @@ func (c *client) JetStreamPublish(subj string, data []byte) error {
...
+	return PushMode, nil
+}
+
+// DeleteModeConsumer deletes the consumer if its mode matches the given mode
+func (c *client) DeleteConsumerByMode(name string, mode ConsumerMode) error {
+	currentMode, err := c.getConsumerMode(name)
+	if errors.Is(err, natsio.ErrConsumerNotFound) {
+		return nil
+	} else {
+		if err != nil {
```
alfrunes:
```suggestion
	} else if err != nil {
```

backend/services/workflows/client/nats/client.go
```diff
@@ -16,26 +16,28 @@ package nats
...
 	"github.com/mendersoftware/mender-server/pkg/log"
 )
 
+type ConsumerMode int
+
 const (
+	PushMode = ConsumerMode(iota)
+	PullMode
+
+	defaultConsumerMode = PullMode
```
alfrunes: I don't understand the value of introducing this const which is used once. I believe it would be more clear to simply use `PullMode` wherever this is used to avoid the extra indirection.

backend/services/workflows/client/nats/client.go
bahaa-ghazal: it is closed here https://github.com/bahaa-ghazal/mender-server/blob/921223902f5225dcc22a583bca6cee180881a98d/backend/services/workflows/app/worker/worker.go#L116.
wouldn't that create a panic?

alfrunes: You're right. Maybe do you think we need a second channel to handle the synchronization part?

bahaa-ghazal: Uff idk 😔. Creating goroutine just to fetch and push a channel in this situation seems .. clumsy. I will try other simpler approach and see if it will work.

bahaa-ghazal: I ended up creating  second channel, hope it is good enough 🤞

backend/services/workflows/app/worker/worker_group.go
```diff
@@ -117,7 +120,93 @@ func (w *workerGroup) RunWorker(ctx context.Context) {
...
+
+		case <-ctxDone:
+			return
+		case <-sidecarDone:
+			return
+		}
+		// Release message
+		if err := msg.AckSync(); err != nil {
+			l.Error(errors.Wrap(err, "failed to ack the message"))
+		}
```
alfrunes: There's quite a bit of code duplication here. Maybe you could refactor and only wrap the first `select` branch in a push/pull `func`?

alfrunes (dismissed):
Looks good.
Would be great if you could remove the code duplication, but I do not consider it a blocker.

alfrunes (approved):
Thanks!

alfrunes:
The failing acceptance tests should be fixed by https://github.com/mendersoftware/mender-server/pull/865

alfrunes (approved):

PR #855, opened 2025-08-07
fix(deviceauth): token verify needs to use X-Forwarded-* for ratelimi…

alfrunes:
…ting

Changelog: None


merlin-northern (approved):
wow nice catch

PR #856, opened 2025-08-07
fix(workflows): move inventory to end of decommission workflow

alfrunes:
To lower the probability of failed decommission causing the device not to show up in the UI. By not removing the device from inventory it is possible to initiate the retry from the UI.

Ticket: None

### DIFF
--- a/backend/services/workflows/tests/docker-compose.yml
+++ b/backend/services/workflows/tests/docker-compose.yml
@@ -5,9 +5,6 @@ services:
     extends:
       file: ../../../tests/docker-compose.yml
       service: acceptance-tester
-    command: # FIXME(QA-733): Align dockerfile with open source
-      - -k
-      - not test_create_artifacts
     volumes:
       - ".:/testing"
     depends_on:
--- a/backend/services/workflows/tests/docker-compose.yml
+++ b/backend/services/workflows/tests/docker-compose.yml
@@ -78,7 +78,7 @@ services:
       HAVE_DEVICECONFIG: 1
       GOCOVERDIR: /cover
       DEPLOYMENTS_ADDR: "mender-deployments:8080"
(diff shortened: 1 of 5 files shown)


backend/services/workflows/tests/tests/test_decommission_device.py
merlin-northern, line 1: could we increment a couple of times the copyright year?

merlin-northern (dismissed):
I have not checked why tests fail

backend/services/workflows/tests/docker-compose.yml
```diff
@@ -78,7 +75,7 @@ services:
       HAVE_DEVICECONFIG: 1
       GOCOVERDIR: /cover
       DEPLOYMENTS_ADDR: "mender-deployments:8080"
-      DEVICE-AUTH_ADDR: "mender-device-auth:8080"
+      DEVICEAUTH_ADDR: "mender-device-auth:8080"
```
merlin-northern, line 78: that seriously scares me.

backend/services/workflows/tests/docker-compose.yml
```diff
@@ -5,9 +5,6 @@ services:
     extends:
       file: ../../../tests/docker-compose.yml
       service: acceptance-tester
-    command: # FIXME(QA-733): Align dockerfile with open source
-      - -k
-      - not test_create_artifacts
```
merlin-northern, line 10: I kind of think that this dance with the open source tests in the ent and vice-versa is pointless. os code, os tests, os repo. ent code, ent tests, ent repo. perhaps it would be easier. idle comment here totally.

merlin-northern (approved):

backend/services/workflows/tests/docker-compose.yml
alfrunes: Happy path, happy tests.

PR #859, opened 2025-08-07
RFC - build(gui): added HMR enabled dev option

mzedel:
with the larger number of incoming design adjustments I thought that might be useful


mineralsfree (approved):
:+1:

PR #863, opened 2025-08-12
test: Update cleanup strategy for acceptance tests

alfrunes:
I missed updating `deviceconfig` and `deviceauth` in #827

### DIFF
--- a/backend/services/deviceauth/tests/common.py
+++ b/backend/services/deviceauth/tests/common.py
@@ -34,6 +34,7 @@ from client import SimpleInternalClient, SimpleManagementClient, BaseDevicesApiC
 import mockserver
 import orchestrator
 import os
+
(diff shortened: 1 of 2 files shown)


merlin-northern (approved):
I totally trust you. cards on  the table: I think each test should start with a dedicated db, migrate it to given version, populate with given data, and run. right now we share a db with all the tests znd noone knows what is where and what version. some tests expect migration some dont, some migrate, some dont. I do not like it. (also an idle comment not related to the changes in this PR)

alfrunes:
> I totally trust you. cards on the table: I think each test should start with a dedicated db, migrate it to given version, populate with given data, and run. right now we share a db with all the tests znd noone knows what is where and what version. some tests expect migration some dont, some migrate, some dont. I do not like it. (also an idle comment not related to the changes in this PR)

I think that's what we do (or at least was the intention) in the full integration test suite. However, it comes with some performance implications since we need to tear down and re-initialize the database for every test. I believe there's certainly ways we can optimize this (both in terms of writing broader tests and reducing startup costs) but I believe that would result in yet another test suite. Maybe the plan to move tests to k8s would be a golden opportunity to create a new test suite which isolates the database instances as you suggest?

PR #866, opened 2025-08-13
chore(gui): redirect to the old upgrade page as new subscription endp…

mineralsfree:
…oint is disabled

this to be reverted once the new subscription flow is possible

Ticket: None


mzedel (dismissed):
For the pipeline to pass we will have to revert the staging tests too 😞

mzedel (dismissed):

mzedel (approved):

PR #868, opened 2025-08-14
MEN-7812: Validate scope in inventory search

frodeha:
**Description**

This change introduces input validation of `scope` properties used with the inventory filter search endpoint. The `scope`  the `filter`, `sort` and `attributes` property of the input search query now has to be one of: 
```
"system"
"identity"
"inventory"
"monitor"
"tags"
```
 Any other value will result in a `400 - Bad Request` result  from the search endpoint.

Introducing this validation also fixes the problem described in the bug report where a malformed `scope` property would produce a `500 - Internal Server Error` when `mongodb` attempted to execute the search query. The reported issue was caused by the scope property containing multiple (2 or more) period signs in direct sequence (ex: `"something..else"` or `"../../something"`) which is incompatible with `mongodb` projection and sort operations. 

Introducing input validation addresses both the original issue and clearly communicates to users of the endpoint what scopes are actually supported by the system and this is therefore (imo) a good solution to the original issue.

**Other notes / questions**

* I'm not exactly 100% sure the list of accepted values for `scope` in this pull request is the complete set of values that should be allowed. I have searched around and these are the ones I could find in use by the UI and Client (afaict), but it's absolutely possible that I have missed some, so please do let me know if that's the case.
* I have taken the liberty of restructuring the inventory integration tests as part of this PR. This is to run search tests on open source _and_ enterprise (not only enterprise as it was before) and was a pre-requisite for me to add tests to cover this change in open source. I have tried to follow established patterns, but my `python` and `pytest` fu is somewhat weak so please do let me know if I could have done things better here.
* I have had a look at the commit structure and conventions used by others in this repository and tried to follow suit. If I've made mistakes or otherwise done things in ways that are not ok - do let me know!

### DIFF
--- a/backend/docs/api/inventory_internal_v2.yaml
+++ b/backend/docs/api/inventory_internal_v2.yaml
@@ -127,11 +127,7 @@ components:
           description: |
             A human readable, unique attribute ID, e.g. 'device_type', 'ip_addr', 'cpu_load', etc.
         scope:
-          type: string
-          description: |
-            The scope of the attribute.
-
-            Scope is a string and acts as namespace for the attribute name.
+          $ref: '#/components/schemas/Scope'
         description:
           type: string
           description: Attribute description.
@@ -198,11 +194,7 @@ components:
       type: object
       properties:
         scope:
-          type: string
-          description: |
-            The scope of the attribute.
-
-            Scope is a string and acts as namespace for the attribute name.
+          $ref: '#/components/schemas/Scope'
         attribute:
           type: string
           description: |
@@ -248,8 +240,7 @@ components:
           type: string
           description: Attribute name.
         scope:
-          type: string
-          description: Attribute scope.
+          $ref: '#/components/schemas/Scope'
       description: Inventory attribute
       example:
         attribute: serial_no
@@ -262,11 +253,7 @@ components:
       type: object
       properties:
         scope:
-          type: string
-          description: |
-            The scope of the attribute.
-
-            Scope is a string and acts as namespace for the attribute name.
+          $ref: '#/components/schemas/Scope'
         attribute:
           type: string
           description: |
@@ -285,3 +272,21 @@ components:
         attribute: serial_no
         scope: inventory
         order: asc
+    Scope:
+      type: string
+      description: |
+        The scope of the attribute.
+
+        Scope is a string and acts as namespace for the attribute name.
+
+        * __inventory__: Attributes reported by the device.
+        * __system__: Attributes populated by the mender-server.
+        * __identity__: Device's identity attributes provided in the device's auth request.
+        * __monitor__: Attributes populated by the monitoring add-on.
+        * __tags__: User-defined attributes associated with the device.
+      enum:
+        - system
+        - identity
+        - inventory
+        - monitor
+        - tags
--- a/backend/docs/api/inventory_management_v2.yaml
+++ b/backend/docs/api/inventory_management_v2.yaml
@@ -159,11 +159,7 @@ components:
           description: |
             A human readable, unique attribute ID, e.g. 'device_type', 'ip_addr', 'cpu_load', etc.
         scope:
-          type: string
-          description: |
-            The scope of the attribute.
-
-            Scope is a string and acts as namespace for the attribute name.
+          $ref: '#/components/schemas/Scope'
         description:
           type: string
           description: Attribute description.
@@ -232,8 +228,7 @@ components:
           type: string
           description: Name of the attribute.
         scope:
-          type: string
-          description: Scope of the attribute.
+          $ref: '#/components/schemas/Scope'
         count:
           type: integer
           description: Number of occurrences of the attribute in the database.
@@ -253,7 +248,7 @@ components:
           type: string
           description: Attribute name.
         scope:
-          type: string
+          $ref: '#/components/schemas/Scope'
         type:
           type: string
           description: Type or operator of the filter predicate.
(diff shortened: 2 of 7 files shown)


backend/tests/integration/tests/test_inventory.py
```diff
@@ -312,11 +312,7 @@ def add_devices_to_tenant(tenant, dev_inventories):
 
     return tenant
 
-
-@pytest.mark.skipif(
-    useExistingTenant(), reason="not feasible to test with existing tenant",
-)
-class TestDeviceFilteringEnterprise:
+class TestDeviceFilteringBase:
```
alfrunes: Since pytest will collect all classes with `Test` prefix, it might be a good idea to use a different class name?
```suggestion
class DeviceFilteringTests:
```

backend/tests/integration/tests/test_inventory.py
```diff
@@ -325,6 +321,275 @@ def logger(self):
             self._logger = logging.getLogger(self.__class__.__name__)
         return self._logger
 
+    def do_test_search_v2(self, user_token, devices, additional_test_cases = []):
+        assert user_token
+        assert len(devices) > 0
+
+        test_cases = [
```
alfrunes, line 328: Would be great if we could use `pytest.mark.parameterize` to pass these test cases so it doesn't fail early.

backend/docs/api/inventory_internal_v2.yaml
```diff
@@ -285,3 +272,15 @@ components:
...
+      description: |
+        The scope of the attribute.
+
+        Scope is a string and acts as namespace for the attribute name.
+      enum:
+        - system
+        - identity
+        - inventory
+        - monitor
+        - tags
```
alfrunes: Missing a newline at the end of the file here.

backend/tests/integration/tests/test_inventory.py
```diff
@@ -325,6 +321,275 @@ def logger(self):
...
+    def test_search_v2_internal(self):
+        """
+        This endpoint technically exists in open-source, but
+        the URL expects a "tenant_id" parameter in the
+        URL ("tenants" is a concept that doesn't exist in open-source)
+        and when one isn't specified, the implementation is
+        identical to search_v2 above afaict.
+
+        As such, I see no need to maintain tests for this endpoint.
+        """
```
alfrunes, line 565: :+1:

backend/docs/api/inventory_management_v2.yaml
```diff
@@ -313,6 +306,18 @@ components:
...
+      description: |
+        The scope of the attribute.
+
+        Scope is a string and acts as namespace for the attribute name.
+      enum:
+        - system
+        - identity
+        - inventory
+        - monitor
+        - tags
```
alfrunes: Would probably be a good idea to document the scope values too.

backend/docs/api/inventory_management_v2.yaml
```diff
@@ -313,6 +306,18 @@ components:
         attribute: serial_no
         scope: inventory
         order: asc
+    Scope:
+      type: string
+      description: |
+        The scope of the attribute.
+
+        Scope is a string and acts as namespace for the attribute name.
```
alfrunes, line 314: Maybe?
```suggestion
        The scope of the attribute.

        Scope is a string and acts as namespace for the attribute name.
        
        * __inventory__: Attributes reported by the device.
        * __system__: Attributes populated by the mender-server.
        * __identity__: Device's identity attributes provided in the deivce's auth request.
        * __monitor__: Attributes populated by the monitoring add-on.
        * __tags__: User-defined attributes associated with the device.
```

alfrunes (commented):
:+1: to enumerating the scope values.
Thanks for extending the integration tests to cover open source! I added a few comments / suggestions.

backend/tests/integration/tests/test_inventory.py
frodeha: I'm not really sure how to do that. The test cases are using values from devices to assert the correct response. Devices are created differently for enterprise and open-source and doesn't exist until `setup_tenants` (enterprise) or `setup_user_and_devices` (open-source) have run. 

I'm quite fresh to pytest so I'm sure I'm missing something. Any pointers?

backend/tests/integration/tests/test_inventory.py
frodeha: Yep, that's a good idea indeed! 🤦 Nice catch - fixed in 4f33330620b8c3896b1a5484a01257d3c248082e

backend/docs/api/inventory_internal_v2.yaml
frodeha: Fixed in c186ad321a0d391579408ae75ebc247547ccb349

backend/docs/api/inventory_management_v2.yaml
frodeha: Yeah, good shout! Added in c186ad321a0d391579408ae75ebc247547ccb349 👍

backend/tests/integration/tests/test_inventory.py
alfrunes: Ahh, I didn't see the test cases are depending on the `devices` fixture. Seems like we're dealing with a :chicken:  and :egg: situation. You could try refactoring the response to use a callback / lambda function to generate the expected response from the fixture output.
If you don't want to enter the refactoring rabbit hole that's totally fine too :rabbit2:

alfrunes (approved):
Thanks for looking into my suggestions.

I don't think the test case refactoring important, so I'll leave it up to you if you wanna change it.

PR #869, opened 2025-08-14
ci: coveralls publish jobs allowed to fail

oldgiova:
Lately we're experiencing some issues with upstream coveralls API, which seems unreliable and are answering 5xx errors. This commit is to make the reporting jobs allowed to fail, instead of blocking pipelines

Ticket: QA-925


oldgiova:
Too many autocorrect from my lazyvim...

PR #872, opened 2025-08-15
MEN-8620 - no more demo + only trial matters for demo intervals

mzedel:


frontend/src/js/utils/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+  ipAddress?: string;
+  isHosted?: boolean;
+  isPreRelease?: boolean;
+  isTrial?: boolean;
+  tenantToken?: string;
+  token: string;
+}
+
+const getInstallScriptArgs = ({ isHosted, isPreRelease, hasMonitor }: Partial<DebConfigurationProps>) => {
+  let installScriptArgs = '--demo';
```
mineralsfree: Maybe we could have an array `arr = ['--demo']` and then 
`if (isPreRelease) arr.push('-c experimental')`
and so on 
and in the end `return arr.join(' ')`;

I find it a bit more clean and typo resistant + avoid reassignment

mineralsfree (dismissed):
Some small improvement suggested, looks good.
I see the storehook test is failing once again, lmk if you want me to help to fix it

mzedel:
meh, storehooks failed for me locally without the commit I had in here before, I'll try to get the store package PR ready to isolate this to nt-gui and then we can tackle this there 👌

frontend/src/js/utils/helpers.ts
mzedel: hmm... the resulting code looks a lot more jarring to me, mostly due to the irregular alignment 🙈... would it be tolerable to at least keep the ternaries in the getSetupArgs array pushes?:

```typescript
const getInstallScriptArgs = ({ isHosted, isPreRelease, hasMonitor }: Partial<DebConfigurationProps>) => {
  const installScriptArgs = ['--demo'];
  if (isPreRelease) installScriptArgs.push('-c experimental');
  if (isHosted && hasMonitor) installScriptArgs.push('--commercial');
  if (isHosted) installScriptArgs.push('--jwt-token $JWT_TOKEN');
  return installScriptArgs.join(' ');
};

const getSetupArgs = ({ deviceType = 'generic-armv6', ipAddress, tenantToken, isTrial }: Partial<DebConfigurationProps>) => {
  const menderSetupArgs = ['--quiet', `--device-type "${deviceType}"`];
  if (tenantToken) menderSetupArgs.push('--tenant-token $TENANT_TOKEN');
  // in production we use polling intervals from the client examples: https://github.com/mendersoftware/mender/blob/master/examples/mender.conf.production
  menderSetupArgs.push(isTrial ? '--demo' : '--retry-poll 300 --update-poll 1800 --inventory-poll 28800');
  // we still need to forward the ipAddress when showing the snippets for a gateway setup
  menderSetupArgs.push(ipAddress ? `--server-ip ${ipAddress}` : `--server-url https://${window.location.hostname} --server-cert=""`);
  return menderSetupArgs.join(' ');
};
```

mineralsfree: Looks like you're leaning toward brevity in most cases, which makes sense, but doesn't improve readability in my opinion. Personally, I don’t see any issue with using `if` and `else` in the second case. But it's not worth spending time discussing it here, IMO, so it's definitely tolerable :)

mzedel: Yeah we can have a chat about each others reasoning on this once you're in the office again 👍

mineralsfree (approved):

PR #874, opened 2025-08-15
fix(deployments): preset number of pending devices when creating deployment

kjaskiewiczz:
Ticket: MEN-8522

### DIFF
--- a/backend/services/deployments/app/app.go
+++ b/backend/services/deployments/app/app.go
@@ -1165,6 +1165,7 @@ func (d *Deployments) CreateDeviceConfigurationDeployment(
 
 	deployment.DeviceList = []string{deviceID}
 	deployment.MaxDevices = 1
+	deployment.Stats[model.DeviceDeploymentStatusPendingStr] = deployment.MaxDevices
 	deployment.Configuration = []byte(constructor.Configuration)
 	deployment.Type = model.DeploymentTypeConfiguration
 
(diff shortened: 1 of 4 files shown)


backend/services/deployments/app/app.go
```diff
@@ -1482,30 +1484,6 @@ func (d *Deployments) createDeviceDeploymentWithStatus(
...
-				ctx, deployment.Id,
-				newStatus, time.Now(),
-			)
-			if err != nil {
-				return nil, errors.Wrap(err,
-					"failed to update deployment status")
-			}
-		}
-	}
-
```
alfrunes, line 1508: I'm not sure if I understand how removing this is related to the ticket. Could you explain why this is not needed?

kjaskiewiczz: Since we are setting stats for pending devices when creating deployment (with this PR), we shouldn't update stats when creating device deployment. There is also no point in updating deployment status, since new device deployments will not trigger status change.

alfrunes: Right! I was a bit confused since the status was taken as an argument. Thought there were cases where we created device deployments with other statuses than pending.

alfrunes (dismissed):

bahaa-ghazal:
Because of the workaround we have, the UI shows double the amount of pending devices. There exists PR that fix it and should be merged simultaneously: https://github.com/mendersoftware/mender-server/pull/787. 
Also when it comes to dynamic deployments the number could be deceiving, where if max_devices is set it could show more devices that are pending than actual devices in the groupe.
Also if the dynamic deployment has no max_devices (which becomes 0), I'm not 100% sure, but the pending devices could go into the negative if device changes a state. There is no dynamic groups in OS so i couldn't test it, but something came to me.

kjaskiewiczz:
@bahaa-ghazal thanks for the comments!
I pulled UI change here (not sure if this will work though).
> Also when it comes to dynamic deployments the number could be deceiving, where if max_devices is set it could show more devices that are pending than actual devices in the group.

That's possible, but it's kind of best we can do, and it was discussed with PO - https://northerntech.atlassian.net/browse/MEN-8522?focusedCommentId=124940

> Also if the dynamic deployment has no max_devices (which becomes 0), I'm not 100% sure, but the pending devices could go into the negative if device changes a state.

Enterprise implementation will be different and will handle this case.

kjaskiewiczz:
@mineralsfree can you take a look at the pipeline failure here?

bahaa-ghazal (approved):

PR #880, opened 2025-08-20
MEN-8657 - log analysis 

mzedel:


frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
```diff
@@ -0,0 +1,204 @@
...
+    backgroundColor: alpha(theme.palette.secondary.light, 0.08),
+    border: `1px solid ${theme.palette.secondary.light}`,
+    borderRadius: theme.shape.borderRadius,
+    padding: theme.spacing(2)
+  }
+}));
+
+const markDownStyleOverrides: MarkdownToJSX.Overrides = {
+  ...defaultOverrides,
+  h1: { component: 'b' },
```
mineralsfree, line 55: Don't you think that might worsen the formatting? Maybe we could at least adjust the font size?

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
```diff
@@ -0,0 +1,204 @@
...
+    setAnalysisError(null);
+    try {
+      const result = await dispatch(generateDeploymentLogAnalysis({ deploymentId: deployment.id, deviceId })).unwrap();
+      setAnalysisResult(result);
+      setIsAnalyzing(false);
+    } catch (error) {
+      console.error('Error generating analysis:', error);
+      setIsAnalyzing(false);
+      if (error.status === 429) {
+        setAnalysisError('You have reached your limit of 50 AI requests per day. Please try again in about 12 hours.');
```
mineralsfree: This will stay `in about 12 hours` forever :), maybe "later" will do the job unless we have a proper cool-down info? is 50 requests a  tenant or organization-wide limit by the way?

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
```diff
@@ -0,0 +1,204 @@
...
+  };
+
+  const onCopyAnalysisClick = () => copy(analysisResult);
+
+  return (
+    <div className="padding-top-small padding-bottom-small">
+      <Header />
+      <div className="flexbox center-aligned">
+        <Button className="margin-right-small" color="secondary" disabled={!isAiEnabled || isAnalyzing} onClick={onGenerateAnalysisClick} variant="contained">
+          {isAnalyzing ? 'Generating summary...' : 'Generate summary'}
```
mineralsfree, line 164: maybe `Summarizing` ? or `Generating a summary...`? Couldn't find this state in the design, so you can ignore this suggestion

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
```diff
@@ -0,0 +1,204 @@
...
+export const AiLogAnalysis = ({ deployment, deviceId }: AiLogAnalysisProps) => {
+  const { classes } = useStyles();
+  const dispatch = useAppDispatch();
+  const [isAnalyzing, setIsAnalyzing] = useState(false);
+  const [analysisResult, setAnalysisResult] = useState('');
+  const [analysisError, setAnalysisError] = useState<string | null>(null);
+  const { aiFeatures = {} } = useSelector(getGlobalSettings);
+  const { isAdmin } = useSelector(getUserRoles);
+
+  const { enabled: isAiEnabled = true } = aiFeatures;
```
mineralsfree: Should it be enabled if not set ?

frontend/src/js/components/deployments/Report.tsx
```diff
@@ -106,6 +110,7 @@ export const DeploymentReport = ({ abort, onClose, past, retry, type, open }) =>
   const { devices = {}, device_count = 0, totalDeviceCount: totalDevices, statistics = {}, type: deploymentType } = deployment;
   const { status: stats = {} } = statistics;
   const totalDeviceCount = totalDevices ?? device_count;
+  const canAi = isHosted && (isPreview || hasAiEnabled);
```
mineralsfree, line 113: maybe it's just me, but it's not obvious from the first glance that `canAi` relates to the server that `can` ai. As we have both org settings and server settings, I think it would be nice to have a clearl separation in variable names? WDYT?

mineralsfree (dismissed):
Few non-blocking concerns, didn't test it though, just stared at the code :+1:

frontend/src/js/components/deployments/Report.tsx
mzedel: Agreed, I haven't found a good split there... we have some features that are tied to both (e.g. auditlogs) - so the user the component is rendered for `has` auditlogs, but he might not be able to use it so he `can`'t auditlog 😬... to make it worse - here the ai has to be enabled on the server, but also in the settings - but the server enabling is enough to show parts of it... maybe... `has<Functionality>Features` for anything coming from the feature flags/ server? and then we stick to `can...`/ `has...` for the user/tenant capabilties?

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
mzedel: 🤦 nope, that shouldn't be there...

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
mzedel: there was a bit of movement around the button & the loading indication... now it's a combination of https://www.figma.com/design/89Xu1zFUuinM8EOYo7A2i4/Deployment-log-AI-summary?node-id=5651-20697&t=7MeOAbDt2MrSOnQ0-4 (that title) + https://www.figma.com/design/89Xu1zFUuinM8EOYo7A2i4?node-id=5762-26568#1390213772 (where there's no button adjustment) - 🙈

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
mzedel: yeah, we don't get any info there + IMO the longer timespan lends it a bit poor to "later" (which I would associate more with a short term)... Maybe head straight to _tomorrow_?
as to who the limit is for - I think it's organization wide, but it's also not fully decided ATM...

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
mzedel: the markdown library will translate the markdown headings to `<Typography variant=... />`, which is a bit too free form here:
Screenshot 2025-08-22 at 15 32 13
vs. with the overrides:
Screenshot 2025-08-22 at 15 31 44

If we see value in the feature I think we can handle the received markdown a bit more extensively, here I thought it would be ok for the "experimental" state of the feature...

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
alfrunes: Could you use the `Retry-After` header? It gives the number of seconds until the ratelimits reset.

mzedel: Nice, I wasn't aware we had this!

frontend/src/js/components/deployments/Report.tsx
mzedel: I created https://northerntech.atlassian.net/browse/MEN-8715 - feel free to modify this with any additional info you would think is helpful!

alfrunes (approved):

PR #881, opened 2025-08-21
fix(deviceconnect): Make sure goroutines recover when panicing

bahaa-ghazal:
Ticket: MEN-7750

### DIFF
--- a/backend/pkg/accesslog/middleware.go
+++ b/backend/pkg/accesslog/middleware.go
@@ -14,14 +14,10 @@
 package accesslog
 
 import (
-	"fmt"
 	"net"
 	"net/http"
 	"os"
-	"path"
-	"runtime"
 	"strconv"
-	"strings"
 
 	"github.com/mendersoftware/mender-server/pkg/netutils"
 )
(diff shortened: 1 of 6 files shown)


backend/pkg/log/log.go
```diff
@@ -289,3 +291,46 @@ func FromContext(ctx context.Context) *Logger {
...
+		err = ErrPanic
+	} else {
+		err = errors.Wrap(opt.err, ErrPanic.Error())
+	}
+
+	if opt != nil && opt.channel != nil {
+		opt.channel <- opt.err
+	}
+
+	l.WithField("panic", r).Error(err)
```
alfrunes: Arguably the most important information after a panic is the stack trace. Could you add another field `trace` with the stack trace starting from the location of the panic?
Hint: checkout the access log middleware ;)

backend/pkg/log/log.go
```diff
@@ -289,3 +291,46 @@ func FromContext(ctx context.Context) *Logger {
...
+
+	if opt == nil || opt.err == nil {
+		err = ErrPanic
+	} else {
+		err = errors.Wrap(opt.err, ErrPanic.Error())
+	}
+
+	if opt != nil && opt.channel != nil {
+		opt.channel <- opt.err
+	}
```
alfrunes: I'm wondering if it's safer to close the channel. You could end up with potential deadlocks here if it's unbuffered or full and there's no receiver.

bahaa-ghazal: Agree, but this could also cause a panic when 2 go routines try to close the same channel. I'll try to make sure this problem is not met

bahaa-ghazal: Ended up going with sending an err to channel. Closing the channel didn't seem like a reliable solution given that other go routines might still use it or close it again.

alfrunes: ACK. In general it's preferable to close a channel once it's no longer used to prevent deadlocks. However, since I'm quite certain we use buffered channels with a single consumer I think it's ok to just send the error.

alfrunes (approved):
Would be nice if you could write a test that covers the SimpleRecovery part. Otherwise LGTM.

PR #884, opened 2025-08-25
Test/e2e convenience

mzedel:
pipeline including passing staging tests: https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/2020512434


alfrunes:
@mzedel I assume this one's not ready for review based on the last commit message? :monocle_face:

frontend/tests/e2e_tests/run.js
alfrunes, line 1: Moving from bash to js makes it a little harder for me to review, so take my review with a drop of :sake:
It's also a good reminder for me to up my TypeScript'fu :muscle: :lotus_position:
However, with the amount of emojis in this scripts makes it seriously hard to resist. :rocket: :new_moon_with_face: :star2:

frontend/pipeline.yml
```diff
@@ -169,7 +169,7 @@ build:frontend:docker:
 
 .template:test:frontend:acceptance:
   stage: test
-  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-cli
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/mendersoftware/mender-test-containers:gui-e2e-testing
```
alfrunes, line 172: Just a word of warning, these images are not versioned which might lead to unexpected breaking changes and increased maintenance burden for supported branches.
It's great for reducing boilerplate in development branches though :balance_scale:

alfrunes (approved):
:+1: Less lines in CI
:pencil2: Consistent use of node tools for invoking test environments.
:rocket: More emojis
:heavy_check_mark: Looks good to me.

frontend/tests/e2e_tests/run.js
```diff
@@ -0,0 +1,532 @@
...
+  const { name, password, user } = credentials;
+  const tenantResult = await composeExec(
+    'tenantadm',
+    `tenantadm create-org --name=${name} --username=${user} --password=${password} ${addons.map(addon => `--addon=${addon}`).join(' ')} ${options}`,
+    config
+  );
+  return tenantResult.out.trim();
+};
+
+const setupTenantToken = async (tenantId, config) =>
```
merlin-northern, line 273: a tenant should be an Object with certain methods, defined separately. perhaps one day we will make more moves towards nodejs. treat as suggestion as per general design of the present approach.

frontend/pipeline.yml
mzedel: Indeed... considering the access to the Gitlab image registry we nowadays use, I think there is an opportunity to step away from the test containers repo and either keep or just version these alongside the server version... adding it for the next QA meeting 👌

frontend/tests/e2e_tests/run.js
mzedel: We have a tenant definition [here](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/types/src/TenantV1.ts) (u know, generated from the API specs 😉), that we make use of in the gui codebase (which however is not really object oriented, but functional) - we could indeed apply this here too if the jump to TypeScript would be a little bit smaller 😬... that would then like look like:
```typescript
const tenant: TenantV1 = JSON.parse(tenantTokenResult.out)
const { tenant_token } = tenant;
```
I'll keep that in mind for the future ✍️ 🗒️

PR #886, opened 2025-08-25
ci: deploy branches on staging cluster

dhaustein:
This introduces GitOps flow for deploying development branches to our staging cluster.  

The CI code has been updated to automate the process of deploying branches through committing HelmRelease, Charts, and values.yaml files to the `mender-helm` repository, so that FluxCD can notice these and create a per-branch environment on the cluster.

**Key changes:**
* added a new `mender-helm-version-bump:branch` job that uses (now very very) changed `.helm-version-bump:` template
  * for dev branches, this will commit Charts and values to a new sub-directory `<branch-name>` in `mender-helm` repo under a new `dev/` parent directory
  * also adds a `HelmRelease` manifest to Flux, so that Flux starts watching the new stuff in the `dev/<branch-name>` sub-directory
* existing staging deployments from `main` I tried not to touch too much

**Example flow:**

1. developer creates a new branch, for example `feature/new-api`, commits changes, and pushes to remote
2. the push to the `feature/new-api` branch triggers a new CI pipeline
3. job `build:backend:docker` builds and publishes container images, saves their `.digest`
4. `mender-helm-version-bump:branch` job runs:
    (triggered for any non-main branch)
    Clones the `mendersoftware/mender-helm` repository
    Creates a new branch named `dev/feature-new-api` if one does not already exist
    Copies the production Helm chart to a new sub-directory named `dev/feature-new-api`
    Updates the `values.yaml` file in the new directory with the image tags of the container images that were just built
    Updates the `Chart.yaml` file with a new pre-release version, for example `4.0.0-12345` where `12345` is the pipeline ID
    Creates a new `HelmRelease` manifest file for the new environment
    Commits and pushes all these changes to the `dev/feature-new-api` branch in the `mender-helm` repo
5. FluxCD reconciles the changes and attempts to deploy the new version in a dedicated namespace on the staging cluster

**TODO:**
* Cleanup - There should be an automated way for a dev branch environment to be cleaned up after it's branch in mender-server is deleted, maybe some combination of Gitlab's Environments and .post jobs that delete the branch in `mender-helm` so that FluxCD removes the environment automatically when it notices the dev branch is missing
  * I threw something together for this here https://gist.github.com/dhaustein/6504b94b542926ec87005f7ae2ecb448
* Everything else - I commented out most of the other steps like linting, tests, frontend etc. because I want to focus on the deployment first, but we need those to work with the dev environments too ofc

Ticket: QA-1109
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -649,9 +649,10 @@ publish:backend:docker:
     name: quay.io/skopeo/stable:${SKOPEO_VERSION}
     # https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#override-the-entrypoint-of-an-image
     entrypoint: [""]
-  rules:
-    - if: '$CI_COMMIT_REF_PROTECTED == "true"'
-      when: on_success
+  rules:  # This job will run for any branch or tag, but not for maintenance branches
+    - if: '$CI_COMMIT_REF_NAME =~ /^\d+\.\d+\.x$/'
+      when: never
+    - if: $CI_COMMIT_TAG || $CI_COMMIT_BRANCH
   before_script:
     - skopeo login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
     - skopeo login --username $DOCKER_HUB_USERNAME --password $DOCKER_HUB_PASSWORD docker.io
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,7 +93,7 @@ variables:
     value: "302.Release-information/03.Open-source-licenses/01.Mender-Server/docs.md"
 
 include:
-  - component: "gitlab.com/Northern.tech/Mender/mendertesting/commit-lint@master"
+  # - component: "gitlab.com/Northern.tech/Mender/mendertesting/commit-lint@master"
   - project: "Northern.tech/Mender/mendertesting"
     file:
       - ".gitlab-ci-github-status-updates.yml"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -107,7 +107,7 @@ include:
       - when: never
 
 stages:
-  - lint
+  # - lint
   - build
   - test
   - publish
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -658,13 +658,14 @@ publish:backend:docker:
       if [ "$CI_COMMIT_REF_PROTECTED" == "true" ]; then
         export MENDER_PUBLISH_REGISTRY="docker.io"
         echo "INFO: Protected branch. Publishing to public registry: ${MENDER_PUBLISH_REGISTRY}"
+        skopeo login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
+        skopeo login --username $DOCKER_HUB_USERNAME --password $DOCKER_HUB_PASSWORD docker.io
       else
         export MENDER_PUBLISH_REGISTRY="${CI_REGISTRY}"
         export MENDER_PUBLISH_REPOSITORY="${MENDER_IMAGE_REPOSITORY}"
         echo "INFO: Unprotected branch. Publishing to internal registry: ${MENDER_PUBLISH_REGISTRY}/${MENDER_PUBLISH_REPOSITORY}"
+        skopeo login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
       fi
-    - skopeo login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -435,66 +212,483 @@ test:prep:
...
+.helm-version-bump:
+  needs:
+    - job: publish:backend:docker
+      artifacts: true
+    # - job: publish:frontend:docker
+    #   artifacts: true
+  rules: # This job will run for any branch or tag, but not for maintenance branches.
+    - if: '$CI_COMMIT_REF_NAME =~ /^\d+\.\d+\.x$/'
+      when: never
+    - if: $CI_COMMIT_TAG || $CI_COMMIT_BRANCH
```
dhaustein, line 227: Note: I switched this to run on anything but main, but I keep a mental note that yes, we will want staging environments from main too

.gitlab-ci.yml
```diff
@@ -435,66 +212,483 @@ test:prep:
...
-      when: delayed
-      start_in: 3 minutes
+    - git clone https://${GITHUB_USER_NAME}:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/${GITHUB_HELM_REPO} /tmp/helm
+    - cd /tmp/helm
+    - | # Set the target branch and target chart directory
+      if [[ "${SYNC_ENVIRONMENT}" == "staging" ]]; then
+        export TARGET_BRANCH="staging"
+        export TARGET_CHART_DIR="${CHART_DIR}" # For staging, we modify the main chart
+      else
+        export TARGET_BRANCH="dev/${SYNC_ENVIRONMENT}"
```
dhaustein, line 248: I made it so it creates not only a dev sub-dir in mender-helm but also pushes to a separate branch. I am wondering whether this is overkill though. Please let me know.

oldgiova:
Before starting to fully review this, I have some questions to address
1. why the rest of the code is commented out? I could assume this is just for PoC check and leave out the noise, right? That's perfectly fine, but the only issue is that it's harder to compare the old and the actual new code.
2. did you already tested this behavior? If not, do you want me to help you testing it on forked repositories?

dhaustein:
> Before starting to fully review this, I have some questions to address
> 
>     1. why the rest of the code is commented out? I could assume this is just for PoC check and leave out the noise, right? That's perfectly fine, but the only issue is that it's harder to compare the old and the actual new code.
> 
>     2. did you already tested this behavior? If not, do you want me to help you testing it on forked repositories?

1. That's on purpose, I wanted to isolate just the deployment part. I can un-comment everything, if it's an easier comparison, let me know, it's not a problem.
2. That would be awesome.

dhaustein:
@oldgiova I cleaned up the changes a bit, so the PR should be a bit easier to follow now. Here's the rundown on the structural changes:

* for now, disabled deployment unrelated jobs by adding `when: never` instead of commenting the code out
* moved the `.helm-version-bump` template to the top of the page, where it can make friends with the other templates
* new `mender-helm-version-bump:branch` job that uses that template
* the `SERVICE:` variable now gets all mender services, and in new format of `service_name:helm_key` (the version bump template was changed to expect this new format)
* added "docstrings" to all the jobs

---

I'll get to actually try the gitops flow now against the sandbox targets you made. :hourglass_flowing_sand:

.gitlab-ci.yml
```diff
@@ -175,6 +177,183 @@ default:
...
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers:aws-k8s-v1-master
+  variables:
+    HELM_PATCH_VERSION: ${CI_PIPELINE_ID}
+    SYNC_ENVIRONMENT: ${CI_COMMIT_REF_SLUG}
+  before_script:
+    - git config --global user.email "${GITHUB_USER_EMAIL}"
+    - git config --global user.name "${GITHUB_USER_NAME}"
+    - export DIGESTS_FOLDER=$(pwd)/.digests
+    - export PROJECT_FOLDER=$(pwd)
+  script:
```
dhaustein, line 240: The stuff under here is an already monstrous amount of bash that has the critical role of doing the updating and publishing of Helm charts and so doing the whole versioned deployments flow.

Now my problem with this is that the script is both important and long. I think it earned the honor of being extracted into a separate sh file that accepts arguments, with proper formatting, break down into functions, docstrings/comments etc instead of it being a one big blob of code that is hard to read.

On the other hand, by shoving it into yet another file we are kind of hiding the logic out of sight.

@lluiscampos What do you think about the above as an architect?

dhaustein: Or, maybe do the script aaaand make the whole thing a ci component, too? Although I don't think it's warranted when the component will be used in only one place, defeats the purpose of a "reusable" component.

lluiscampos: As an architect _external to this team_ having a dedicated script sounds good to me!

The CI of this repository already uses external scripts here and there (comes to mind the changelog updates) so I think that the Mender Server team (aka owners of this repo) will like the proposal :smiley: 

Wrt CI/CD component I don't think it is appropriate until we have the need to share it somewhere else.

PR #892, opened 2025-08-29
conf(tests): run compose tests as `linux/amd64` always

frodeha:
**Description**

The integration and acceptance tests use a `mender-artifact` binary built for `x86-64` in many tests, and these tests always fail if the `acceptance-tester` and `integration-tester` are built for `arm64` (which is the case on MacOS unless a platform is defined).

I believe this should make no difference for non MacOS users (CI included) as this should be how the "testers" are already being built on those platforms.

### DIFF
--- a/backend/tests/docker-compose.yml
+++ b/backend/tests/docker-compose.yml
@@ -3,6 +3,7 @@ include:
 
 services:
   acceptance-tester:
+    platform: linux/amd64
     scale: 0
     image: "${CI_REGISTRY_IMAGE:-localhost/backend-tester}:${CI_ACCEPTANCE_IMAGE_TAG:-acceptance}"
     build:
@@ -21,6 +22,7 @@ services:
       - workflows-worker
 
   integration-tester:
+    platform: linux/amd64
     scale: 0
     image: "${CI_REGISTRY_IMAGE:-localhost/backend-tester}:${CI_INTEGRATION_IMAGE_TAG:-integration_v2}"
     build:


kjaskiewiczz (approved):

backend/tests/docker-compose.yml
```diff
@@ -3,6 +3,7 @@ include:
 
 services:
   acceptance-tester:
+    platform: linux/amd64
```
alfrunes, line 6: Won't this conflict with machines running docker natively on arm64 (or any other arch)?

frodeha: It would I suppose, yes. I'm not sure how big of a problem that is though. Many of the tests don't run properly on non-amd64 architectures (emulated or otherwise) right now, and I guess I'm the first one to work with them locally on a different architecture. 

I suppose you could argue that having some tests succeed and some not (which is the current case) is better than having the suite outright fail on platforms that are currently not used by the team (which I assume is the case - please do correct me if I'm wrong) is superior. I basically did this modification because it was an easy way to make it work for everyone currently (or so I think anyway!).

Another solution is to compile mender-artifact for arm, add it to the project, and selectively mount the binary (amd64 / arm64) based on the `DOCKER_PLATFORM` environment variable for example. Would you prefer something like that?

I'm happy to do either. Let me know! :)

mzedel: As a fellow user of this workaround (even if it is just to run the local setup without warnings): let's do this for now and think about what to do about the machines running docker natively on arm64 once we run into them 😉

alfrunes: As long as CI is :green_circle: and nobody :scream_cat: it's all good from my end.

PR #896, opened 2025-09-01
MEN-8686: Handle error conditions during trial signup gracefully in gui

frodeha:
**Description**

While investigating what could/would cause the backend to hang occasionally (the root cause of MEN-8686) I found that the `gui` is not handling `500 - Internal Server Error`, `timeouts` and other `"connectivity errors"` from the `/tenants/trial` backend endpoint. This is what causes the UI to "hang forever" on the loading page when the `backend` is struggling (with what is yet to be determined).

Although not the root cause of [MEN-8686](https://northerntech.atlassian.net/browse/MEN-8686), this is imo worth fixing as it improves the experience of the user if/when they experience this issue in the wild.

[MEN-8686]: https://northerntech.atlassian.net/browse/MEN-8686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/js/components/login/Signup.tsx
+++ b/frontend/src/js/components/login/Signup.tsx
@@ -132,11 +132,13 @@ export const Signup = () => {
       plan: 'enterprise',
       ts: captchaTimestamp
     };
-    return dispatch(createOrganizationTrial(signup)).catch(() => {
-      setFormValues({ ...formValues, ...formData, captcha: '' });
-      setIsStarting(true);
-      setLoading(false);
-    });
+    return dispatch(createOrganizationTrial(signup))
+      .unwrap()
+      .catch(() => {
+        setFormValues({ ...formValues, ...formData, captcha: '' });
+        setIsStarting(true);
+        setLoading(false);
+      });
   };
 
   const onUserDataSubmit = (formData: UserData) => {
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/login/Signup.tsx
```diff
@@ -132,11 +132,15 @@ export const Signup = () => {
...
     };
-    return dispatch(createOrganizationTrial(signup)).catch(() => {
-      setFormValues({ ...formValues, ...formData, captcha: '' });
-      setIsStarting(true);
-      setLoading(false);
-    });
+    return dispatch(createOrganizationTrial(signup))
+      .unwrap()
+      .catch(() => {
+        setTimeout(() => {
```
mzedel: This will work with the timeout, but once the error response is caught in here, we can reset the form right away I think (with the timeout the request times out, then the UI waits 5s and only then resets the form)...

mzedel (dismissed):
Sorry you had to go through this 😬 - the reset should have been fixed in #473 ([here](https://github.com/mendersoftware/mender-server/pull/473/files#diff-39bdba35e93e5c8ab23c3d30bd06d34d2178d6c115bbe529dbf5d0419204d33eL135-R142) to be exact) - unfortunately that didn't go through 😞

It would work as is, with the timeout removed it would be a bit quicker to return the user to a useful state, but I can also adjust that later on...

Thanks a lot!

frontend/src/js/components/login/Signup.tsx
frodeha: Aight! Removed the timeout in 744dfd1deb37675b32e708266673928dce56f1a7.

And of course, happy to do it! Thanks for the quick review 😄 🙌

mzedel (approved):

PR #904, opened 2025-09-02
MEN-8649: Clean OS services from tenantadm related code 🧹

bahaa-ghazal:
Ticket: MEN-8649

Tried to not remove too much to not make OS even more different from enterprise.


backend/services/deviceauth/cache/cache.go
```diff
@@ -217,20 +172,13 @@ func (rl *RedisCache) DeleteLimit(ctx context.Context, name string) error {
 func (rl *RedisCache) Throttle(
 	ctx context.Context,
 	rawToken string,
-	l ratelimits.ApiLimits,
 	tid,
 	id,
 	idtype,
 	url,
 	action string,
 ) (string, error) {
```
alfrunes, line 180: `Throttle` but without the "Throttling" part :upside_down_face:

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -283,85 +271,12 @@ func (d *DevAuth) signToken(ctx context.Context) jwt.SignFunc {
...
-			return "", err
-		}
-	} else {
-		// ignore identity data when tenant verification is off
-		// it's possible that the device will provide old auth token or old tenant token
-		// in the authorization header;
-		// in that case we need to wipe identity data from the context
-		ctx = identity.WithContext(ctx, nil)
-	}
+	ctx = identity.WithContext(ctx, nil)
```
alfrunes, line 279: This hack should be unnecessary after the migration go gin-gonic, right?
That is, if there's no identity middleware for this API.

backend/services/useradm/user/useradm.go
```diff
@@ -417,29 +345,6 @@ func (ua *UserAdm) UpdateUser(ctx context.Context, id string, userUpdate *model.
...
 
-	if len(userUpdate.Email) > 0 && userUpdate.Email != user.Email {
-		if ua.verifyTenant {
-			err := ua.cTenant.UpdateUser(ctx,
-				idty.Tenant,
-				id,
-				&tenant.UserUpdate{
-					Name: string(userUpdate.Email),
-				},
-				ua.clientGetter())
```
alfrunes, line 428: Wow, this seems like a leftover. This should never happen under any circumstances since we removed duplicating users from tenantadm service. :memo:

alfrunes (approved):
The changes looks good to me, great job!
I created a new upstream branch as we also need to merge these changes to enterprise before committing it to `main`.

PR #911, opened 2025-09-02
MEN-8719 - allow using ai feedback for training

mzedel:


frontend/src/js/components/settings/Global.tsx
```diff
@@ -228,7 +233,12 @@ export const GlobalSettingsDialog = ({
 
   const onEditDeltaClick = () => setShowDeltaConfig(true);
 
-  const onToggleAiClick = current => saveGlobalSettings({ aiFeatures: { enabled: !current } });
+  const onToggleAiClick = useCallback(current => saveGlobalSettings({ aiFeatures: { ...aiFeatures, enabled: !current } }), [aiFeatures, saveGlobalSettings]);
+
+  const onToggleAiTrainingClick = useCallback(
+    ({ target: { checked } }) => saveGlobalSettings({ aiFeatures: { ...aiFeatures, trainingEnabled: checked } }),
```
mzedel, line 239: FYI @alfrunes - so far it's not used for anything, but will be once the feedback backend adjustment is being made... eventually we could use it for the decision on what feedback to feed back to the model 🤞... or any other key if you like 👌

mineralsfree (approved):

frontend/src/js/components/settings/Global.tsx
alfrunes: It's not so important what it's called as long as we have it tracked somewhere :memo:
I'll put it in the ticket.

PR #913, opened 2025-09-02
MEN-8644 - subscription self service UI changes

mineralsfree:


mineralsfree:
@alfrunes, This PR introduces the new subscription UI

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -57,6 +57,14 @@ const contactReasons = {
...
       </div>
     )
+  },
+  specialHandling: {
+    id: 'specialHandling',
+    alert: (
+      <div>
+        You have a special agreement with Northern.tech, please contact <SupportLink variant="email" /> for any change in your subscription.
+      </div>
+    )
```
alfrunes: The non-trial customers that does not have a Stripe subscription should fall back into the same handling as enterprise customers. That is, they will get the form to request a quote. This should bind to the current API to request plan change `POST /api/management/v2/tenantadm/tenants/{id}/plan`.

mineralsfree: The enterprise form you described is already shown. If it won't be confusing for a user which effectively have Professional or Basic plan, but shown enterprise form, I could remove this Alert information or change its description. What would you suggest?

alfrunes: I tested it locally, and I don't see the alert. However, the form looks incorrect. In the following screenshot I have a professional account which does not have a subscription in Stripe:
image
If I click the upgrade button it will show the the next step, but the button to "Change subscription" is not present.
image
Only if I manually select enterprise plan does it show the expected form:
image

mineralsfree: The input should select the enterprise form and disable the radio group as soon as GET `/billing/subscription` fails with 404, I'll double-check that on my side, but could you please confirm that this requests fails as expected in the tenant service when you test it?

alfrunes (commented):
It's a :+1: from me.
But please wait for @mzedel's for the code review.

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -298,8 +298,8 @@ export const getUserBilling = createAsyncThunk(`${sliceName}/getUserBilling`, (_
 );
 
 export const getUserSubscription = createAsyncThunk(`${sliceName}/getUserSubscription`, (_, { dispatch }) => {
-  const tasks = [dispatch(getBillingPreview({ preview_mode: 'next' })).unwrap(), dispatch(getCurrentSubscription()).unwrap()];
-  Promise.all(tasks).then(([currentPreview, currentSubscription]) => dispatch(actions.setSubscription({ ...currentPreview, ...currentSubscription })));
+  const tasks = [dispatch(getBillingPreview({ preview_mode: 'next' })), dispatch(getCurrentSubscription()).unwrap()];
```
mzedel: does the removed `.unwrap()` not make the dispatch call resolve immediately instead of after the thunk finished?

frontend/src/js/store/organizationSlice/thunks.ts
```diff
@@ -310,7 +310,11 @@ export const getBillingPreview = createAsyncThunk(`${sliceName}/getBillingPrevie
 );
 
 export const getCurrentSubscription = createAsyncThunk(`${sliceName}/getCurrentSubscription`, () =>
-  Api.get(`${tenantadmApiUrlv2}/billing/subscription`).then(res => res.data)
+  Api.get(`${tenantadmApiUrlv2}/billing/subscription`).then(res => res.data).catch(e=> {
+    if (e.response && e.response.status === 404){
+      throw new Error('Request failed with status code 404')
+    }
+  })
 );
```
mzedel: considering that the error gets eaten up otherwise, should this not forward all others error instead or use the `commonErrorHandler`? otherwise I would assume the `getUserSubscription` above would return without user feedback in case they have a regular subscription and would expect to be able to adjust it.

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -131,7 +124,7 @@ export const SubscriptionPage = () => {
   }, [stripeAPIKey]);
 
   useEffect(() => {
-    if (plan) {
+    if (plan && !specialHandling) {
```
mzedel, line 130: `specialHandling` is missing in the dependency array

Looking at it, disabling the check for the selectedAddon setting in the `useEffect` below seems risky too...

frontend/src/js/components/settings/organization/BillingDetails.tsx
```diff
@@ -26,25 +26,25 @@ const Address = props => {
...
+  const theme = useTheme();
   const displayNames = new Intl.DisplayNames('en', { type: 'region' });
   return (
     <div>
-      <div>
-        <b>{name}</b>
-      </div>
-      <div>{line1}</div>
-      <div>
+      <Typography style={{ fontWeight: 'bold' }} color={theme.palette.text.secondary}>
```
mzedel:
```suggestion
      <Typography color="text.secondary">
        <b>{name}</b>
      </Typography>
```

frontend/src/js/components/settings/organization/BillingDetails.tsx
```diff
@@ -58,7 +58,7 @@ export const BillingDetails = props => {
             {billing.address && <Address address={billing.address} email={billing.email} name={billing.name} />}
             {card && !hideCard && <CardDetails card={card} containerClass={billing.address ? 'margin-left-x-large' : ''} />}
           </div>
-          <Button className="margin-top-x-small" onClick={() => setChangeBilling(true)} size="small">
+          <Button disabled={editDisabled} variant="outlined" className="margin-top-x-small" onClick={() => setChangeBilling(true)} size="medium">
```
mzedel:
```suggestion
          <Button disabled={editDisabled} variant="outlined" className="margin-top-x-small" onClick={() => setChangeBilling(true)}>
```
to reduce places to adjust if a new button size becomes the default 🙃

frontend/src/js/common-ui/forms/TextInput.tsx
```diff
@@ -32,7 +32,9 @@ export const TextInput = ({
   type,
   validations = '',
   numericValidations = {},
-  value: passedValue = ''
+  value: passedValue = '',
+  requiredRendered = true,
+  width: inputWidth
```
mzedel: `inputWidth` seems rather misleading here, as the input is `OutlinedInput`, whereas the inputWidth is applied to the `FormControl`

frontend/src/js/common-ui/forms/Form.tsx
```diff
@@ -131,10 +131,11 @@ export const Form = ({
   onSubmit,
   showButtons,
   submitLabel,
-  submitRef
+  submitRef,
+  validateOnSubmitMode = false
```
mzedel: considering the other modes available I would rather pass the mode and default to `onChange` here (and align with RHF)

```suggestion
  mode = 'onChange'
```
and then pass on the mode + adjust the button condition accordingly...

frontend/src/js/common-ui/forms/CountrySelect.tsx
```diff
@@ -38,7 +40,9 @@ export const CountrySelect = (props: CountrySelectProps) => {
       options={countries}
       className={classes.autocomplete}
       autoHighlight
-      renderInput={params => <TextField {...params} label="Country" id={id || 'country'} />}
+      renderInput={params => (
+        <TextField {...params} error={error} helperText={helperText} label="Country or region" id={id || 'country'} style={{ width: 500 }} />
```
mzedel: The width in here seems to disagree with the width defined in the class above...

frontend/src/js/components/settings/PlanExpandedForm.tsx
```diff
@@ -36,19 +35,30 @@ export const PlanExpandedForm = (props: PlanExpandedFormProp) => {
...
 
   return (
-    <div className={className}>
-      <div className="flexbox center-aligned">
-        <TextInput hint="Billing email" label="Billing email" id="email" required validations="isEmail,trim" />
-        <MenderHelpTooltip className="required" id={HELPTOOLTIPS.planUpgradeEmail.id} />
+    <>
+      <div className={className}>
+        <Typography variant="subtitle1">Your billing details</Typography>
+        <TextInput requiredRendered={false} width={500} hint="Company name" label="Company name" id="name" required validations="isLength:2,trim" />
```
mzedel: can you maybe put the repeated `requiredRendered={false} width={500} required` in a common prop object or at least define the 500 width as a shared const?

mzedel (changes requested):

mzedel (approved):

PR #920, opened 2025-09-04
chore: update dependabot configuration

kjaskiewiczz:


merlin-northern (dismissed):
I just trust you on that one.

kjaskiewiczz:
@merlin-northern sorry, you were too fast with review.
I updated configuration to trigger all kind of updates for 4.0.x branch.

.github/dependabot.yml
```diff
@@ -42,6 +44,53 @@ updates:
         exclude-patterns:
           - "@playwright*"
           - "playwright*"
+    target-branch: main
+  # maintenance branch - 4.0.x
```
merlin-northern, line 48: will this comment go obsolete sooner or later? (below the same.

merlin-northern (approved):
I still trust you

.github/dependabot.yml
kjaskiewiczz: at the time the comment will become obsolete, the configuration will become obsolete too

mzedel (approved):
LGTM, but I assume it will get pretty noisy here once merged 🙈 + the config file will grow a lot if we have multiple releases to support 😕

kjaskiewiczz:
@mzedel yes, it can become pretty noisy, so the idea is that we will enable it and see how much effort it is actually.
If we'll decide it's too much for us, then I'll revert it.

PR #938, opened 2025-09-08
QA-1174: Migrate golangci-lint to v2

bahaa-ghazal:


bahaa-ghazal:
Should merge it to enterprise before merging it to OS.

bahaa-ghazal:
Btw, the reason why new rules was added after the migration is because the linters stylecheck, gosimple, and staticcheck has been merged inside the staticcheck. We didn't use stylechecker before, which i think is the reason for the new rules.

backend/.golangci.yml
```diff
@@ -15,121 +16,153 @@ run:
...
-    - ineffassign
     - lll
-    - staticcheck
-    - typecheck
-    - unused
+  settings:
+    staticcheck:
+      checks:
+        - all
+        # New rules that is introduced with the migration:
```
alfrunes:
```suggestion
        # New rules introduced by migration to v2
```

backend/.golangci.yml
```diff
@@ -15,121 +16,153 @@ run:
...
-linters-settings:
-  gocyclo:
-    # default is 30.
-    min-complexity: 20
+        # New rules that is nice to have:
+        # - '-ST1006' # Poorly chosen receiver name
+        # - '-ST1011' # Poorly chosen name for variable of type "time.Duration"
+        # - '-ST1019' # Importing the same package multiple times
+        # - '-QF1001' # Apply De Morgan's law
+        # - '-QF1007' # Merge conditional assignment into variable declaration
```
alfrunes: I think there are a lot of rules that would be nice to have, but I think it's better to consult with the upstream documentation if we want to enable more rules or this file could easily get bloated.
```suggestion
```

alfrunes (dismissed):

alfrunes (approved):

PR #964, opened 2025-09-14
Refactor/rhf subscription page

mzedel:


frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +81,345 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
+          </>
+        )}
+      </div>
+      <div>
+        {selectedPlan.id !== PLANS.enterprise.id && previewPrice && !specialHandling && (
+          <div className="margin-top margin-left-x-large">
+            <SubscriptionSummary
+              isPreviewLoading={isPreviewLoading}
+              plan={selectedPlan}
+              addons={selectedAddons}
```
mineralsfree, line 342: `selectedAddons` are of the wrong type causing issues in UI.  Please, adjust the `selectedAddons` or the `addons` in `SubscripionSummary`

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +72,343 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
-    setLimit(snappedValue);
+    if (value !== snappedValue) {
+      setValue('limit', snappedValue);
+    }
   };
 
-  const onSendRequest = async (message = '', requestedAddons = '') => {
-    await dispatch(
+  const onEnterpriseRequest = ({ message }: { message: string }) =>
+    dispatch(
```
mineralsfree, line 203: According to the design, we should clear the user message input once enterprise request is submitted

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
```diff
@@ -28,34 +28,39 @@ interface EnterpriseRequestExpandedProps {
...
-            className="margin-left-small"
-            color="secondary"
-            disabled={!message}
-            onClick={() => onSendRequest({ message, selectedAddons })}
-            variant="contained"
-          >
-            Submit request
-          </Button>
-        </div>
+        <Form
```
mineralsfree: This component (`EnterpriseRequestExpanded`) is not really used in the UI anymore, so maybe this should rather be removed :grimacing:

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +72,343 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
+      .unwrap()
+      .catch(error => {
+        if (!isTrial && error.message && error.message.includes('404')) {
+          setSpecialHandling(true);
+        }
+      });
+  }, [isTrial, dispatch]);
+
+  // Form submission is handled by individual components within the form
+  const onSubmit = (data: FormData) => console.log(JSON.stringify(data));
```
mineralsfree, line 421: Should it be removed?

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +72,343 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
+  const [loadingFinished, setLoadingFinished] = useState(!stripeAPIKey);
+  const [currentFormValues, setCurrentFormValues] = useState<Partial<FormData>>({});
+  const [previewPrice, setPreviewPrice] = useState<PreviewPrice>();
+  const [order, setOrder] = useState();
+  const [specialHandling, setSpecialHandling] = useState(false);
+
+  const initialValues: FormData = {
+    selectedPlan: plan.id,
+    selectedAddons: enabledAddons.reduce((acc, addon) => {
+      if (addon.enabled && !isTrial) {
```
mineralsfree: maybe `filter`  + `map`  here? :thinking: :

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +72,343 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
-    setLimit(Number(value));
-  };
+  }, [couldGetPreview, debouncedLimit, dispatch, selectedPlan.id, selectedPlan.minimalDeviceCount, selectedAddons, setOrder, setPreviewPrice]);
 
   const handleDeviceLimitBlur = (event: ChangeEvent<HTMLInputElement>) => {
     const value = Number(event.target.value);
     const snappedValue = Math.ceil(value / DIVISIBILITY_STEP) * DIVISIBILITY_STEP;
-    setLimit(snappedValue);
+    if (value !== snappedValue) {
+      setValue('limit', snappedValue);
```
mineralsfree: The limit here is not debounced, therefore not allowing user to input the new limit (instantly changing it to the snapped value)

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +72,343 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
-      (acc, addon) => {
-        acc[addon.name] = addon.enabled && !isTrial;
-        return acc;
-      },
-      {} as Record<AvailableAddon, boolean>
-    );
-    setSelectedAddons(newSelectedAddons);
-  }, [enabledAddons, isTrial]);
+    if (limit < PLANS[selectedPlan.id].minimalDeviceCount) {
+      setValue('limit', PLANS[selectedPlan.id].minimalDeviceCount);
```
mineralsfree: max device limit (1000) should also be enforced somewhere

mineralsfree (changes requested):
important things: 
- Addons in summary are not shown correctly due to changed addons object passed as a prop
- device limit input needs debounce before changing the user input
- device max limit is not enforsed (max 1000) + lower limit (current limit) is not enforsed
+ few smaller things

frontend/src/js/components/settings/EnterpriseRequestExpanded.tsx
mzedel: It's still used in the Upgrade component, but I'll remove this along with the entire Upgrade section in a separate PR 👌

frontend/src/js/components/subscription/SubscriptionPage.tsx
mzedel: there are a few scenarios where RHF will try to submit/ calls `onSubmit` (e.g. if a user presses enter in the device limit) - so this is just a cautionary noop to not cause errors in these cases... - I can of course remove the log to make it disappear completely!

frontend/src/js/components/subscription/SubscriptionPage.tsx
mzedel: should now again be handled in a more similar way to how it's currently done 👌

frontend/src/js/components/subscription/SubscriptionPage.tsx
mzedel: should now be reflected in the summary & preview requests & adjusted onBlur - so the user input remains, but the relevant notifications get set & backend error messages stay reduced...

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -69,187 +81,338 @@ const ContactReasonAlert = ({ reason }: ContactReasonProps) => (
...
+          <>
+            <Typography variant="subtitle1" className="margin-top margin-bottom-x-small">
+              2. Set a device limit
+            </Typography>
+            <TextInput
+              id="limit"
+              label="Number of devices"
+              type="number"
+              InputProps={{
+                inputProps: { min: 1, step: DIVISIBILITY_STEP },
```
mineralsfree: Setting the new minimum to 1 is somehow causing the step to become 1.

mineralsfree (changes requested):
There are 2 issues still

- it's possible to decrease device limit if you select a new addon
image
-  A step when using arrows is 1 instead of 50. 
image

frontend/src/js/components/subscription/SubscriptionPage.tsx
mzedel: Right, the step mdn entry explains it 🙈: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number#step - I still left the possibility to type something smaller though, as otherwise the value reset would kick in & the notification to reach out was gone all too soon 👌

mineralsfree (approved):
LGTM

PR #966, opened 2025-09-19
fix(deviceauth): `request_size_limit` configuration not applied

alfrunes:
Ticket: MEN-8788


kjaskiewiczz (approved):
It's better than nothing, but it won't prevent provisioning workflow to fail if the identity data is just below 1MB.

PR #967, opened 2025-09-22
MEN-8770 - feat(gui): added chinese server to login notifications

mzedel:
should show up like this: 
Screenshot 2025-09-22 at 10 37 45


oldgiova (dismissed):

frontend/src/js/store/appSlice/constants.ts
```diff
@@ -44,6 +45,7 @@ export const defaultReports = [{ ...emptyChartSelection, group: null, attribute:
 export const BEGINNING_OF_TIME = '2016-01-01T00:00:00.000Z';
 
 export const locations = {
+  cn: { key: 'cn', title: 'CN', location: 'cn.hosted.mender.io', icon: FlagCN },
```
mineralsfree: Maybe we should use `China` in title? Just in case people don't recognize flag and `CN`  :smile:

mineralsfree (dismissed):

mineralsfree (approved):

frontend/src/js/store/appSlice/constants.ts
```diff
@@ -44,6 +45,7 @@ export const defaultReports = [{ ...emptyChartSelection, group: null, attribute:
 export const BEGINNING_OF_TIME = '2016-01-01T00:00:00.000Z';
 
 export const locations = {
+  cn: { key: 'cn', title: 'China', location: 'cn.hosted.mender.io', icon: FlagCN },
```
oldgiova, line 48: Sorry for the very late catch; this should have been `hosted.mender.cn` instead. I see that this file has been refactored, where should I set this now? @mzedel

mzedel: 😬 that's now in: https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/store/src/appSlice/constants.ts#L32 - I'll set it along with some other adjustments that are needed and release new packages from there 👌

oldgiova: Thank you so much!

PR #970, opened 2025-09-22
chore: added compose files to dependabot surveillance

mzedel:


.github/dependabot.yml
```diff
@@ -136,6 +136,19 @@ updates:
...
+    directories:
+      - /
+      - /compose
+    package-ecosystem: docker-compose
+    schedule:
+      interval: monthly
+    groups:
+      backend-docker-compose-dependencies:
+        patterns:
+          - "*"
```
alfrunes, line 150: Will it discover `compose/docker-compose.enterprise.yml` and friends?...
I guess only one way to find out :man_shrugging:

alfrunes (approved):

.github/dependabot.yml
mzedel: they wrote it's working similar to the `Dockerfile` support so with the directory setting it could work, but yep we'll see 🤷

PR #971, opened 2025-09-22
ci: Change coveralls carryforward flag to `all`

alfrunes:


mzedel (approved):
Interesting! I haven't seen this to be an option in the docs, but given the current status it won't hurt 😄

PR #991, opened 2025-09-23
MEN-8733: CLI command to push identity data to inventory

alfrunes:
Came accross a couple bugs while working on the implementation:
 * 5f072ed7a3e82af1af1944a6e01a88f03272aa22
 * b7e4316916b841587330c1b47f88efa700fbcc67

I also added the conditional `If-Modified-Since` to the internal inventory API to prevent writing devices that have not changed.
**Not covered in this scope**:
 * The implementation is highly inefficient due to the use of `findAndModify` operation returns the entire device documentation upon update to reindex the text field.
   * We could change the update strategy to do the text field update asynchronously or do a get and set when the document changes.


backend/services/inventory/api/http/api_inventory.go
```diff
@@ -532,12 +542,18 @@ func (i *InternalAPI) PatchDeviceAttributesInternalHandler(
...
 	cause := errors.Cause(err)
 	switch cause {
 	case store.ErrNoAttrName:
 		rest.RenderError(c, http.StatusBadRequest, cause)
 		return
+	case store.ErrWriteConflict:
+		// Write conflict is only returned by the database if the filter conditions
+		// does not match and trigger an insert on an existing device ID.
+		// Therefore If-Unmodified-Since precondition failed.
+		rest.RenderError(c, http.StatusPreconditionFailed, cause)
```
kjaskiewiczz, line 555: I don't think this one is documented

backend/services/deviceauth/main.go
```diff
@@ -295,6 +315,28 @@ func cmdPropagateStatusesInventory(args *cli.Context) error {
...
+		defer cancel()
+	}
+	db, err := mongo.NewDataStoreMongo(makeDataStoreConfig())
+	if err != nil {
+		return err
+	}
+
+	inv := config.Config.GetString(dconfig.SettingInventoryAddr)
+	c := cinv.NewClient(inv, false)
+	rateLimiter := rate.NewLimiter(rate.Limit(args.Float64("rate-limit")), 1)
```
kjaskiewiczz, line 332: do we need upper limit for rate-limit?

alfrunes: For now it's a CLI argument, I think it's ok to leave that to the server administrator.

backend/services/inventory/api/http/api_inventory.go
alfrunes: Updated the specs :memo:

kjaskiewiczz (approved):

PR #992, opened 2025-09-23
ci: disable curl globbing when submitting coverage data

alfrunes:


mzedel (approved):
Ok for me, but I'm not sure if the `-f` is really needed as with all the failures we come across from coveralls the error codes might at least point to either handled or new issues 🙃... OTOH there's likely not a lot we can do in either case and _retry after wait_ is often the more promising option...

alfrunes:
> Ok for me, but I'm not sure if the `-f` is really needed as with all the failures we come across from coveralls the error codes might at least point to either handled or new issues 🙃... OTOH there's likely not a lot we can do in either case and _retry after wait_ is often the more promising option...

Yeah, since the job is allowed to fail I figured it would be fine. I think it's even worse to silently mark it as success even if the job sends a bad request.

PR #996, opened 2025-09-25
test(gui): made rapid pagination test disregard button actionability

mzedel:
+ now rely more on final page info instead
- forcing the click turned out to still check for an actionable button

### DIFF
--- a/frontend/tests/e2e_tests/integration/04-deployments.spec.ts
+++ b/frontend/tests/e2e_tests/integration/04-deployments.spec.ts
@@ -150,8 +150,12 @@ test.describe('Deployments', () => {
     // 10 clicks as anything leading outside of the 50 + something releases present (considering the 10 item page size)
     for (let clickAttempt = 0; clickAttempt < 10; clickAttempt++) {
       const paginationButton = page.getByRole('button', { name: 'next' });
+      if (await paginationButton.isDisabled()) {
+        continue;
+      }
       await paginationButton.click({ noWaitAfter: true, force: true });
     }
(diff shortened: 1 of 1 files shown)


oldgiova (approved):

frontend/tests/e2e_tests/integration/04-deployments.spec.ts
```diff
@@ -150,8 +150,12 @@ test.describe('Deployments', () => {
     // 10 clicks as anything leading outside of the 50 + something releases present (considering the 10 item page size)
     for (let clickAttempt = 0; clickAttempt < 10; clickAttempt++) {
       const paginationButton = page.getByRole('button', { name: 'next' });
+      if (await paginationButton.isDisabled()) {
+        continue;
+      }
       await paginationButton.click({ noWaitAfter: true, force: true });
     }
+    await expect(page.getByText(/-([5|6]\d) of \1/)).toBeVisible(); // depending on the deployment speed of other tests there might be slightly more than 50
```
merlin-northern, line 158: this regexp puzzles me. 51 of 51? we assume always the last page here? why not just \d+ then?

merlin-northern (approved):

frontend/tests/e2e_tests/integration/04-deployments.spec.ts
mzedel: Yep we have to make sure it's at least 50 (the test creates 50 deployments to make the pagination somewhat useful), but may be a few more (from other tests against the same environment) and we have to make sure to land on the last page (so no clicks have caused an overflow or haven't been picked up due to delays in the interaction/ list retrieval). Does this explain it @merlin-northern?

PR #997, opened 2025-09-25
MEN-8723 - chore(gui): moved deployment retry setting to setting section

mzedel:
- this is to reduce confusion & clarify the permanent "setting" character

Ticket: MEN-8723

~this needs: https://github.com/NorthernTechHQ/nt-gui/pull/731~

in deployment creation:
Screenshot 2026-05-22 at 14 27 41

in settings:
Screenshot 2026-05-22 at 14 27 52

### DIFF
--- a/frontend/src/js/components/deployments/__snapshots__/CreateDeployment.test.tsx.snap
+++ b/frontend/src/js/components/deployments/__snapshots__/CreateDeployment.test.tsx.snap
@@ -1,6 +1,4069 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`CreateDeployment Component > renders correctly 1`] = `undefined`;
+exports[`CreateDeployment Component > renders correctly 1`] = `
+.emotion-0 {
+  position: fixed;
+  z-index: 1300;
+  right: 0;
+  bottom: 0;
+  top: 0;
+  left: 0;
+  z-index: 1200;
+}
+
+.emotion-0 .MuiDrawer-paper {
+  width: 100vw;
+  min-width: 0;
+}
+
+@media (min-width:600px) {
+  .emotion-0 .MuiDrawer-paper {
+    width: 60vw;
+    min-width: 600px;
+  }
+}
+
+.emotion-0 .MuiDrawer-paper>.MuiDivider-root {
+  margin-left: -64px;
+  margin-right: -64px;
+}
+
+.emotion-1 {
+  position: fixed;
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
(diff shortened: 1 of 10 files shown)


frontend/src/js/components/settings/Global.tsx
```diff
@@ -191,6 +201,22 @@ export const GlobalSettings = () => {
...
           />
         )}
+        {canManageUsers && (
+          <SettingsItem
+            title="Default update attempts"
+            description="Set the default number of times each device will attempt an update. This can be overridden when creating a new deployment."
+            secondary={
+              <NumberField
+                min={0}
+                max={100}
```
mineralsfree, line 216: The input accepts out-of-range values, so we need validation here (there is no validation on the backend).

mineralsfree: Or maybe `allowOutOfRange=false` but better ask UX :)

frontend/src/js/components/settings/Global.tsx
```diff
@@ -137,14 +139,22 @@ export const GlobalSettings = () => {
     if (!window.sessionStorage.getItem(settingsKeys.initialized) || !timer.current || !canManageUsers || !isValidInterval(debouncedOfflineThreshold)) {
```
mineralsfree: minor, but invalid `debouncedOfflineThreshold` will cause `retries` not to change

frontend/src/js/components/deployments/__snapshots__/CreateDeployment.test.tsx.snap
```diff
@@ -4396,363 +4130,182 @@ label[data-shrink=false]+.MuiInputBase-formControl .emotion-16:focus::-ms-input-
...
+      min="1"
+      step="1"
+      style="clip-path: inset(50%); overflow: hidden; white-space: nowrap; border: 0px; padding: 0px; width: 1px; height: 1px; margin: -1px; position: fixed; top: 0px; left: 0px;"
+      tabindex="-1"
+      type="number"
+      value="1"
+    />
+    <svg
+      aria-hidden="true"
+      aria-label="Default is undefined. This can be changed in the global settings"
```
mineralsfree: Maybe it's better to pass the proper props in `CreateDeployment.test.tsx`?

mineralsfree (changes requested):
I think we need to take care of the limits, as the current `max` in `NumberInput` only affects the stepper being disabled. Others are minor

frontend/src/js/components/settings/Global.tsx
mzedel: 👌 separated the 2 now in different effects

frontend/src/js/components/settings/Global.tsx
mzedel: I'll keep it aligned with the current implementation for now as this is dictated by the backend and disconnected from the deployments context

mineralsfree (approved):

PR #998, opened 2025-09-25
test(gui): adjusted timeouts after jsdom upgrade

mzedel:


oldgiova (approved):

frontend/src/js/components/settings/webhooks/Webhooks.test.tsx
```diff
@@ -87,7 +87,7 @@ describe('Webhooks Component', () => {
     await waitFor(() => expect(screen.queryByText(/webhook details/i)).toBeNull());
   });
 
-  it('can be configured', { timeout: TIMEOUTS.refreshDefault }, async () => {
+  it('can be configured', { timeout: 3 * TIMEOUTS.fiveSeconds }, async () => {
```
merlin-northern, line 90:
```suggestion
const UnitTimeoutDefault = 3 * TIMEOUTS.fiveSeconds
```
or perhaps something in `TIMEOUTS` ?

merlin-northern (approved):

frontend/src/js/components/settings/webhooks/Webhooks.test.tsx
mzedel: There is a default of 5 seconds (https://vitest.dev/config/#testtimeout), which is more than enough for most of the tests!
However some of the tests (as all adjusted here) do more extensive user interactions than others and these shouldn't take that much longer than the expected timeouts in order to point to actual slow downs in the interaction... so give some time, but not too much or we won't see these components getting lazy 🫣

PR #1000, opened 2025-09-26
ci: openapi specs: ASSEMBLE!

mzedel:


frodeha:
Okay, I'm going to go ahead and stick my hand in the fire a little bit here... I personally really like this, Manuel - thank you for putting it together! 

I have played around with it locally this morning and I think it's going to serve us well. The state of the spec(s) are really quite "rough" after the merge, but that's not really the fault of the work you've done here. I especially like:
* The breakdown of service spec into sub specs (management, device and internal). I think it will be easier to work with them this way. The convention/rule should probably be that there's max 4 such sub specs though (management, device, internal, common) to hopefully hit a sweet spot in terms of manageability. We can always iterate though.
* The introduction of the shared definitions for common API types. I can't really see much reason for these not to be shared and copy/pasting them all over is just error prone. 
* The choice of Redocly as our tool of choice. It's really quite powerful and whatever it can't do out of the box we can easily make plugins for. I wrote one this morning for removing the `root.***` tags (which rubs me the wrong way - haha) and it was straight forward enough.

All in all, big 👍 from me on this! 🙌 

--- 

Further things I would love for us to do (not in this PR and I'm happy to help) is to establish some structure/rules:
* Can we converge/agree on one (and only one) filename pattern for the specs? I appreciate that there's probably many opinions on this, but right now there doesn't seem to be any established format (at least not one that stands out to me). For example, to me it would be infinitely helpful if we adapt, for example, `<service_name>_<api_type>_<api_version>.yaml` (`deployments_devices_v1.yaml` `deployments_internal_v1.yaml`) or something like that for all specs. Happy to contribute on this, just want to see if there's an appetite.
* Tags. Right now we're prefixing tags with the filename to avoid collisions. I'm thinking we could probably write a plugin that prefixes tags with the Name of the root object of the individual spec files and establish a _rule_ that this Name is always of some format ("Name: Deployments Devices" or something). Thoughts? Again, not asking for this to be done by you or in this PR - just brainstorming.

frodeha (approved):

PR #1002, opened 2025-10-01
MEN-8727 - feat(gui): let device deployments history access our log viewer as well

mzedel:


frontend/src/js/components/devices/device-details/Deployments.tsx
```diff
@@ -154,8 +142,13 @@ export const Deployments = ({ device }) => {
   const [page, setPage] = useState(DEVICE_LIST_DEFAULTS.page);
   const [perPage, setPerPage] = useState(10);
   const [isChecking, setIsChecking] = useState(false);
+  const [showsDeploymentLogForId, setShowsDeploymentLogForId] = useState('');
   const { classes } = useStyles();
+  const { hasAiEnabled, isHosted } = useSelector(getFeatures);
+  const isPreview = useSelector(getIsPreview);
+  const canAi = isHosted && (isPreview || hasAiEnabled);
```
aleksandrychev, line 149: I guess canAi will be used in different places, maybe move to selectors, or getFeatures? not sure what place is better

aleksandrychev (approved):

frontend/src/js/components/devices/device-details/Deployments.tsx
mzedel: I hope soon enough we'll have it enabled on hosted, so that we can rely on `hasAiEnabled`, as that would then be enough to cover on prem too...

PR #1003, opened 2025-10-01
ME-568 - fix(gui): fixed an issue that could cause the deployment device limit to be ignored

mzedel:
also fixes https://northerntech.atlassian.net/browse/MEN-8642

### DIFF
--- a/frontend/src/js/components/deployments/deployment-wizard/DeviceLimit.tsx
+++ b/frontend/src/js/components/deployments/deployment-wizard/DeviceLimit.tsx
@@ -39,15 +39,15 @@ export const DeviceLimit = props => {
 
   const [shouldLimit, setShouldLimit] = useState(false);
   const [error, setError] = useState('');
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(numberDevices);
 
   const { classes } = useStyles();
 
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/deployments/deployment-wizard/DeviceLimit.tsx
```diff
@@ -96,7 +96,7 @@ export const DeviceLimit = props => {
         <div className={classes.limitSelection}>
           Finish deployment after{' '}
           <FormControl error={!!error}>
-            <OutlinedInput value={value} placeholder={String(numberDevices)} onChange={handleLimitChange} type="text" />
+            <OutlinedInput value={value} onChange={handleLimitChange} type="text" />
```
aleksandrychev, line 99: maybe type=number? but I guess then UI will look differently

aleksandrychev (approved):

frontend/src/js/components/deployments/deployment-wizard/DeviceLimit.tsx
mzedel: that's following the MUI recommendation from here: https://mui.com/material-ui/react-text-field/#type-quot-number-quot

PR #1021, opened 2025-10-06
Ci/test script adjustments

mzedel:
successful staging job is [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/11657646041), key fix is in the last commit 👌


kjaskiewiczz (approved):
thanks!

PR #1029, opened 2025-10-10
Ci/feedback disable

mzedel:
staging test runs:
- https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/11677315107
- https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/11677315123


kjaskiewiczz (approved):
thanks for fixes!

PR #1042, opened 2025-10-17
spec surprises

mzedel:
especially the last commit is what I'm not sure about, as there was an issue about it: https://northerntech.atlassian.net/browse/MEN-7152, but the response is still not in line with what [mender-artifact](https://github.com/mendersoftware/mender-artifact/blob/master/artifact/metadata.go#L579) has or the [device API](https://docs.mender.io/api/#device-api-deployments-schemas-update) says vs. [management API](https://docs.mender.io/api/#management-api-deployments-v2-schemas-update) - yet they both share the same model...

### DIFF
--- a/backend/docs/api/common/schemas.yaml
+++ b/backend/docs/api/common/schemas.yaml
@@ -65,3 +65,21 @@ components:
           description: ID of the new device.
       required:
       - device_id
+    Scope:
+      type: string
+      description: |
+        The scope of the attribute.
+
+        Scope is a string and acts as namespace for the attribute name.
+
+        * __inventory__: Attributes reported by the device.
+        * __system__: Attributes populated by the mender-server.
+        * __identity__: Device's identity attributes provided in the device's auth request.
+        * __monitor__: Attributes populated by the monitoring add-on.
+        * __tags__: User-defined attributes associated with the device.
+      enum:
+        - system
+        - identity
+        - inventory
+        - monitor
+        - tags
--- a/backend/docs/api/deployments/internal_v1.yaml
+++ b/backend/docs/api/deployments/internal_v1.yaml
@@ -742,12 +742,6 @@ components:
       - deployment
       - device
       type: object
-    MetadataAny:
-      description: meta_data is an object of unknown structure as this is dependent
-        of update type (also custom defined by user)
-      additionalProperties:
-        type: string
-      type: object
     LastDeviceDeploymentsStatuses:
       example:
         device_deployment_last_statuses:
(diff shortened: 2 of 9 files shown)


frodeha (commented):
In reference to the last commit (changing the type of Update.Metadata). What problem (I assume you encountered one that lead you to this change) are you fixing? The change (although I don't disagree with it) should not result in any difference in the actual server response. The empty interface (`interface{}`) in Go is similar to `any` in Typescript in that it can hold literally anything. 

Update.Metadata is, from what I can tell, only used in one place and there it's assigned a `map[string]interface{}`, and that will always be serialized to a JSON object (which is what the spec says it should be). As I said I don't disagree at all with the fix, it's good as far as I'm concerned, but I'm not sure I see how it fixes a/your problem?

For the rest of the PR it looks good as far as I can tell, but I'd love for someone else to chime in as well!

mzedel:
> should not result in any difference in the actual server response

that's what I think was the assumption on implementation, but using an artifact with metadata as example this currently leads to:
```json
[
  {
    "Key": "containers",
    "Value": ["debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623"]
  }
]
```
or an empty array if there's none.

What it should be:
```json
{
  "containers": ["debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623"]
}
```
and here the metadata property gets omitted if there's none.

The device API seems to return this + mender-artifact itself shows this/ writes this to the artifact, but v1 of the API also specifies this, e.g. the `Artifact` returned here: https://docs.mender.io/api/#management-api-deployments-show-artifact will contain an `Update` with an object and not a list - 

In the generated types this results in one endpoint expecting to process an array, whereas the other an object, this then causes a type error when retrieving data from one or the other endpoint when the UI gets to merge the data from both locations... Since we don't handle this specific data in the UI at all, we can ([here](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/store/src/releasesSlice/index.ts#L29-L30)) ignore that part of the response objects - but the underlying issue to me looks more like an implementation bug that got aligned into documentation 🫣.

frodeha:
Right, I see now how this change fixes the problem. When we're reading the artifact from the database it's returned to us by the mongodb driver in an intermediate representation that represents map properties as elements with a `Key` ("containers") and `Value` (list of strings) properties. Since we don't specify any type to unmarshal this representation into (e.g `interface{}`), the mongodb driver just uses the intermediate representation as is. This intermediate representation is then marshalled as-is to json - e.g a list of objects with `Key` and `Value`. 

However, if we specify a destination type of `map[string]interface{}` the mongodb driver transforms the intermediate representation transparently and we end up with the correct data in the correct way. 

I tested it locally and the change does indeed work as described and in retrospect it's obvious why. Nice - LGTM!

frodeha (approved):

PR #1053, opened 2025-10-27
test: Make ssh_is_opened timeout equal to provided value

michalkopczan:
Fix ssh_is_opened so that the wait parameter is correctly used, by forwarding it to MenderDevice::run. Before the fix, the timeout was forwarded to MenderDevice::run, while the run itself was tried up to "wait" times on top of that. This resulted in 600*600 seconds (100 hours) of total timeout. If the client device did not become ready for any reason it caused the pipelines to seemingly hang.

Changelog: None
Ticket: QA-1147

### DIFF
--- a/backend/tests/integration/testutils/infra/device.py
+++ b/backend/tests/integration/testutils/infra/device.py
@@ -179,19 +179,15 @@ class MenderDevice:
         """
         waited = -1
         t0 = int(time.time())
-        raise_exception = None
-        for _ in redo.retrier(max_sleeptime=wait, attempts=wait, sleeptime=1):
-            try:
-                self.run("true", hide=True, wait=wait)
-                raise_exception = None
-                break
-            except Exception as e:
-                raise_exception = e
-            finally:
-                waited = int(time.time()) - t0
-        if raise_exception:
-            logger.error("Can't open ssh after %d s of waiting and trying" % waited)
-            raise (raise_exception)
+        try:
+            self.run("true", hide=True, wait=wait)
+        except Exception as e:
+            waited = int(time.time()) - t0
+            logger.error(
+                "Can't open ssh to %s after %d s of waiting and trying"
+                % (self.host_string, waited)
+            )
+            raise (e)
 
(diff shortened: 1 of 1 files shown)


michalkopczan:
I added reviewers based on gut feeling.
I'd like to gen a view on whether this can break anything that I may not expect (i.e. it is OK for checking if the client device in tests got up correctly, but want to make sure if it isn't used anywhere else that I may have missed).

If there's anyone more relevant, please add him. Thanks!

backend/tests/integration/testutils/infra/device.py
```diff
@@ -179,19 +179,15 @@ def ssh_is_opened(self, wait=10 * 60):
...
-                raise_exception = e
-            finally:
-                waited = int(time.time()) - t0
-        if raise_exception:
-            logger.error("Can't open ssh after %d s of waiting and trying" % waited)
-            raise (raise_exception)
+        try:
+            self.run("true", hide=True, wait=wait)
+        except Exception as e:
+            waited = int(time.time()) - t0
```
merlin-northern, line 185: does waited make sense here now?

michalkopczan: No really. Didn't want to remove it, because I didn't know if maybe some tests or anyone really, depends on this message.
So if you or @kjaskiewiczz don't see a problem with removing it, I will do it.

vpodzime: Let me vote for keeping it. It's cheap and if it differs from the expected wait time significantly, it can lead us to other issue(s).

vpodzime (approved):
Looks good to me. (but I have no authority here)

michalkopczan:
@merlin-northern , @kjaskiewiczz , can I merge? Any other comments fro your side? Thanks.

kjaskiewiczz (approved):

merlin-northern:
> @merlin-northern , @kjaskiewiczz , can I merge? Any other comments fro your side? Thanks.

if KJ approved the discussion is over.

PR #1055, opened 2025-10-29
Fix list artifacts breaking change

frodeha:
**Description**
Reverts the breaking API change introduced in https://github.com/mendersoftware/mender-server/pull/1042/commits/33fae571932e14ca5b16afd26bc8e410b5c2a92b by adding a new `metadata` property on the Release type that is of the correct type while maintaining the old (and incorrectly formatted) `meta_data` property as-is.

I have also gone through the API documentation and updated examples and type references while also adding a note to the `meta_data` type indicating that it is deprecated and clients should rather use `metadata`. 

Thanks to Alf's contribution the documented type of the old and deprecated `meta_data` property in the API spec(s) has been updated to be in line with what the server actually produces (e.g reflect the bug in the API spec). This should hopefully result in consistent type generation for mgmt v1+v2, internal and device APIs. 

I suppose technically this is a breaking change to the API spec of (at least) mgmt v2. I would argue that this is worth doing because a) the server never responded with data that respected this spec to begin with and b) if anyone generated types from these they would either have to manually change them (in which case they should be some version of "ok" with this) or the de-serialization of the data simply wouldn't work.

Example response from any endpoint returning an artifact:
```bash
{
  "id": "b9d00e99-fad0-456c-904a-0c6f70f17073",
  "name": "test-1.15",
  "device_types_compatible": [
    "qemux86-64"
  ],
  "info": {
    "format": "mender",
    "version": 3
  },
  "signed": false,
  "updates": [
    {
      "type_info": {
        "type": "container-setup"
      },
      "files": null,
      # This is the new and correctly formatted `metadata` property
      "metadata": {
        "containers": [
          "debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623",
          "hello-world@sha256:2557e3c07ed1e38f26e389462d03ed943586f744621577a99efb77324b0fe535"
        ]
      },
      # This is the old and incorrectly formatted `meta_data` property
      "meta_data": [
        {
          "Key": "containers",
          "Value": [
            "debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623",
            "hello-world@sha256:2557e3c07ed1e38f26e389462d03ed943586f744621577a99efb77324b0fe535"
          ]
        }
      ]
    }
  ],
  "artifact_provides": {
    "artifact_name": "test-1.15"
  },
  "artifact_depends": {
    "device_type": [
      "qemux86-64"
    ]
  },
  "size": 5120,
  "modified": "2025-10-29T13:20:19.157Z"
}

### DIFF
--- a/backend/services/deployments/docs/management_api.yml
+++ b/backend/services/deployments/docs/management_api.yml
@@ -803,6 +803,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -831,6 +832,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -861,6 +863,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -974,6 +977,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -1002,6 +1006,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -1032,6 +1037,7 @@ paths:
                             checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
+                        metadata: {}
                         meta_data: {}
                     artifact_provides:
                       artifact_name: "test"
@@ -1418,6 +1424,7 @@ paths:
                       checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                       size: 123
                       date: 2016-03-11T13:03:17.063+0000
+                  metadata: {}
                   meta_data: {}
               artifact_provides:
                 artifact_name: "test"
@@ -1969,6 +1976,7 @@ definitions:
                   checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                   size: 123
                   date: 2016-03-11T13:03:17.063+0000
+              metadata: {}
               meta_data: {}
           artifact_provides:
             artifact_name: "test"
@@ -2032,6 +2040,7 @@ definitions:
                     checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                     size: 123
                     date: 2016-03-11T13:03:17.063+0000
+                metadata: {}
                 meta_data: {}
             artifact_provides:
               artifact_name: "test"
@@ -2085,10 +2094,14 @@ definitions:
           type: array
           items:
             $ref: "#/definitions/UpdateFile"
+        metadata:
+          type: object
+          description: an object of unknown structure as this is dependent of update type (also custom defined by user)
         meta_data:
           type: object
           description: |
-              meta_data is an object of unknown structure as this is dependent of update type (also custom defined by user)
+              Deprecated: Please use `metadata` instead.
+              An object of unknown structure as this is dependent of update type (also custom defined by user)
   ArtifactInfo:
       description: |
           Information about artifact format and version.
@@ -2178,6 +2191,7 @@ definitions:
               checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
               size: 123
               date: 2016-03-11T13:03:17.063+0000
+          metadata: {}
           meta_data: {}
       artifact_provides:
         artifact_name: "test"
@@ -2317,6 +2331,7 @@ definitions:
                   checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                   size: 123
                   date: 2016-03-11T13:03:17.063+0000
+              metadata: {}
               meta_data: {}
           artifact_provides:
             artifact_name: "test"
@@ -2345,6 +2360,7 @@ definitions:
                   checksum: cc436f982bc60a8255fe1926a450db5f195a19ad
                   size: 123
                   date: 2016-03-11T13:03:17.063+0000
+              metadata: {}
               meta_data: {}
           artifact_provides:
             artifact_name: "test"
--- a/backend/services/deployments/docs/management_api.yml
+++ b/backend/services/deployments/docs/management_api.yml
@@ -804,7 +804,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -833,7 +833,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -864,7 +864,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -978,7 +978,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -1007,7 +1007,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -1038,7 +1038,7 @@ paths:
                             size: 123
                             date: 2016-03-11T13:03:17.063+0000
                         metadata: {}
-                        meta_data: {}
+                        meta_data: []
                     artifact_provides:
                       artifact_name: "test"
                       rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
@@ -1425,7 +1425,7 @@ paths:
                       size: 123
                       date: 2016-03-11T13:03:17.063+0000
                   metadata: {}
-                  meta_data: {}
+                  meta_data: []
               artifact_provides:
                 artifact_name: "test"
                 rootfs-image.checksum: "32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99"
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: backend/services/deployments/model/update.go [hunk 1]
```diff
@@ -43,7 +44,44 @@ type UpdateFile struct {
...
+	//     "<key>" : <value>,
+	//     "<key>": <value>
+	//   }
+	// Unfortunately we can't change the just type of MetaData as that breaks clients
+	// (mender-cli for example) that are programmed to consume the
+	// incorrect representation (despite MetaData being documented to be a JSON object).
+	// We therefore maintain Meta_Data for backwards compatibility. It is encoded in
+	// the format of the incorrect representation in the custom `MarshalJSON`
+	// member function below.
+	Meta_Data interface{} `json:"meta_data,omitempty"`
```
alfrunes:
```suggestion
	MetaData map[string]interface{} `json:"metadata,omitempty" valid:"optional" bson:"-"`
	// Meta_Data is the deprecated API representation of MetaData.
	// By mistake the value of MetaData was exposed in the API formatted as
	// the intermediate mongodb driver representation of a map[string]interface{}
	// which looks something like this:
	//   "meta_data": [
	//	   { "Key": "<key>", "Value": <value> },
	//	   { "Key": "<key>", "Value": <value> }
	//   ]
	// instead of it's intended format as a JSON object which looks like this:
	//   "meta_data": {
	//     "<key>" : <value>,
	//     "<key>": <value>
	//   }
	// Unfortunately we can't change the just type of MetaData as that breaks clients
	// (mender-cli for example) that are programmed to consume the
	// incorrect representation (despite MetaData being documented to be a JSON object).
	// We therefore maintain Meta_Data for backwards compatibility. It is encoded in
	// the format of the incorrect representation in the custom `MarshalJSON`
	// member function below.
	Meta_Data interface{} `json:"meta_data,omitempty" bson:"metadata,omitempty"`
```

alfrunes: Maybe we also need to handle the bson serialization as well to ensure that we populate both fields.

frodeha: Hmm, I actually only intended to populate the one field (`MetaData map[string]interface{}`) in the database (e.g the opposite way of how maybe you thought about it?). In my mind the `Meta_Data` is more of a shadow/virtual property that's populated based on whatever data is in the `MetaData` at JSON serialization time. This is also why I didn't add the `bson` annotations (although I can add `bson:"metadata,omitempty"` to `MetaData map[string]interface{}`) because `Meta_Data` isn't read from the database at all post this change.

alfrunes: Isn't this data coming from the database already? What I was concerned about is that we change the type of existing data in the database.

frodeha: Yeah, it's coming from the database. However, in the database it's stored in the correct format (e.g `map[string]interface{}`) as far as I can tell (it does in my local anyway!).

We store the metadata exactly how it's [provided to us by mender-artifact](https://github.com/mendersoftware/mender-server/blob/main/backend/services/deployments/app/app.go#L1068) (type is `map[string]interface{}`) . This gives us a document in the database that looks something like like this:
```
{
    _id: 'c124e31e-44ce-43c7-ad73-5f8e9c25d449',
    meta: { description: '' },
    meta_artifact: {
      name: 'test-1.15',
      device_types_compatible: [ 'qemux86-64' ],
      info: { format: 'mender', version: Long('3') },
      signed: false,
      updates: [
        {
          typeinfo: { type: 'container-setup' },
          files: null,
          metadata: {
            containers: [
              'debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623',
              'hello-world@sha256:2557e3c07ed1e38f26e389462d03ed943586f744621577a99efb77324b0fe535'
            ]
          }
        }
      ],
      provides: { artifact_name: 'test-1.15' },
      provides_idx: [ { key: 'artifact_name', value: 'test-1.15' } ],
      depends: { device_type: [ 'qemux86-64' ] },
      depends_idx: [ { device_type: 'qemux86-64' } ]
    },
    size: Long('5120'),
    modified: ISODate('2025-10-29T12:43:43.708Z')
}
```

Before, we decoded the Metadata property of the Update type into an empty interface. When we do that, the bson decoder doesn't know what destination type we actually want and puts it's representation of the metadata object in there. Whatever type that is (I haven't prioritized figuring that out) it marshals to JSON like this:

```
"metadata": [
  {
    "Key": "containers",
    "Value": ["debian@sha256:066051f6674f6a3293bbd5a190081b1ae7fcae655a3884db59ebb3a2831da623",   'hello-world@sha256:2557e3c07ed1e38f26e389462d03ed943586f744621577a99efb77324b0fe535']
  }
]
```

In this PR I suggest that we decode the data in the database into a new property of the correct type (map[string]interface{}) - I named this `metadata` (🤷) - and post-process this to also populate the old `meta_data` property in the old (but wrong) format so we maintain backwards compatibility.

alfrunes: You're totally right! What an interesting choice of default for de-serializing an object to an `interface{}`.

alfrunes (dismissed):

backend/services/deployments/docs/management_api.yml
```diff
@@ -2085,10 +2094,14 @@ definitions:
...
             $ref: "#/definitions/UpdateFile"
+        metadata:
+          type: object
+          description: an object of unknown structure as this is dependent of update type (also custom defined by user)
         meta_data:
           type: object
           description: |
-              meta_data is an object of unknown structure as this is dependent of update type (also custom defined by user)
+              Deprecated: Please use `metadata` instead.
+              An object of unknown structure as this is dependent of update type (also custom defined by user)
```
alfrunes: Isn't this an array of key/value objects?

backend/services/deployments/docs/management_api_v2.yml
```diff
@@ -834,11 +840,14 @@ definitions:
...
+      metadata:
+        type: object
+        description: an object of unknown structure as this is dependent of update type (also custom defined by user
       meta_data:
         type: array
         description: |
-          meta_data is an array of objects of unknown structure as this
-          is dependent of update type (also custom defined by user)
+          Deprecated: Please use `metadata` instead.
+          An object of unknown structure as this is dependent of update type (also custom defined by user)
```
alfrunes, line 850: ... like this one. Is this bug only for the v2 API?

backend/services/deployments/docs/management_api.yml
alfrunes: https://github.com/frodeha/mender-server/pull/1

backend/services/deployments/docs/management_api_v2.yml
mzedel: yes, `meta_data` is correctly defined in device api + management v1, but wrong in v2. For the UI that makes the workaround also pretty useless as it would have to check what api version the data comes from when merging data in the UI... luckily we don't use any of that data so far, but whoever does will have problems.

backend/services/deployments/docs/management_api.yml
mzedel: it's not defined and just happens to be an array of key/ value objects now, but I would imagine it could change with a different db driver... the prior description is still here: https://docs.mender.io/api/#management-api-deployments-v2-schemas-update

frodeha: > Isn't this an array of key/value objects?

Right, yes. I didn't change it because I didn't want to further break compatibility. Thinking more closely about it now though that doesn't make any sense as anyone who's generated types from the wrong spec would fail to use them to de-serialize. I went ahead and merged in your PR there, Alf, thank you!

> it's not defined and just happens to be an array of key/ value objects now, but I would imagine it could change with a different db driver

That changes with this PR though. Now it's well defined to always be in this format at a code level (in all it's incorrectness) even if we swap drivers or whatever.

backend/services/deployments/docs/management_api_v2.yml
frodeha: When the type definitions are equivalent (which I believe they will be now that Alf's PR is merged into this one) the workaround should work right? Or am I missing something? I understand how it would not be when the spec was different obviously - an oversight on my part - but is there anything else making the workaround less than good?

alfrunes: > yes, meta_data is correctly defined in device api + management v1, but wrong in v2.

Are you sure? Looking at the [mender-cli code](https://github.com/mendersoftware/mender-cli/blob/b99768eaa2163d58cfc16c339df17a8582682082/client/deployments/client.go#L60) it is using the v1 endpoint for listing artifacts and the metadata has been a slice since the feature was introduced.

mzedel: No... because some endpoints will return meta data in the proper format in meta_data, yet the list endpoint will return both in meta_data and metadata, since the UI has to assemble data from multiple endpoints.
As an example: if I were to get artifact details from the show artifact endpoint (so `meta_data` proper format) and then retrieve the list of artifacts (v2 endpoint, botched `meta_data`, proper `metadata`) the typing will raise a conflict and I'll have to handle this separately (separate types for both endpoints, excluding any kind of metadata references, separate handling for both). With this change here, I get to pick the type information per endpoint and use this as a shortcut, but at this point I'll rather just drop all metadata info and only care about it if we ever use anything of this in the UI...

frodeha: But all endpoints that previously returned artifacts with `meta_data` (botched or not) will now also return `metadata` (always non botched). Can't you just use `metadata` and ignore `meta_data`?

mzedel: @alfrunes yes: https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/store/src/releasesSlice/index.ts#L30 - that exists because of this, https://docs.mender.io/api/#management-api-deployments-show-artifact gets you to the definition: https://docs.mender.io/api/#management-api-deployments-schemas-update

mzedel: @frodeha of course using it is not a problem, but the types will be generated purely based on the specs and they will clash - the key point of the type generation would be to get a reliable info about what we get from the backend, that gets somewhat diminished if we have to ignore parts of them... just to be clear: it's not a problem for the UI, I'm aware what's going on, but external consumers will likely run into the same issue and they will lack context

frodeha: Yeah, sure, fair - I agree. That's however why I thought (ref above) that we should fix the specs so it reflects what the server actually responds with also for `meta_data` (which is what was done in Alf's PR - but I think I need to have another once over to make sure it's all done).

Specifically this means that we update specs for all deployment APIs to document that `meta_data` is a list of objects (which it always is now after this PR). That should result in generated types that can be trusted (I would think/hope) that have `meta_data` as a list and `metadata` as an object.

This is I suppose technically a breaking change to the API spec of (at least) mgmt v2. I would argue that this is worth doing because a) the server never responded with data that respected it to begin with and b) if anyone generated types from these they would either have to manually change them (in which case they should be some version of "ok" with this (this is what you have done to work around this I believe)) or the de-serialization simply wouldn't work.

frodeha: I (optimistically) went ahead and did so in 937ed3bce5358bea9b7bd67eecfe8138ce813c8f. Please let me know if you disagree on any of this upon re-review :)

alfrunes: Please correct me if I'm wrong, but to my understanding the server makes no difference between the different endpoint when it comes to serializing the `meta_data`. All endpoints that retrieves an `Update` type (whether it's `/releases` or `/artifacts`, v1 or v2) the server always returned a list of key/value pair which is not aligned with the API specs. It does have a different JSON serialization, but that is never used AFAICT since this data is coming from the artifact metadata. Which means that if there are any clients out there using the specs, they must have used the same workaround as you highlighted.
On the other hand, we have only received one report of the issue in mender-cli, so I'm starting to think whether we should fix the serialization issue in the backend instead and make a new mender-cli release.

frodeha: > Please correct me if I'm wrong, but to my understanding the server makes no difference between the different endpoint when it comes to serializing the meta_data. All endpoints that retrieves an Update type (whether it's /releases or /artifacts, v1 or v2) the server always returned a list of key/value pair which is not aligned with the API specs. It does have a different JSON serialization, but that is never used AFAICT since this data is coming from the artifact metadata. Which means that if there are any clients out there using the specs, they must have used the same workaround as you highlighted.

Correct.

> On the other hand, we have only received one report of the issue in mender-cli, so I'm starting to think whether we should fix the serialization issue in the backend instead and make a new mender-cli release.

If we want to do that then we can simply close this PR and progress with [this](https://github.com/mendersoftware/mender-cli/pull/310) instead. Fixing the serialization issue on main is what's causing the breakage in mender-cli. If we go this route we should probably update all the deployment specs to ensure that `meta_data` is an object (not a list) everywhere (e.g the inverse operation of https://github.com/mendersoftware/mender-server/commit/937ed3bce5358bea9b7bd67eecfe8138ce813c8f).

alfrunes: So are we actually following up on this TODO comment from 2017? :sweat_smile:
https://github.com/mendersoftware/deployments/blob/ea0239c8e4d7e939cf682aaec5e226beee5d98f9/resources/images/update.go#L46

frodeha: _"It's not a short term priority!"_ - 8 years later... 😂 

Jokes aside, yeah, looks that way!

alfrunes (approved):

kjaskiewiczz (approved):

PR #1056, opened 2025-10-29
fix(gui): fixed an issue that prevented closing details view of an ongoing delta generation

mzedel:


frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
```diff
@@ -253,6 +223,17 @@ export const DeltaGenerationDetailsDrawer = ({ jobId, onClose, open }: DeltaGene
       .finally(() => setIsLoading(false));
   }, [dispatch, jobId]);
 
+  useEffect(() => {
+    if (!jobId) {
+      return;
+    }
+    clearInterval(timer.current);
+    if (!['failed', 'success'].includes(deltaJob?.status)) {
```
aleksandrychev, line 231: if you have failed|success statuses somewhere in const it might worth using from that, but if used only once here it's ok to keeps as is

aleksandrychev (approved):

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
mzedel: it's defined in the type coming from the `deltaJob` definition - so any unsupported value will be causing errors 👌

mzedel: argh sorry, no I assumed wrong: ts doesn't catch this with the `.includes` check... hmm I could make this more explicit so: `if ('failed' !== deltaJob?.status && 'success' !== deltaJob?.status) {` - would put this in a follow up PR if you like 👌

mzedel: https://github.com/mendersoftware/mender-server/pull/1061 just in case...

PR #1079, opened 2025-11-04
docs(deployments): document 'sort' option in the internal endpoint for getting deployments

kjaskiewiczz:
So in the end the functionality was already there - the internal endpoint and management ones are reusing same logic quite a lot and support for sorting is present everywhere - internal, management v1 & v2. It was just not documented.
I tested locally and it works just fine.

Ticket: MEN-8914

### DIFF
--- a/backend/docs/api/deployments/internal_v1.yaml
+++ b/backend/docs/api/deployments/internal_v1.yaml
@@ -236,6 +236,15 @@ paths:
         name: created_after
         schema:
           type: integer
+      - description: |
+          Supports sorting the deployments list by creation date.
+        in: query
+        name: sort
+        schema:
+          enum:
+          - asc
+          - desc
+          type: string
       responses:
         '200':
           content:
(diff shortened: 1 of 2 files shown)


backend/docs/api/deployments/internal_v1.yaml
```diff
@@ -236,6 +236,15 @@ paths:
...
           type: integer
+      - description: |
+          Supports sorting the deployments list by creation date.
+        in: query
+        name: sort
+        schema:
+          enum:
+          - asc
+          - desc
+          type: string
```
alfrunes, line 247: I think you found a bug in the GH syntax highlighter for yaml. This does not render correctly :shrug:

merlin-northern (approved):

alfrunes (approved):

backend/docs/api/deployments/internal_v1.yaml
merlin-northern: yeah this is misleading as hell. long standing feature of the gh highlighter.

PR #1081, opened 2025-11-04
Introduce docker compose healthcheck

joelguittet:
The purpose of this Pull-Request is to introduce healthcheck mechanisms on mongo, nats and traefik containers.

The initial intention is to face some issues when running mender-server on a "slow" machine, with mongo regularly not being ready and multiple containers depending on mongo failing on startup.
I also increased the restart value because workflows container was failing. It depends on mongo, successfully connects but complains with "failed to apply migrations" error (version reported 0.0.0 while expecting 1.0.0).


docker-compose.yml
```diff
@@ -348,6 +365,11 @@ services:
 
   mongo:
     image: mongo:${MONGO_VERSION:-8.0}
+    healthcheck:
+      test: echo 'db.runCommand("ping").ok' | mongosh localhost:27017/test --quiet
```
alfrunes:
```suggestion
      test: 
      - mongosh
      - --quiet
      - --norc
      - --eval
      - 'db.runCommand("ping").ok && exit(0)'
```

PR #1084, opened 2025-11-06
Introduce docker compose healthcheck

joelguittet:
The purpose of this Pull-Request is to introduce healthcheck mechanisms on mongo, nats and traefik containers.

The initial intention is to face some issues when running mender-server on a "slow" machine, with mongo regularly not being ready and multiple containers depending on mongo failing on startup.
I also increased the restart value because workflows container was failing. It depends on mongo, successfully connects but complains with "failed to apply migrations" error (version reported 0.0.0 while expecting 1.0.0).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       context: .
       dockerfile: ./backend/services/create-artifact-worker/Dockerfile
     image: ${MENDER_IMAGE_REGISTRY:-docker.io}/${MENDER_IMAGE_REPOSITORY:-mendersoftware}/create-artifact-worker:${MENDER_IMAGE_TAG:-latest}
-    restart: on-failure:3
+    restart: on-failure:20
     depends_on:
       - workflows
     environment:
@@ -21,7 +21,7 @@ services:
       context: .
       dockerfile: ./backend/services/deployments/Dockerfile
     image: ${MENDER_IMAGE_REGISTRY:-docker.io}/${MENDER_IMAGE_REPOSITORY:-mendersoftware}/deployments:${MENDER_IMAGE_TAG:-latest}
-    restart: on-failure:3
+    restart: on-failure:20
     command: [server, --automigrate]
     depends_on:
       mongo:
@@ -63,7 +63,7 @@ services:
       context: .
       dockerfile: ./backend/services/deviceauth/Dockerfile
     image: ${MENDER_IMAGE_REGISTRY:-docker.io}/${MENDER_IMAGE_REPOSITORY:-mendersoftware}/deviceauth:${MENDER_IMAGE_TAG:-latest}
-    restart: on-failure:3
+    restart: on-failure:20
     command: [server, --automigrate]
     depends_on:
       mongo:
@@ -98,7 +98,7 @@ services:
       context: .
       dockerfile: ./backend/services/deviceconfig/Dockerfile
     image: ${MENDER_IMAGE_REGISTRY:-docker.io}/${MENDER_IMAGE_REPOSITORY:-mendersoftware}/deviceconfig:${MENDER_IMAGE_TAG:-latest}
-    restart: on-failure:3
+    restart: on-failure:20
     command: [server, --automigrate]
     environment:
       DEVICECONFIG_INVENTORY_URI: http://inventory:8080
@@ -127,7 +127,7 @@ services:
       context: .
       dockerfile: ./backend/services/deviceconnect/Dockerfile
     image: ${MENDER_IMAGE_REGISTRY:-docker.io}/${MENDER_IMAGE_REPOSITORY:-mendersoftware}/deviceconnect:${MENDER_IMAGE_TAG:-latest}
-    restart: on-failure:3
+    restart: on-failure:20
     command: [server, --automigrate]
     depends_on:
       mongo:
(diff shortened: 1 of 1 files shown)


joelguittet:
Should be ready for review now, as coverage doesn't make real sens here I suppose it's normal to fail.

merlin-northern:
> Should be ready for review now, as coverage doesn't make real sens here I suppose it's normal to fail.

Hey Joel. thank you for another lovely contribution. we need to discuss internally a bit, mainly because of the similarities to other places in the Mender ecosystem.

docker-compose.yml
```diff
@@ -348,6 +365,11 @@ services:
 
   mongo:
     image: mongo:${MONGO_VERSION:-8.0}
+    healthcheck:
+      test: echo 'db.runCommand("ping").ok' | mongosh localhost:27017/test --quiet
```
alfrunes: Can invoke `mongosh` directly.
```suggestion
      test: ["CMD", "mongosh", "mongodb://localhost:27017/test", "--quiet", "--eval",  'db.runCommand("ping").ok']
```

docker-compose.yml
```diff
@@ -359,8 +381,13 @@ services:
...
   nats:
-    image: nats:2.12
-    command: [-js]
+    image: nats:2.12-alpine
+    command: [-js, -m, '8222']
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:8222/healthz | grep -q 'ok'"]
+      interval: 10s
+      timeout: 3s
+      retries: 5
```
alfrunes, line 390: I would prefer keeping the NATS scratch image as it is the image we advice using in production, in addition to being significantly smaller.

docker-compose.yml
```diff
@@ -330,6 +342,11 @@ services:
       - --providers.file.directory=/etc/traefik/config
       - --providers.docker=true
       - --providers.docker.exposedByDefault=false
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:8080/ping | grep -q 'OK'"]
```
alfrunes:
```suggestion
      test: ["CMD", "wget", "-q", "http://localhost:8080/ping"]
```

alfrunes (commented):
Generally I would advice using the [helm chart](https://github.com/mendersoftware/mender-helm/) instead of the compose setup for any installations.
There are health and liveliness endpoints in the mender-server containers too, but they're all built from scratch so their not so useful in a compose environment. Without these there will still be a startup race condition as the traefik healthcheck doesn't really check the liveliness of it's routes.

docker-compose.yml
joelguittet: @alfrunes no we need to check for the OK in the output. for that grep and CMD-SHELL.

docker-compose.yml
joelguittet: Will try

docker-compose.yml
joelguittet: Not possible, the scratch image doesn't contains wget/grep and so it's not possible to implement the healthcheck. We need to move on the alpine image.

alfrunes: Sorry, that was what I was trying to say. I don't think this heathcheck is strictly necessary.

docker-compose.yml
alfrunes: According to [the docs](https://doc.traefik.io/traefik/v3.4/operations/ping/#configuration-options), the ping endpoint always returns status code 200 with a body of OK. `wget` will fail with non-zero exit code on client/server and network errors. But maybe better to swap `wget` with the `traefik healthcheck` CLI command?

docker-compose.yml
joelguittet: Working, applied

docker-compose.yml
joelguittet: It is mandatory and to be honest it is the main purpose of this PR, else we can just close it.
I should introduce you a bit of context. I'm using mender-server in a CI workflow to test updates etc. I setup a mender-server in the CI, and shutdown at the end, so that I don't impact any development or production workflow, and I'm sure to start testing with a fresh environment.
This is done on a self-hosted runner on a "small" machine. I got numbered of issues when starting the containers with lots of failure difficult to understand because of the "slow" CPU.
This has been solved by adding the healthcheck so that mender containers starts properly. Else, mender containers starts en generally fails, just because nats and or mongo containers are not ready.

docker-compose.yml
joelguittet: Good point on the doc. I tried with the traefik healthcheck command too, but didn't find the right one, not sure why. Trying now without the grep.

joelguittet: Need to revert back, it's showed as unhealthy with your proposal :(

joelguittet:
@alfrunes ok so applied some of your suggestions, but i face sometimes issues with traefik being unhealthy, so i think it's better to keep as is. For nats, alpine is definitively required to have wget/grep.

docker-compose.yml
alfrunes: From your description I was assuming that mongodb is the problem, so I thought that the nats healthcheck wasn't strictly necessary. If that's not the case, let's move it to alpine.

docker-compose.yml
alfrunes: Ahh, it seems like the healthcheck command depends on the static configuration (CLI flags) in order to discover the proxy process. Let's keep it as is for now, until the static configuration is migrated to a configFile.

alfrunes (approved):
Thank you for contributing! :heart: 
I have nothing more to add.

joelguittet:
@alfrunes thanks for merging.
I have the same for the 4.0.x branch: https://github.com/mendersoftware/mender-server/pull/1085

PR #1086, opened 2025-11-06
Update rate limit configuration for consistency and tiers

alfrunes:
This PR aligns the rate limit configuration to use `rate_limits` - consistent with deployments service and correct spelling :wink:

This PR specifically does:
* Backport ratelimit configuration validation from enterprise
* Refactor authenticated ratelimit defaults to `useradm` and `deviceauth` service
* Rename root rate limit configuration to `rate_limit` for consistency.
* Update default `rate_limits` for deviceauth service to account for tiers (and lower default rule to 30 requests per minute for devices).


backend/services/deviceauth/config.yaml
```diff
@@ -171,6 +171,14 @@
...
 #         # .Trial    (jwt:"mender.trial")  bool
 #         event_expression: "{{with .Identity}}{{.Subject}}{{end}}"
+#       - name: checkUpdate
+#         interval: 5m
+#         quota: 1
+#         event_expression: "{{with .Identity}}{{.Subject}}{{end}}"
+#       - name: checkUpdate/micro
+#         interval: 24h
+#         quota: 1
+#         event_expression: "{{with .Identity}}{{.Subject}}{{end}}"
```
frodeha: We have thus far operated as if "tiers" isn't a thing in open source. Should we maybe only introduce the micro tier limits in enterprise to stay with this?

merlin-northern: actually good catch because the identity will not contain tier in builds originating from this repo.

alfrunes: Right, should I remove tiers from open source then or should I add tiers to identity? :juggling_person:

alfrunes: I went for the latter, but I can also change this commit and amend the change in enterprise...

frodeha: From my perspective I would prefer the former. Reason being that I'd like to avoid having to (needlessly) do all future tier dependent rate limit implementations in open source before merging them to enterprise (the place they are required). 

Further to that, tiers as a concept doesn't exist in open source (as we have chosen to implement them), so having to go through the open source -> enterprise merge because of them seems like wasted effort to me.

alfrunes: Makes sense. I updated the PR, removing the additional rules for tiers.

frodeha (dismissed):
LGTM

alfrunes:
Sorry, I had to align the deviceauth change with the commit message (changelog). I increased it to 30 requests per device per minute by default accounting for the sum of polling intervals from clients.

frodeha (approved):

kjaskiewiczz (approved):

PR #1087, opened 2025-11-06
ci: consolidate coveralls jobs

alfrunes:
This commit consolidates all coveralls jobs to use the coveralls CLI and letting the CLI discover the parameters from GitlabCI to make sure the parameters are used consistently.

~~NOTE: I'm not sure if this work for the parallel builds for the backend the job reports multiple coverage reports in a single job.~~
Edit: it works ✨


alfrunes:
The last commit is just a trivial change to trigger the full coverage job paths.

mzedel (approved):

PR #1088, opened 2025-11-06
(doc) Add documentation that `retries` is restricted to professional and enterprise plans

nickanderson:
This change attempts to properly document that `retries` requires the professional or enterprise plan.


mzedel (commented):
Change looks good, but the conventional commit format wants to have the commit message like `docs: Add documentation that ..` 👌

mzedel (approved):

alfrunes (approved):

PR #1089, opened 2025-11-10
fix(gui): increased reliance on theme defined button colors to align more w/ updated theme

mzedel:
that should bring https://github.com/NorthernTechHQ/nt-gui/pull/405 in and ease spotting differences between next theme design & current implementation


chiachenglu (dismissed):
Looks good to me 👌

aleksandrychev (approved):

PR #1096, opened 2025-11-18
fix(gui): fixed pagination in device search results

mineralsfree:
Ticket: MEN-8196


mzedel (approved):
Nice, here I expected more to be done 😅

PR #1102, opened 2025-11-20
test(gui): ensured unlimited enterprise tenants work until tier related limit handling lands

mzedel:


frodeha (approved):
That should work. LGTM 👍

PR #1104, opened 2025-11-21
MEN-8935: Add request content length to access logs

frodeha:
**Description**
This PR adds the value of the Content-Length header of each request to the access log. I think (hope) that this is actually all that's required to meet the acceptance criteria. However, please do see the ticket for further details and let me know if I'm wrong in that.

Example:
```
time="2025-11-21T10:30:56Z" level=info byteswritten=-1 caller="accesslog.AccessLogger.LogFunc@middleware_gin.go:134" method=POST path=/api/management/v1/deployments/artifacts path_params= qs= request_id=e6a47e52-fbe4-44f2-91b8-518d1e240cd7 responsetime=14.43ms contentlength=5581 status=201 ts="2025-11-21T10:30:56.882Z" type=HTTP/1.1 user_id=<> useragent="<>"
```

### DIFF
--- a/backend/pkg/accesslog/middleware_gin.go
+++ b/backend/pkg/accesslog/middleware_gin.go
@@ -107,6 +107,9 @@ func (a AccessLogger) LogFunc(
 	logCtx["responsetime"] = latency.String()
 	logCtx["status"] = c.Writer.Status()
 	logCtx["byteswritten"] = c.Writer.Size()
+	if length := c.Request.ContentLength; length > -1 {
+		logCtx["contentlength"] = length
+	}
 
 	var logLevel logrus.Level = logrus.InfoLevel
 	if code >= 500 {


backend/pkg/accesslog/middleware_gin.go
```diff
@@ -107,6 +107,7 @@ func (a AccessLogger) LogFunc(
 	logCtx["responsetime"] = latency.String()
 	logCtx["status"] = c.Writer.Status()
 	logCtx["byteswritten"] = c.Writer.Size()
+	logCtx["size"] = c.Request.ContentLength
```
merlin-northern: I would mark that it comes from ContentLength perhaps `logCtx["length"]` or something like that? ContentLength is not always set and not always required by the HTTP RFC so do you think we should add this log field always? or skip when not present?

frodeha: > I would mark that it comes from ContentLength perhaps logCtx["length"] or something like that? 

Fair point. I went off the acceptance criteria too literally. I'd go with "contentlength" then - same as we have "responsetime" and "byteswritten" already if that sounds okay to you?

> ContentLength is not always set and not always required by the HTTP RFC so do you think we should add this log field always? or skip when not present?

Hmm, `c.Request.ContentLength` will be `-1` when there's no header set (same as `c.Writer.Size()`) so in that way it's at least consistent. On the other hand, it could interfere with the log aggregation we're planning to do based on this field (ref the ticket) so maybe it's better to not include it, yeah...

Edit: Something like 137da33ab3056aa80eca0dc4226dddd24b84e7bc?

alfrunes (approved):

PR #1107, opened 2025-11-21
fix(deviceconnect): device ping/pong handler causes timeouts

alfrunes:
When there's traffic on the websocket connection, the ping/pong handler is never called and therefore SetReadDeadline is never called which in turn causes the connection to terminate.
This PR removes the calls to Conn.SetReadDeadline to rely on TCP keep-alive instead.

### DIFF
--- a/backend/services/deviceconnect/api/http/device.go
+++ b/backend/services/deviceconnect/api/http/device.go
@@ -241,15 +241,10 @@ func (h DeviceController) connectWSWriter(
 	defer ticker.Stop()
 	conn.SetPongHandler(func(string) error {
 		ticker.Reset(pingPeriod)
-		return conn.SetReadDeadline(time.Now().Add(pongWait))
+		return nil
 	})
 
 	conn.SetPingHandler(func(msg string) error {
-		err := conn.SetReadDeadline(time.Now().Add(pongWait))
-		if err != nil {
-			return err
-		}
-
 		ticker.Reset(pingPeriod)
 		return conn.WriteControl(
 			websocket.PongMessage,


kjaskiewiczz:
It makes sense, but I'm not sure if it works. Did you test it with idle connection?
When testing locally, I've found that lack of it is causing even more issues, that's why I reintroduced it here - https://github.com/mendersoftware/mender-server/pull/1028
And it's present in `saas.2` https://github.com/mendersoftware/mender-server/blob/e6e19498bda0678296612ec0f6f87994194502a0/backend/services/deviceconnect/api/http/device.go#L252 version which we claim works best and version we are rolling back to for a long time now.

alfrunes:
> It makes sense, but I'm not sure if it works. Did you test it with idle connection? When testing locally, I've found that lack of it is causing even more issues, that's why I reintroduced it here - #1028 And it's present in `saas.2`
> 
> https://github.com/mendersoftware/mender-server/blob/e6e19498bda0678296612ec0f6f87994194502a0/backend/services/deviceconnect/api/http/device.go#L252
> version which we claim works best and version we are rolling back to for a long time now.

The main issue I'm trying to address is that the I/O deadline is only reset on ping/pong. However, whenever data is received form the device the timer to send a new ping/pong is reset. If the client implements the same logic, then whenever there is traffic coming from the device, this will prevent the ping/pong handler to run and therefore setup the connection for a timeout. An alternative would be to add a `SetReadDeadline` for every package received, but I suspect that will make the connections more sensitive to load.

kjaskiewiczz:
@alfrunes I understand the change and, as I wrote in the first place, it makes sense.
What I'm trying to say is that I was testing it with very similar change (without ReadDeadline in ping handler, but with ReadDeadling in pong handler), and it was not working properly for me. If I remember correctly it was not working properly on production too.
When I was re-introducing ReadDeadline in the ping handler I even mentioned that it doesn't really make sense to me, but it's helping with keeping the connection stable :shrug:
We deployed it to production and it was running fine for some time. 
If you tested it locally and it's working fine, then we can give it a try, but we have to monitor deviceconnect errors after deployment and we also need be ready for rollback (which I think we are anyway).

kjaskiewiczz (approved):
As mentioned in the comments I'm not sure if it won't cause issues on production, so be careful when releasing to hosted.

alfrunes:
> As mentioned in the comments I'm not sure if it won't cause issues on production, so be careful when releasing to hosted.

There is one difference, and that is [the ticker](https://github.com/mendersoftware/mender-server/blob/9b318f82587bab341e15b96a2843d61012505881/backend/services/deviceconnect/api/http/device.go#L264) is not reset when receiving messages. Which means that (currently) we don't send ping/pong if we receive data from the device which sets up the connection for a timeout. I believe this could happen for instance every time you start a single device deployment or just actively use the remote terminal. I don't think it's a good idea to use the deadlines in the first place as it only takes a few seconds of CPU throttling to make the connections time out.

PR #1110, opened 2025-11-25
Ci/firefox downloads

mzedel:
successful download test on firefox against staging is [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/12207658978) (unrelated failure due to rate limits)


mineralsfree (approved):
LGTM

PR #1111, opened 2025-11-25
ci: added request cooldown on e2e page init to try to avoid rate limits

mzedel:


mzedel:
closing as this doesn't seem to be enough: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/12207658978/artifacts/file/frontend/tests/e2e_tests/test-results/01-releases-Files-allows-release-tags-filtering-baseline-firefox/video.webm still fails due to the per minute rate limits

PR #1118, opened 2025-11-27
fix(gui): Update URL from browser's console welcome message

MuchoLucho:
Changelog: None
Ticket: None

Probably one of my silliest commits


mzedel (approved):
Oh my, thanks for catching this... I've seen it every day for months now and never noticed

PR #1140, opened 2025-12-02
test(gui): let tests properly wait for rbac restricted elements to appear

mzedel:
- prior the rbac info might not have been resolved in time to show the expected ui parts

FYI @alfrunes


alfrunes (approved):
Thanks!

PR #1148, opened 2025-12-03
Fix/delta things

mzedel:
depends on https://github.com/mendersoftware/mender-server-enterprise/pull/819

### DIFF
--- a/frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
+++ b/frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
@@ -108,7 +108,14 @@ const statusColumns = [
     key: 'totalTime',
     title: 'Total time',
     cellProps: { style: { width: '10%' } },
-    render: ({ totalTime }) => totalTime
+    render: ({ total_time }) => {
+      if (!total_time) {
+        return '-';
+      }
+      const duration = dayjs.duration(`PT${total_time}`.toUpperCase());
+      const format = duration.asSeconds() >= 60 ? 'HH:mm' : 'HH:mm:ss'; // allowing quick generations to also show something
+      return duration.format(format);
+    }
   },
   {
     key: 'toArtifactSize',
@@ -142,44 +149,6 @@ const statusColumns = [
     render: () => ''
   }
 ];
-
-const DELTA_GENERATION_TIMEOUT_MINUTES = 60;
-
-const getTotalTime = (started?: string, finished?: string): string => {
-  if (!started) {
-    return '-';
-  }
-  const startTime = dayjs(started);
-  if (!finished) {
-    const duration = dayjs.duration(startTime.diff(dayjs()));
-    if (duration.asMinutes() > DELTA_GENERATION_TIMEOUT_MINUTES) {
-      return `${DELTA_GENERATION_TIMEOUT_MINUTES}:00`;
-    }
-  }
-  const endTime = dayjs(finished);
-  const duration = dayjs.duration(startTime.diff(endTime));
-  if (duration.minutes() !== Math.abs(duration.minutes())) {
-    // negative time calculated => something's off
-    return '-';
-  }
-  return duration.format('HH:mm');
-};
-
-// Look for completion patterns in the log
-const finishingPattern = /(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}).*?(:?completed|finished|done)/i;
-const getFinishedTimeFromLog = (log?: string): string | undefined => {
-  if (!log) {
-    return;
-  }
-  const lines = log.split('/n');
-  for (const line of lines) {
-    const match = line.match(finishingPattern);
-    if (match) {
-      return match[1];
-    }
-  }
-  return;
-};
 
 type EnhancedJobDetailsItem = DeltaJobDetailsItem &
   DeltaJobsListItem & {
@@ -242,17 +211,13 @@ export const DeltaGenerationDetailsDrawer = ({ jobId, onClose, open }: DeltaGene
     if (!deltaJob) {
       return;
     }
-    const { details, started, target_size, delta_artifact_size, to_release, to_version, from_release, from_version } = deltaJob;
-    const finished = getFinishedTimeFromLog(details);
-    const totalTime = getTotalTime(started, finished);
+    const { target_size, delta_artifact_size, to_release, to_version, from_release, from_version } = deltaJob;
     const dataSaved = target_size && delta_artifact_size ? Math.max(0, target_size - delta_artifact_size) : 0;
 
     return {
       ...deltaJob,
       toRelease: to_release || to_version || '-',
       fromRelease: from_release || from_version || '-',
-      finished,
-      totalTime,
       dataSaved
     };
   }, [deltaJob]);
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
```diff
@@ -242,17 +204,13 @@ export const DeltaGenerationDetailsDrawer = ({ jobId, onClose, open }: DeltaGene
     if (!deltaJob) {
       return;
     }
-    const { details, started, target_size, delta_artifact_size, to_release, to_version, from_release, from_version } = deltaJob;
```
alfrunes, line 245: FYI, there's no difference between to/from_release and to/from_version.

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
```diff
@@ -299,7 +257,7 @@ export const DeltaGenerationDetailsDrawer = ({ jobId, onClose, open }: DeltaGene
             </div>
             <LinedHeader className="margin-top-large" heading="Status" />
             <DetailsTable className={classes.table} columns={statusColumns} items={[combinedData]} />
-            {combinedData.status === 'failed' && combinedData.log && <Code className="log">{combinedData.log}</Code>}
```
alfrunes, line 302: Interesting. AFAIK there never was any serialized attribute with the name `log`, it's always been `details` but there is an OAS example with an attribute log :thinking:

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
```diff
@@ -143,44 +143,6 @@ const statusColumns = [
...
-  if (duration.minutes() !== Math.abs(duration.minutes())) {
-    // negative time calculated => something's off
-    return '-';
-  }
-  return duration.format('HH:mm');
-};
-
-// Look for completion patterns in the log
-const finishingPattern = /(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}).*?(:?completed|finished|done)/i;
-const getFinishedTimeFromLog = (log?: string): string | undefined => {
```
alfrunes, line 170: Yikes :face_with_peeking_eye:

alfrunes (commented):
Interesting read. I made some adjustments to the specs in the other PR. Regarding the duration parsing, does `dayjs` handle Go formatted durations or should we change it to number of seconds/milliseconds/nanoseconds? Or would you rather just aggregate the duration from the start to finished diff?

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
mzedel: will both always be set? one is used as the fallback to the other a few lines down as they so far have existed only in one of the models...

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
mzedel: https://github.com/mendersoftware/mender-server-enterprise/pull/718 + 
https://github.com/mendersoftware/mender-server-enterprise/pull/746 + https://github.com/mendersoftware/mender-server-enterprise/pull/748 brought us here 😅 - so it did exist in the swagger specs, then also OAS, then it (+ other parts) didn't exist in the actual response, then it got aligned and now... it looks like the spec example is left to adjust after this 🙃

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
mzedel: what you do ~for love~ to show data in the UI 🌈

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
alfrunes: Hmm.. In my latest changes I make a literally copy of the value that is always stored (when the object is first created), so if those changes get an :heavy_check_mark: it's safe to say they'll always be the same.

mzedel:
> does `dayjs` handle Go formatted durations or should we change it to number of seconds/milliseconds/nanoseconds? Or would you rather just aggregate the duration from the start to finished diff?

@alfrunes that's what I meant with https://github.com/mendersoftware/mender-server-enterprise/pull/819#discussion_r2584533699 - I think the potential for a progress update via the `total_time` before it is actually finished would be neat, but based on the [regex](https://github.com/iamkun/dayjs/blob/dev/src/plugin/duration/index.js#L13) used in dayjs to parse string inputs it seems like we would run into issues for any very quick generation...
using the example durations from the go docs:
```js
const duration1 = '1h2m0.3s';
const duration2 = '300ms';
console.log(dayjs.duration(`PT${duration1}`.toUpperCase()).format('HH:mm')
// '01:02'
console.log(dayjs.duration(`PT${duration2}`.toUpperCase()).format('HH:mm')
// '00:300'
```
and even if we add seconds for the second duration
```
console.log(dayjs.duration(`PT${duration2}`.toUpperCase()).format('HH:mm:ss')
// '00:300:00'
```
the misunderstanding persists 😬 ...

alfrunes:
> > does `dayjs` handle Go formatted durations or should we change it to number of seconds/milliseconds/nanoseconds? Or would you rather just aggregate the duration from the start to finished diff?
> 
> @alfrunes that's what I meant with [mendersoftware/mender-server-enterprise#819 (comment)](https://github.com/mendersoftware/mender-server-enterprise/pull/819#discussion_r2584533699) - I think the potential for a progress update via the `total_time` before it is actually finished would be neat, but based on the [regex](https://github.com/iamkun/dayjs/blob/dev/src/plugin/duration/index.js#L13) used in dayjs to parse string inputs it seems like we would run into issues for any very quick generation... using the example durations from the go docs:
> 
> ```js
> const duration1 = '1h2m0.3s';
> const duration2 = '300ms';
> console.log(dayjs.duration(`PT${duration1}`.toUpperCase()).format('HH:mm')
> // '01:02'
> console.log(dayjs.duration(`PT${duration2}`.toUpperCase()).format('HH:mm')
> // '00:300'
> ```
> 
> and even if we add seconds for the second duration
> 
> ```
> console.log(dayjs.duration(`PT${duration2}`.toUpperCase()).format('HH:mm:ss')
> // '00:300:00'
> ```
> 
> the misunderstanding persists 😬 ...

Ahh, ok. I added rounding to nearest second here: https://github.com/mendersoftware/mender-server-enterprise/pull/819/commits/4486f5f0a268e8caee34fe3cb18c5c6d4d75b3f9 since anything under half a second is really not that interesting...
Would that help? Then there won't be any fractions, milliseconds, microseconds or nanoseconds.

mzedel:
> Ahh, ok. I added rounding to nearest second here: [mendersoftware/mender-server-enterprise@4486f5f](https://github.com/mendersoftware/mender-server-enterprise/commit/4486f5f0a268e8caee34fe3cb18c5c6d4d75b3f9) since anything under half a second is really not that interesting... Would that help? Then there won't be any fractions, milliseconds, microseconds or nanoseconds.

🙌 yep that should work great! Thank you!

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
```diff
@@ -108,7 +108,14 @@ const statusColumns = [
     key: 'totalTime',
     title: 'Total time',
     cellProps: { style: { width: '10%' } },
-    render: ({ totalTime }) => totalTime
+    render: ({ total_time }) => {
+      if (!total_time) {
+        return '-';
+      }
+      const totalTime = total_time / 1000 / 1000; // API value is in nanoseconds, dayjs needs milliseconds
```
alfrunes: FYI, I corrected the total time to return a formatted duration string.

frontend/src/js/components/releases/DeltaGenerationDetailsDrawer.tsx
alfrunes: :exploding_head:

alfrunes (approved):

mineralsfree (approved):
LGTM

Issue #1095, opened 2025-11-17
[DevAuth] Incorrect/Insufficient Error Handling in datastore_mongo.AddDevice


OlliMartin:
Hi, 

we are trying to run Mender with a Azure-managed Mongo database.
We see some issues, most importantly the device auth service returning 500 for every device auth request after the first one. 
(The first one correctly returns 401, the device is not yet accepted)

Trying to find a solution, we've looked into the code and it seems the [error handling](https://github.com/mendersoftware/mender-server/blob/132f68c4a545d91664bf54c495e3ff1114705ee5/backend/services/deviceauth/store/mongo/datastore_mongo.go#L327C1-L332C3) in `datastore_mongo` is insufficient/incorrect:

``` go
	if _, err := c.InsertOne(ctx, d); err != nil {
		if strings.Contains(err.Error(), "duplicate key error") {
			return store.ErrObjectExists
		}
		return errors.Wrap(err, "failed to store device")
	}
```

The `strings.Contains([..])` call is valid for traditional mongo, but does not work for azure, because there a different error message is returned:

<img width="1606" height="215" alt="Image" src="https://github.com/user-attachments/assets/01ca3322-1060-4d77-a5cb-4fc82cdd82ec" />

In contrast, self-hosted mongo returns:

```"E11000 duplicate key error collection: inventory.devices index"```

which explains why it works there (the `strings.Contains` returns true).

All other places, at least in this file, check for record duplication with 

``` go
        if mongo.IsDuplicateKeyError(err) {
            return store.ErrObjectExists
        }
```

I do not see an obvious reason why the special error handling would be required here and _our_ problem should be solved if `mongo.IsDuplicatedKeyError` is used to check if the authset already exists, because Azure returns code 11000 for index duplicates as expected.

I would be happy to raise a PR to use `mongo.IsDuplicateKeyError` instead of `strings.Contains(err.Error(), "duplicate key error")` if the approach is fine. :)

Best regards
Olli

**As a side note:**
We were wondering why there is a tenant related index in the first place, since we are using the community edition from the helm chart. It seems when cleaning up the tenant related code in #904, the migrations were missed (or left there on purpose?). 
Since Azure does not support sparse indexes (our tenant_id will always be empty string) we probably need to drop said index and recreate one without tenant_id ? Is there a plan to clean up this index? 
[...]

PR #1152, opened 2025-12-03
fix(deviceauth): Use mongo's error code to check for record duplication in AddDevice

OlliMartin:
As outlined in #1095, the error handling in deviceauth datastore_mongo uses a string comparison to check if a record is already present or not. This creates problems if the underlying database does not respond _exactly_ with the expected error _message_.

This PR updates the code to use an existing, more robust function instead, which checks on the error code returned by the mongo server.

__Note:__ The same string comparison happens in the useradm module too, but I was not sure if (and how) it should be included in the change, since there no duplicate handling seems to be required/implemented yet:

``` go
var (
	// user not found
	ErrUserNotFound = errors.New("user not found")
	// token not found
	ErrTokenNotFound = errors.New("token not found")
	// duplicated email address
	ErrDuplicateEmail = errors.New("user with a given email already exists")
	// duplicated Personal Access Token name
	ErrDuplicateTokenName = errors.New("Personal Access Token with a given name already exists")
	// etag doesn't match
	ErrETagMismatch = errors.New("ETag doesn't match")
)
```

### DIFF
--- a/backend/services/deviceauth/store/mongo/datastore_mongo.go
+++ b/backend/services/deviceauth/store/mongo/datastore_mongo.go
@@ -325,7 +325,7 @@ func (db *DataStoreMongo) AddDevice(ctx context.Context, d model.Device) error {
 	c := db.client.Database(DbName).Collection(DbDevicesColl)
 
 	if _, err := c.InsertOne(ctx, d); err != nil {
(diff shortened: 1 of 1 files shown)


alfrunes (approved):
Thank you! :pray:

PR #1155, opened 2025-12-04
MEN-9071: Prevent race condition re-caching invalid value

bahaa-ghazal:


backend/pkg/redis/redis.go
```diff
@@ -17,17 +17,28 @@ package redis
...
 	"strings"
+	"time"
 
 	"github.com/redis/go-redis/v9"
 )
 
 type Client = redis.Cmdable
 
+const (
+	TemporaryValue           = "TEMP_VALUE"
```
alfrunes:
```suggestion
	CacheInvalidValue           = "TEMP_VALUE"
```
Would it be an option to reserve the empty string as an invalidated cache entry?
```suggestion
	CacheInvalidValue           = ""
```

backend/pkg/redis/redis.go
```diff
@@ -17,17 +17,28 @@ package redis
...
+	TemporaryValueExpiration = time.Second * 10 // 10s
+)
+
+var (
+	ErrTemporaryCache = errors.New("temporary cache entry")
```
alfrunes: I think the naming here is a bit arbitrary.
```suggestion
	ErrCacheInvalid = errors.New("cache invalidated")
```

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/mendersoftware/mender-server/pkg/log"
 	"github.com/mendersoftware/mender-server/pkg/mongo/oid"
 	"github.com/mendersoftware/mender-server/pkg/plan"
+	mredis "github.com/mendersoftware/mender-server/pkg/redis"
```
alfrunes, line 31: Adding redis as a direct dependency here defeats some of the purpose of having the cache abstraction in the first place. OTOH, I don't think we will use a different cache (at least not in the short term), but this is tying the dependency even harder into the product.
At this point, I think this will add too much complexity to the task, so let's keep it like this and then we can refactor later as needed.

backend/pkg/redis/redis.go
```diff
@@ -155,3 +166,44 @@ func ClientFromConnectionString(
 		Result()
 	return rdb, err
 }
+
+// Invalidates a cache entry by swapping it with a temporary short-lived entry.
+// When a function receives this temporary value, it should count it as a cache miss,
+// but also stop the caching of the value fetched DB value.
+func SafeCacheDel(ctx context.Context, c Client, key string) error {
```
alfrunes: Invalidating and deleting are two separate operations.
```suggestion
func InvalidateCache(ctx context.Context, c Client, key string) error {
```

backend/pkg/redis/redis.go
```diff
@@ -155,3 +166,44 @@ func ClientFromConnectionString(
...
+	if cmd.Val() == TemporaryValue {
+		cmd.SetErr(ErrTemporaryCache)
+	}
+
+	return cmd
+}
+
+func IsInvalidCacheEntry(cacheEntry string) bool {
+	return cacheEntry == "" || cacheEntry == TemporaryValue
+}
```
alfrunes: Here, both an empty string and `TEMP_VALUE` are invalid value, wouldn't it suffice to only reserve the empty string? (ref. previous comment).

backend/services/deviceauth/devauth/devauth.go
alfrunes: Just a side note; why the `mredis` import name? Why not simply `redis`?

backend/pkg/redis/redis.go
alfrunes: Just noticed that the empty string is an invalidated cache entry. Can we change this constant too?

bahaa-ghazal: The reason why we can't use empty string is because we need to differentiate between cache miss and this temprory value. Here are the 2 scenarios :
1. We get a empty string (cache miss): we fetches the data from db then we **cache** the results .
2. We get  the temp value:  we fetches the data from db then we **dont cache** the results.

backend/services/deviceauth/devauth/devauth.go
bahaa-ghazal: `redis` ->  `github.com/redis/go-redis/v9`
`mredis` (mender redis) -> `/mender-server/pkg/redis`
This is the convention that was used previously.

backend/pkg/redis/redis.go
bahaa-ghazal: we count `TEMP_VALUE` and `""` as cache miss. This function check if we have a cache miss. the name could use some refactoring.

alfrunes: But why two values? I believe the empty string is sufficient.

backend/pkg/redis/redis.go
alfrunes: Won't a cache miss return a `redis.Nil` error?

bahaa-ghazal: You are totally correct. Still little bit rusty after coming back 🙃

backend/pkg/redis/redis.go
```diff
@@ -155,3 +166,44 @@ func ClientFromConnectionString(
 		Result()
 	return rdb, err
 }
+
+// Invalidates a cache entry by swapping it with a temporary short-lived entry.
```
alfrunes:
```suggestion
// Invalidates a cache entry by swapping it with a temporary short-lived empty value.
```

alfrunes (dismissed):
LGTM

alfrunes (dismissed):

alfrunes (approved):

PR #1157, opened 2025-12-04
MEN-8923: Fix delete releases of finished deployments

frodeha:
**Description**

This PR fixes the reported bug where you can't delete a release even if all deployments that used it are finished. This is done by only considering active deployments when we check if the artifact name is in use.


backend/tests/integration/tests/test_deployments.py
```diff
@@ -2636,3 +2636,58 @@ def test_list_artifacts_v2(self, clean_mongo):
...
+            deployments_v2.URL_RELEASES + "?name=" + "deployments-phase-testing",
+        )
+        assert r.status_code == 409
+
+        # Perform the upgrade to mark the deployment as finished
+        status_code = try_update(devs[0], default_device_type="foo")
+        assert status_code == 204
+
+        dep = get_deployment(depid, user_token)
+        print(dep)
```
alfrunes: A little debug print :monocle_face: 
I don't mind :upside_down_face:

alfrunes (dismissed):
Short and sweet. Just as I like it :)

alfrunes (approved):
:zap:

backend/tests/integration/tests/test_deployments.py
frodeha: Man, that's annoying me so I had to fix it! 😂 
The call to `get_deployment` isn't needed either. That was also debug stuff that I left by mistake. Thanks for pointing it out 🙏

PR #1160, opened 2025-12-04
feat(workflows): support for accessing data from objects via json

merlin-northern:
With this change the input parameters that are objects with json representation can be referenced via `identifier.json.field` syntax. For instance: in a device is passed as an input parameter, then anywhere you use `device.json.tier` you will refer to the Device.Tier field, since it is decoreated with json `tier`.

Changelog: Commit
Ticket: MEN-9096

### DIFF
--- a/backend/services/workflows/app/processor/job.go
+++ b/backend/services/workflows/app/processor/job.go
@@ -17,6 +17,8 @@ package processor
 import (
 	"strings"
 
+	"github.com/thedevsaddam/gojsonq"
+
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 
@@ -36,6 +38,7 @@ func NewJobProcessor(job *model.Job) *JobProcessor {
 	}
 }
 
+//nolint:gocyclo
 func (j JobProcessor) ProcessJSON(
 	data interface{},
 	ps *JobStringProcessor,
@@ -61,6 +64,18 @@ func (j JobProcessor) ProcessJSON(
 				len(key) > len(workflowInputVariable) {
 				key = key[len(workflowInputVariable):]
 				for _, param := range j.job.InputParameters {
+					// support for key === "device.jsonInput.tier"; accessing the input params
+					// objects by JSON identifiers. for instance, if a whole device is passed
+					// as an input parameter, then with `device.jsonInput.tier` you can refer
+					// to the Device.Tier field
+					if strings.HasPrefix(key, param.Name+jsonRefInputVariable) {
+						return gojsonq.New().FromString(param.Value).Find(
+							strings.TrimPrefix(
+								key,
+								param.Name+jsonRefInputVariable,
+							),
+						)
+					}
 					if param.Name == key && param.Raw != nil {
 						return j.ProcessJSON(param.Raw, ps)
 					}
--- a/backend/services/workflows/app/processor/string.go
+++ b/backend/services/workflows/app/processor/string.go
@@ -15,6 +15,7 @@
 package processor
 
 import (
+	"encoding/json"
 	"html"
 	"net/url"
 	"os"
@@ -30,6 +31,7 @@ import (
 const (
 	workflowEnvVariable   = "env."
 	workflowInputVariable = "workflow.input."
+	jsonRefInputVariable  = ".jsonInput"
 	regexVariable         = `\$\{(?P<options>(?:(?:[a-zA-Z]+)=(?:[a-zA-Z0-9]+);)*)` +
 		`(?P<name>[^;\}\|]+)(?:\|(?P<default>[^\}]+))?}`
 	regexOutputVariable = `(.*)\.json\.(.*)`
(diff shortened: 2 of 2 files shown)


backend/services/workflows/app/processor/job.go
```diff
@@ -61,6 +63,18 @@ func (j JobProcessor) ProcessJSON(
...
+					// as an input parameter then with `device.json.tier` you can refer
+					// to the Device.Tier field
+					if strings.HasPrefix(key, param.Name+".json") {
+						return gojsonq.New().FromString(param.Value).Find(
+							strings.TrimPrefix(
+								key,
+								param.Name+".json.",
+							),
+						)
+					}
```
alfrunes: Shouldn't this be in the `ProcessJobString` interface so it applies uniformly to all kinds of strings?

backend/services/workflows/app/processor/job.go
```diff
@@ -61,6 +63,18 @@ func (j JobProcessor) ProcessJSON(
 				len(key) > len(workflowInputVariable) {
 				key = key[len(workflowInputVariable):]
 				for _, param := range j.job.InputParameters {
+					// support for key === "device.json.tier"; accessing the input params
+					// objects by json identifiers. for instance if a whole device is passed
+					// as an input parameter then with `device.json.tier` you can refer
+					// to the Device.Tier field
+					if strings.HasPrefix(key, param.Name+".json") {
```
alfrunes: I believe this format conflicts with the format for output variables
https://github.com/mendersoftware/mender-server/blob/e357e4983eaa04be17fb944b34286911234aa597/backend/services/workflows/app/processor/string.go#L35

Maybe something like `workflow.jsonInput.<inputName>.<jsonpath>` could work?

alfrunes (changes requested):
I think your proposal makes sense and it is highlighting the issue of not having input parameter types.
However, I'd like to request a couple changes to address potential conflicts with the rest of the variable expansion logic - particularly the output variable expansion.

backend/services/workflows/app/processor/job.go
merlin-northern: this is not a good idea. I want to restrict it only to input parameters, as stated in the commit. reasoning: check the data the ProcessJobString is called with as arguments when a device is accepted:
* `-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIB...`
* `b8:01:01:01:01:01`
* `accepted`

do you honestly think that this is where we want the syntax to be supported?

backend/services/workflows/app/processor/job.go
merlin-northern: sure. fixed.

backend/services/workflows/app/processor/job.go
alfrunes: But isn't that what ProcessJobString is doing? It's expanding input parameters in the input string. I don't think it is recursively expanding the inputs. Or maybe I'm misunderstanding?

merlin-northern: so I even took a debugger and I checked with the https://github.com/mendersoftware/mender-server-enterprise/pull/827 and the `ProcessJobString` is _not_ called with the `${workflow.input.devices}` at _all_. what is wrong exactly to put it only for the JSON case? we can expand if needed.

merlin-northern: @alfrunes okey, lets scratch that, I added the support for any `workflow.input.$identifier.jsonInput.$field` pretty much anywhere -- just for you! 😘 there is a type switch which maybe is not very elegant but we have to move on.

merlin-northern:
@alfrunes the backend integration fails in main this is unrelated to this change.

backend/services/workflows/app/processor/job.go
merlin-northern: and I had to lint ignore gocyclo -- both of the functions are seriously in need of refactoring but it is far far beyond the scope and one day we have to do something about it.

backend/services/workflows/app/processor/string.go
```diff
@@ -136,6 +139,38 @@ SubMatchLoop:
...
+						return fmt.Sprintf("%d", d)
+					case int64:
+						return fmt.Sprintf("%d", d)
+					case bool:
+						if d {
+							return "true"
+						} else {
+							return "false"
+						}
+					}
```
alfrunes, line 159: This snippet will not output anything if dataType is none of the types in this switch (e.g. `null`). How about using the json codec by default?
```suggestion
					switch d := dataAny.(type) {
					case string:
						return d
					default:
						b, _ := json.Marshal(dataAny)
						return string(b)
					}
```

alfrunes: Or even just 
```suggestion
					return fmt.Sprint(dataAny) // may not be too nice with floats
```

alfrunes (commented):
I do find the logic here a bit confusing as the json parsing also has to go inside the ProcessJSON hack.
There's also still the shared prefix ambiguity as you cannot have an input variable with a suffix containing `.jsonInput`, but I don't think that will create any conflicts.

backend/services/workflows/app/processor/string.go
merlin-northern: as you wish. (and I did want to only support primitive types) but it is ok.

alfrunes (dismissed):

alfrunes (approved):

PR #1161, opened 2025-12-04
feat(inventory): internal patch /tenants/:tenant_id/device/:device_id/attributes

merlin-northern:
feat(inventory): internal patch /tenants/:tenant_id/device/:device_id/attribute
    
The new endpoint PATCH /tenants/:tenant_id/device/:device_id/attribute
works exactly like /tenants/:tenant_id/device/:device_id/attribute/scope/:scope
with the exception that it does not override the given attributes scope.
In other words: what you give in the body will maintain the scope.
    
Changelog: Commit
Ticket: MEN-9096

### DIFF
--- a/backend/services/inventory/api/http/api_inventory.go
+++ b/backend/services/inventory/api/http/api_inventory.go
@@ -50,19 +50,20 @@ const (
 	uriAttributes       = "/attributes"
 	uriDeviceAttributes = "/device/attributes"
 
-	apiUrlInternalV1         = "/api/internal/v1/inventory"
-	uriInternalAlive         = "/alive"
-	uriInternalHealth        = "/health"
-	uriInternalTenants       = "/tenants"
-	uriInternalDevices       = "/tenants/:tenant_id/devices"
-	urlInternalDevicesStatus = "/tenants/:tenant_id/devices/status/:status"
-	uriInternalDeviceDetails = "/tenants/:tenant_id/devices/:device_id"
-	uriInternalDeviceGroups  = "/tenants/:tenant_id/devices/:device_id/groups"
-	urlInternalAttributes    = "/tenants/:tenant_id/device/:device_id/attribute/scope/:scope"
-	urlInternalReindex       = "/tenants/:tenant_id/devices/:device_id/reindex"
-	apiUrlManagementV2       = "/api/management/v2/inventory"
-	urlFiltersAttributes     = "/filters/attributes"
-	urlFiltersSearch         = "/filters/search"
+	apiUrlInternalV1             = "/api/internal/v1/inventory"
+	uriInternalAlive             = "/alive"
+	uriInternalHealth            = "/health"
+	uriInternalTenants           = "/tenants"
+	uriInternalDevices           = "/tenants/:tenant_id/devices"
+	urlInternalDevicesStatus     = "/tenants/:tenant_id/devices/status/:status"
+	uriInternalDeviceDetails     = "/tenants/:tenant_id/devices/:device_id"
+	uriInternalDeviceGroups      = "/tenants/:tenant_id/devices/:device_id/groups"
+	urlInternalAttributes        = "/tenants/:tenant_id/device/:device_id/attribute/scope/:scope"
+	urlInternalAttributesNoScope = "/tenants/:tenant_id/device/:device_id/attributes"
+	urlInternalReindex           = "/tenants/:tenant_id/devices/:device_id/reindex"
+	apiUrlManagementV2           = "/api/management/v2/inventory"
+	urlFiltersAttributes         = "/filters/attributes"
+	urlFiltersSearch             = "/filters/search"
 
 	apiUrlInternalV2         = "/api/internal/v2/inventory"
 	urlInternalFiltersSearch = "/tenants/:tenant_id/filters/search"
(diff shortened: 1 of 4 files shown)


merlin-northern:
@alfrunes thanks for pointing out the discrepancy between https://github.com/mendersoftware/mender-server-enterprise/pull/827/files and this. fixed!

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -530,7 +530,9 @@ func (i *InternalAPI) PatchDeviceAttributesInternalHandler(
 		notModifiedAfter = &parsed
 	}
 	for i := range attrs {
-		attrs[i].Scope = c.Param("scope")
+		if attrs[i].Scope != model.AttrScopeSystem {
+			attrs[i].Scope = c.Param("scope")
+		}
```
alfrunes: I think having an internal endpoint without the scope path components would be less confusing.

alfrunes (commented):
I'm not sure if it's a good idea to start making exceptions like these to an endpoint. I believe this endpoint is clearly too narrow and a need for a more general internal endpoint seems to be required.
May I suggest we create a new internal endpoint? The current one not only suffers from inconsistent resource definitions, but also grammatical error in the API path.

backend/services/inventory/api/http/api_inventory.go
merlin-northern: even less confusing than overriding what you give in the scope in the body? :> ok do not worry -- I can create a new ep, which makes a lot more work, insanity with mmock alone will add tons of complexity. how about a parameter in the existing endpoint?

merlin-northern: discussed privately. we will do a new ep.

merlin-northern: so, @alfrunes just for you this is the new endpoint implementation. I want to skip the units this time, and create a backend integration test for the whole flow.

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -50,19 +50,20 @@ const (
...
+	apiUrlInternalV1             = "/api/internal/v1/inventory"
+	uriInternalAlive             = "/alive"
+	uriInternalHealth            = "/health"
+	uriInternalTenants           = "/tenants"
+	uriInternalDevices           = "/tenants/:tenant_id/devices"
+	urlInternalDevicesStatus     = "/tenants/:tenant_id/devices/status/:status"
+	uriInternalDeviceDetails     = "/tenants/:tenant_id/devices/:device_id"
+	uriInternalDeviceGroups      = "/tenants/:tenant_id/devices/:device_id/groups"
+	urlInternalAttributes        = "/tenants/:tenant_id/device/:device_id/attribute/scope/:scope"
+	urlInternalAttributesNoScope = "/tenants/:tenant_id/device/:device_id/attribute"
```
alfrunes: The damage is already done to the path right above, but I think we should fix the grammar on the new one.
```suggestion
	urlInternalAttributesNoScope = "/tenants/:tenant_id/device/:device_id/attributes"
```

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -529,8 +530,18 @@ func (i *InternalAPI) PatchDeviceAttributesInternalHandler(
...
 		notModifiedAfter = &parsed
 	}
+	scopeFromUrl := true
+	r := strings.NewReplacer(":tenant_id", tenantId, ":device_id", deviceId)
+	if strings.Contains(
+		c.Request.RequestURI,
+		r.Replace(urlInternalAttributesNoScope),
+	) {
+		scopeFromUrl = false
+	}
```
alfrunes: This could be simplified to a simpler expresion:
```suggestion
	scopeFromUrl := c.FullPath() == (apiUrlInternalV1 + urlInternalAttributes)
```

alfrunes (commented):
Thanks for creating a new endpoint, I believe this will cause less confusion. I have a couple very minor inputs, but otherwise looks very good.

backend/services/inventory/api/http/api_inventory.go
merlin-northern: amazin! I really forgot about the FullPath existence. changed.

backend/services/inventory/api/http/api_inventory.go
merlin-northern: sure. fixed.

merlin-northern:
@alfrunes the backend integration fails in main this is unrelated to this change.

alfrunes:
> @alfrunes the backend integration fails in main this is unrelated to this change.

Ouch! What happened to the integration tests? :monocle_face:

merlin-northern:
> > @alfrunes the backend integration fails in main this is unrelated to this change.
> 
> Ouch! What happened to the integration tests? 🧐

I wish I knew. there is a rather long thread on it. with no conclusions I may add.

alfrunes (approved):
LGTM.
Could you also add a simple acceptance test for the new endpoint too?

merlin-northern:
> LGTM. Could you also add a simple acceptance test for the new endpoint too?

could I please merge this, merge the workflows one, sync to ent, reimplement the sync using those and then create a test? I meed this 😘

alfrunes (approved):

PR #1162, opened 2025-12-05
fix(gui): replace condition for checking if billing profile is empty

alfrunes:
Changelog: None

### DIFF
--- a/frontend/src/js/components/subscription/SubscriptionDrawer.tsx
+++ b/frontend/src/js/components/subscription/SubscriptionDrawer.tsx
@@ -35,6 +35,7 @@ import {
   requestPlanUpgrade,
   startCardUpdate
 } from '@northern.tech/store/thunks';
+import { isEmpty } from '@northern.tech/utils/helpers';
 
 import CardSection from '../settings/CardSection';
 import { PlanExpandedForm } from '../settings/PlanExpandedForm';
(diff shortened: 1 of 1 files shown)


mzedel (dismissed):

alfrunes:
Sorry, I messed up the card detail condition in the translation. One sec...

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
```diff
@@ -162,11 +163,11 @@ export const SubscriptionDrawer = (props: SubscriptionDrawerProps) => {
         </div>
       )}
 
-      {isEdit || (isTrial && !billing) ? (
+      {isEdit || (isTrial && isEmpty(billing)) ? (
```
mzedel, line 166:
```suggestion
      {isEdit || cardDetailsDisabled) ? (
```

mzedel (approved):

frontend/src/js/components/subscription/SubscriptionDrawer.tsx
alfrunes: I also noticed that `cardDetailsDisabled` could be simplified to just `isEmpty(billing)`. I can do some cleanup in a follow-up PR :broom:

mzedel: hmm so it's not possible for any custom deal customers that are on a plan?

alfrunes: Ohh, yeah. You're right! It's not possible to sign up if the customer is not coming from trial.

PR #1166, opened 2025-12-07
feat(gui): delayed feedback form for 3h after first login

mineralsfree:
Should be merged after [nt-gui#435](https://github.com/NorthernTechHQ/nt-gui/pull/435)

- Fixed previous feedbackCollectedAt diff thas was always negative value
- Added a new check to ensure feedback is not shown until 3 hours have passed since the first login

Ticket: MEN-8896

### DIFF
--- a/frontend/src/js/components/header/Header.tsx
+++ b/frontend/src/js/components/header/Header.tsx
@@ -288,7 +288,7 @@ export const Header = ({ isDarkMode }) => {
   const { standard: deviceLimit } = useSelector(getDeviceLimits);
   const feedbackProbability = useSelector(getFeedbackProbability);
   const firstLoginAfterSignup = useSelector(getIsFirstLogin);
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/header/Header.tsx
```diff
@@ -340,8 +340,9 @@ export const Header = ({ isDarkMode }) => {
 
   useEffect(() => {
     const today = dayjs();
-    const diff = dayjs.duration(dayjs(feedbackCollectedAt).diff(today));
-    const isFeedbackEligible = diff.asMonths() > 3;
+    const lastFeedbackCollectedMonthsAgo = today.diff(feedbackCollectedAt, 'months');
+    const firstLoginHoursAgo = today.diff(firstLoginTimestamp, 'hours');
```
mzedel, line 344: the `firstLoginTimestamp` seems to be missing in the effect dependency array

mzedel (changes requested):
Thanks for catching the wrong diff calculation! just a tiny thing...

mzedel (approved):

PR #1167, opened 2025-12-08
chore(deplyoments): workaround for urllib HTTPResponse.headers.get re…

merlin-northern:
…moval

see also:https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html#what-are-the-important-changes

Ticket: None

### DIFF
--- a/backend/services/deployments/Makefile
+++ b/backend/services/deployments/Makefile
@@ -15,9 +15,13 @@ tests/%: ../../../backend/docs/api/deployments/%.yaml
 		-c tests/.openapi-generator.yml \
 		-o $(dir $@) \
 		--additional-properties=packageName=$* && \
-		sed -i.gitignore -e 's/"log":false/"log":False/g' tests/*/test/test_*.py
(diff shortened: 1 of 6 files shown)


merlin-northern:
@mzedel fyi there is also this: https://github.com/mendersoftware/mender-server/pull/1171 so we can pick.

mzedel:
> @mzedel fyi there is also this: #1171 so we can pick.

#1171 would be my preference between this and keeping the old, vulnerable, pinned version around... I also came across the sed approach, but your comment sums this up properly... IMO upgrading the generator is the way to go and it's... _something_ that this was possible to stay and an upgrade hasn't happened for so long... I created https://northerntech.atlassian.net/browse/MEN-9108 to track this.

alfrunes:
Thanks @mzedel for the ticket. I merged the 2.3.0 version pin. There's also a dependency constraint form `kubernetes` that also requires `2.3.0` for the integration test runner. I believe we are now in a position where we can centralize the generated Python clients using the new (`dist`) OpenAPI specs which would make it easier to maintain consistency for the generator versions.

merlin-northern:
so I am closing this one

PR #1169, opened 2025-12-08
MEN-8935: add `bytesprocessed` to accesslog entries

alfrunes:
This should cover the actual number of bytes consumed from the client which should also covered chunked transfer encoding.

### DIFF
--- a/backend/pkg/accesslog/middleware_gin.go
+++ b/backend/pkg/accesslog/middleware_gin.go
@@ -17,6 +17,7 @@ package accesslog
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


backend/pkg/accesslog/middleware_gin.go
```diff
@@ -107,6 +108,9 @@ func (a AccessLogger) LogFunc(
 	logCtx["responsetime"] = latency.String()
 	logCtx["status"] = c.Writer.Status()
 	logCtx["byteswritten"] = c.Writer.Size()
+	if bl, ok := c.Request.Body.(*BodyLogger); ok && bl != nil {
```
oldgiova, line 111: Nice solution for tracking requests size!

oldgiova (commented):
LGTM as much as I can review here.

oldgiova (approved):

frodeha (approved):
Looks good to me! 👍

PR #1171, opened 2025-12-08
chore: Align urllib3 version between integration and acceptance runner

alfrunes:


merlin-northern (approved):
I am ok with this, but plase consider https://github.com/mendersoftware/mender-server/pull/1167 too -- it took me some time

alfrunes:
@merlin-northern I think I'll go ahead and merge this one as it doesn't really block your PR. There still seems to be a few more issues, and I believe we should centralize the generated python client so it's easier to maintain the generator version. But that requires a bit more planning.

PR #1172, opened 2025-12-08
Chore: openapi-generator update + related api spec and test fixes

mzedel:


backend/services/deployments/Makefile
```diff
@@ -10,14 +10,11 @@ tests/%: ../../../backend/docs/api/deployments/%.yaml
 	docker run --rm -t -v $(MAKEDIR):/work -v $(MAKEDIR)/../../../backend/docs/api:/api -w /work \
 		--ulimit nofile=65536:65536 \
 		-u $(shell id -u):$(shell id -g) \
-		openapitools/openapi-generator-cli:v4.3.1 generate \
+		openapitools/openapi-generator-cli:v7.12.0 generate \
```
alfrunes, line 13: I see the latest release is `v7.17.0` is there a particular reason why you chose this version?

backend/services/deployments/Makefile
```diff
@@ -10,14 +10,11 @@ tests/%: ../../../backend/docs/api/deployments/%.yaml
...
 		-g python -i /api/deployments/$(notdir $<) \
 		-c tests/.openapi-generator.yml \
 		-o $(dir $@) \
-		--additional-properties=packageName=$* && \
-		sed -i.gitignore -e 's/"log":false/"log":False/g' tests/*/test/test_*.py
-		# see https://github.com/OpenAPITools/openapi-generator/issues/11421
-		# there is a bug in the python generator which produces non-python syntax
```
alfrunes, line 20: Thank you for cleaning this up!

backend/services/deployments/tests/tests/test_configuration_deployments.py
```diff
@@ -378,40 +378,28 @@ def test_fail_no_upgrade(self, api_client_int, clean_db, mongo):
...
             dc = SimpleDeviceClient()
-            nodep = {}
-            try:
-                dc.get_next_deployment(
-                    dev.fake_token_mt(other_tenant_id),
-                    artifact_name="dontcare",
-                    device_type=dev.device_type,
-                )
-            except ValueError:
-                nodep = None
```
alfrunes, line 389: Yikes. There might've been some context here that I'm missing, but this looks like a rather strange pattern.
Thanks for cleaning up :broom:

backend/services/deviceconfig/tests/tests/test_device_config.py
```diff
@@ -77,18 +77,18 @@ def test_config_device_set_get_remove(self, device_id):
         assert {"id": device_id, "reported": {}, "configured": {}} == {
             k: data[k] for k in ("id", "reported", "configured")
         }
-        assert "reported_ts" in data.keys()
+        assert "updated_ts" in data.keys()
```
alfrunes, line 80: This looks like the assertion was wrong in the past. I wonder how the test did passed before. Did the old client library just inject "empty" values for missing attributes?

alfrunes (approved):
Thanks for advancing the OAS generators by 5 years :zap:
Seems like the old client has been burying some actual issues with the tests over this time.
In the future it would be great if we could bundle also the python client with the new OAS so we have a single generator reference to maintain.
Anyway, superb work! :sparkles: 
If you believe it's possible, let's see if we can yank the version to the latest in a follow-up PR.

backend/services/deviceconfig/tests/tests/test_device_config.py
mzedel: afaict the new client will omit any undefined properties, but the old one didn't, so I assumed this was the way to go...

backend/services/deployments/Makefile
mzedel: just as a cooldown to not run into any recent changes, I'll use your approval to merge & create a follow up PR right away 👌 => https://github.com/mendersoftware/mender-server/pull/1173

PR #1174, opened 2025-12-09
chore: bump golang container image to 1.25.5

alfrunes:


merlin-northern (approved):
looks good to me. as I was doing similar things locally I was thinking, can this be controlled globally by some variable?

PR #1175, opened 2025-12-09
Fix/demo intervals on trial

mzedel:


frontend/src/js/components/devices/dialogs/PhysicalDeviceOnboarding.tsx
```diff
@@ -84,7 +83,23 @@ export const ExternalProviderTip = ({ hasExternalIntegration, integrationProvide
   </MenderTooltipClickable>
 );
 
-export const DeviceTypeSelectionStep = ({ hasConvertedImage, hasExternalIntegration, integrationProvider, onboardingState, onSelect, selection = '' }) => {
+interface DeviceTypeSelectionStepProps {
+  hasConvertedImage: boolean;
+  hasExternalIntegration: boolean;
+  integrationProvider: keyof typeof EXTERNAL_PROVIDER;
+  onboardingState: typeof initialState;
```
mineralsfree, line 90: maybe we should export the slices types from the store. `typeof initialState` works but a bit misleading

mineralsfree (approved):
Looks good

frontend/src/js/components/devices/dialogs/PhysicalDeviceOnboarding.tsx
mzedel: yes, it felt bad when I wrote it as I didn't want to go through the package release cycle 🙈 - but I'll follow up in nt-gui right away!

Issue #1180, opened 2025-12-10
OpenAPI Spec defines Attribute.Value as string (instead of object?)


OlliMartin:
Hi, 

> inb4: This is quite honestly a nitpick, but for everyone running code generators against the Open API spec provided, this creates a lot of problems; At least for languages with strict type-checking.

We are using the openapi spec provided to generate a C# client. This is working very well for us - so a big thanks in that regard!

However, the [Attribute](https://github.com/mendersoftware/mender-server/blob/main/backend/docs/api/inventory/schemas.yaml#L29) schema is consistently defined as `type: string`:

```
        value:
          type: string
          description: |
            The current value of the attribute.

            Attribute type is implicit, inferred from the JSON type.

            Supported types: number, string, array of numbers, array of strings.
            Mixed type arrays are not allowed.
```

This trips up at least C# and I'm assuming other generators with strict type checkers as well, since - as the description states - the value can also be a list of string or numbers (say for example for NICs).

Is there a reason it is maintained as string? Would you accept a PR that aligns the type to `object` ? 
In general it's a bit of a headache having to differentiate between json tokens, but doable. However, at least in C#, we cannot make string work.

We patched it manually for inventory mgmt v1 + v2; I'm not sure if other places would need to be adjusted as well.
Of course it would be nicer if we could just use the OpenAPI spec as is. :)

Cheers
Olli

alfrunes:
Hello 👋 
Thanks again for the report. I see there was a PR with corrections to the specs https://github.com/mendersoftware/mender-server/pull/1182 already.
Does this solve your problem?

OlliMartin:
Definitely, thanks a lot!

Sorry, I forgot to close the issue. :)
Cheers

PR #1182, opened 2025-12-10
spec adjustments for #1180

mzedel:
see #1180 for context

### DIFF
--- a/backend/docs/api/inventory/management_v1.yaml
+++ b/backend/docs/api/inventory/management_v1.yaml
@@ -674,7 +674,15 @@ components:
           type: string
           description: Attribute description.
         value:
-          type: string
+          oneOf:
+            - type: string
+            - type: number
+            - type: array
+              items:
+                type: string
+            - type: array
+              items:
+                type: number
           description: |
             The current value of the attribute.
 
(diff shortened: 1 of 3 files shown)


backend/docs/api/inventory/management_v1.yaml
```diff
@@ -674,7 +674,6 @@ components:
           type: string
           description: Attribute description.
         value:
-          type: string
```
merlin-northern, line 677: I am having some doubts, have you tested it? what would the oapi-codegen generate? why not use `oneOf`?

mzedel: The model code says https://github.com/mzedel/mender-server/blob/main/backend/services/inventory/model/device.go#L67 - so I thought to be correct that would be the way, OTOH `oneOf` would be more helpful => I'll adjust!

As to oapi-codegen: we can have this as a requirement as soon as we have anything actually relying on this - so far the only generators actually using the specs seem to be outside of the backend.

merlin-northern (approved):
very pretty

backend/docs/api/inventory/management_v1.yaml
OlliMartin: JFYI: I checked for C# and the union type is generated as expected (using openapi-generator); I had some doubts, which is why I proposed object. 
NSwag will fail to generate, which is fine.. we'll switch.

Thanks a lot for the very quick response!

PR #1186, opened 2025-12-11
MEN-9082: Get inventory device statistics

frodeha:
**Description**

This PR adds a new endpoint in inventory that provides statistics about devices. For now it only provides number of accepted and pending devices per tier, but I can see the set of statistics being expanded in the future should the need arise.

Tiers doesn't really exist as a concept in open source, so this endpoint won't come into it's own until it's merged to enterprise. On its own it also won't really work properly as no tier information is propagated to inventory yet. The endpoint is however implemented by trusting the agreement that there will (at some point) be a system scoped attribute called "tier" propagated from deviceauth. As such, the endpoint will not return anything but zeros for micro and system tier (even after the merge to enterprise) for the time being, but this should change automatically when [MEN-9069](https://northerntech.atlassian.net/browse/MEN-9096) is done.

**Notes on implementation**
* The endpoint itself is fairly generic ("statistics"), so I wanted to leave some room for future extension (without having to break backwards compatibility) should the need arise. Hence I opted with introducing a level of nesting in the response (e.g `"devices_by_status"`). I went back and forth with myself a couple of times on a) should I even do this and b) if so, what should the name of the property be. This is what I ended up with, but opinions to the contrary are welcome. 
* I had to pull in some bits and pieces from the tiers implementation in enterprise (the tier definitions and a integration test support function) to make this endpoint work while also (hopefully) not making the merge to enterprise too painful.
* The integration test is disabled for now as tier information is not yet propagated to inventory and the assertions fail because all the devices look like standard devices because of it. I will make sure to come back and enable the test when [MEN-9069](https://northerntech.atlassian.net/browse/MEN-9096) is done.

[MEN-9069]: https://northerntech.atlassian.net/browse/MEN-9069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MEN-9069]: https://northerntech.atlassian.net/browse/MEN-9069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kjaskiewiczz:
I'd ask @mzedel and/or @mineralsfree to review the first commit (the one with API spec).

frodeha:
> I'd ask @mzedel and/or @mineralsfree to review the first commit (the one with API spec).

Added now. That being said, I already discussed this with Manuel directly before implementing which is why I didn't initially add him.

mzedel (commented):
LGTM! Thank you for bringing this to OS 🙏 - I'll use this to get the UI-backend communication part ready right away!

For context: it worked great with https://github.com/NorthernTechHQ/nt-gui/pull/451 + https://github.com/mendersoftware/mender-server/pull/1201 .

mzedel (dismissed):
I rechecked the aggregation pipeline & would suggest anything discovered, can still be adjusted in a follow up + it will be easier to do when seeing it in action with the related UI.

mzedel (approved):

frodeha:
@kjaskiewiczz: I welcome your review of these changes!

However, seeing as this PR is blocking the progress of the entire tiers epic at this point and we have committed to doing a demo of features that require this PR in today's retro and I have Manuel's approval I will go ahead and merge the PR as it stands.

Please do feel free to have a look and let me know if there's anything you'd like to change about it and I will be more than happy to make follow-up PR's addressing any and all comments.

I hope you're okay with this slightly unorthodox process this one time - I promise to not make it a habit! 🙏

PR #1187, opened 2025-12-11
fix: ensure consistent CSP header for `wss` and `https` domains

alfrunes:


frontend/httpd.conf
```diff
@@ -47,7 +47,7 @@ http {
...
+      add_header Content-Security-Policy "default-src 'none'; child-src 'self' blob:; worker-src 'self' blob:; connect-src 'self' wss://localhost wss://*.hosted.mender.io https://localhost https://*.mender.io https://api.stripe.com https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://stats.g.doubleclick.net; script-src 'self' https://*.googletagmanager.com https://js.stripe.com https://cdn.jsdelivr.net/npm/cookieconsent@3/build/ https://www.google.com www.gstatic.com 'unsafe-inline'; img-src 'self' data: https://*.google-analytics.com https://*.googletagmanager.com https://*.tile.openstreetmap.org; font-src 'self'; frame-src https://js.stripe.com https://hooks.stripe.com https://www.google.com; style-src 'self' https://cdn.jsdelivr.net/npm/cookieconsent@3/build/ 'unsafe-inline'; object-src 'none'; base-uri 'self'; form-action 'self'; frame-ancestors 'none';";
```
mzedel: would something like:
```suggestion
      add_header Content-Security-Policy "default-src 'none'; child-src 'self' blob:; worker-src 'self' blob:; connect-src 'self' wss://$host https://$host https://api.stripe.com https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://stats.g.doubleclick.net; script-src 'self' https://*.googletagmanager.com https://js.stripe.com https://cdn.jsdelivr.net/npm/cookieconsent@3/build/ https://www.google.com www.gstatic.com 'unsafe-inline'; img-src 'self' data: https://*.google-analytics.com https://*.googletagmanager.com https://*.tile.openstreetmap.org; font-src 'self'; frame-src https://js.stripe.com https://hooks.stripe.com https://www.google.com; style-src 'self' https://cdn.jsdelivr.net/npm/cookieconsent@3/build/ 'unsafe-inline'; object-src 'none'; base-uri 'self'; form-action 'self'; frame-ancestors 'none';";
```
also work? or maybe only `'self'`?

OTOH: there is still the discrepancy between `wss://*.hosted.mender.io` and `https://*.mender.io` (I'm specifically looking at `docker.mender.io`)

alfrunes: > OTOH: there is still the discrepancy between wss://*.hosted.mender.io and https://*.mender.io (I'm specifically looking at docker.mender.io)

Right! That was my intention, but muscle memory took over :see_no_evil: 

Do you recall why these specific rules were added in the first place?

alfrunes: I removed the `wss` domain entirely to instead rely on self. To avoid CSP bypass in the compose environment I added `Host` rules to Traefik.

mzedel: Not 100% as it's been a while 🙈.
But I think the regular focus was to cover the regular Mender setups (localhost + all the dev/ production mender.io domains) and websocket connections were only supposed to be allowed to the backend the page is served from to prevent diverting the connection to a different backend (due to the access the troubleshoot addon has to the device).

mzedel (approved):

PR #1189, opened 2025-12-11
test(gui): let client logs always be collected, not only on failures

mzedel:
added you @lluiscampos as you worked on this recently and might be able to assess the usefulness for this


frontend/tests/e2e_tests/run.js
lluiscampos, line 680: Pre-existing, but this log is miss-leading: it will say failed if it for example cannot save the "error.txt" log file but it will say success if it cannot save the client or docker-compose logs. This is something that I had issues with locally and took me a while to understand 🫠

lluiscampos (approved):
Good change! (I had something like this locally while debugging)

PR #1193, opened 2025-12-11
QA-1417: Pin QEMU image for e2e tests to the previous working version

lluiscampos:
This alias is for yesterday's image, which contains the demo certificates in `/etc/ssl/certs`.

It is a temporary workaround while we figure out the actual problem with the yocto build.


elkoniu (approved):

mzedel (approved):
Thank you @lluiscampos! With the branch protected on Gitlab I triggered a run and the job passed: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/12410124234

PR #1200, opened 2025-12-15
chore: add `traefik` as a valid routing hostname

alfrunes:
Required by integration tests.


merlin-northern (dismissed):
as usual, faster than me.

merlin-northern:
@alfrunes we have to trigger the backend integration on changes to those files, right? but lets get this one in first.

mzedel (approved):

PR #1201, opened 2025-12-15
Refac/use tiered counts

mzedel:


frontend/src/js/components/dashboard/Devices.tsx
```diff
@@ -52,7 +52,7 @@ export const Devices = ({ clickHandle }) => {
     const issueRequests = Object.keys(availableIssueOptions).map(key =>
       dispatch(getIssueCountsByType({ type: key, options: { filters: [], selectedIssues: [key] } }))
     );
-    return Promise.all([dispatch(getDeviceCount(DEVICE_STATES.accepted)), dispatch(getDeviceCount(DEVICE_STATES.pending)), ...issueRequests]);
+    return Promise.all([dispatch(getAllDeviceCounts()).unwrap(), ...issueRequests]);
```
mineralsfree, line 55: This never handled `unwrap` confuses me a bit; why do we need it here?

mineralsfree (dismissed):
👍

frontend/src/js/components/dashboard/Devices.tsx
mzedel: primarily to not have the `Promise.all` settle when the `dispatch` call is done - I guess the `issueRequests` should get unwrapped for completeness... longer term I want to avoid having more instances where the early resolve caused issues (e.g. on app init or group creation errors getting swallowed) by not requiring the consumer to check if the returned promise is actually the right one...

mineralsfree (approved):

PR #1203, opened 2025-12-15
fix(deviceconnect): do not send close frame already received from peer

alfrunes:


backend/services/deviceconnect/api/http/device.go
```diff
@@ -186,12 +187,22 @@ func (h DeviceController) Connect(c *gin.Context) {
...
+			select {
+			case errChan <- err:
 
-		case <-time.After(time.Second):
-			l.Warn("Failed to propagate error to client")
+			case <-time.After(time.Second):
+				l.Warn("Failed to propagate error to client")
+			}
 		}
+		_ = c.Error(err)
```
merlin-northern, line 205: could we have a comment around this line? we do not very often see this in the code.

backend/services/deviceconnect/api/http/management.go
```diff
@@ -634,8 +628,27 @@ func (h ManagementController) ConnectServeWS(
...
+			// Notify writer to handle error
+			select {
+			case errChan <- err:
+
+			case <-time.After(time.Second):
+				l.Warn("Failed to propagate error to client")
+			}
+		}
+	}
+	return err
```
merlin-northern, line 651: this is almost the same as the above. could we have it like one uniform call to handle the errors?

merlin-northern (approved):

backend/services/deviceconnect/api/http/management.go
alfrunes: https://github.com/mendersoftware/mender-server/pull/1225

PR #1206, opened 2025-12-16
MEN-8669: New release select dialog in deployment page

mineralsfree:


frontend/tests/e2e_tests/utils/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+//    limitations under the License.
+import type { Page } from '@playwright/test';
+
+import { timeouts } from './constants.ts';
+
+export const selectReleaseByName = async (page: Page, name: string) => {
+  await page.getByRole('button', { name: 'Select a release' }).click();
+  await page.getByRole('textbox', { name: 'Search releases...' }).fill(name);
+  await page.waitForTimeout(timeouts.default);
+  return page.locator(`#deployment-release-container > div:has-text('${name}')`).click();
```
mzedel: considering the number of places the
```
page.locator(`#deployment-release-container > div:has-text('${name}')`)
```
part exists, I would suggest having yet another util here with something like:
```typescript
export const locateReleaseByName = (name: string) => {
  const container = page.locator('#deployment-release-container')
  return container.getByText(name)
}
```
for the better locator stability compared to looking for the child `div` & to keep the container selector limited to fewer places... although I'm not sure if there may not be situations where you get multiple matches down the dom tree - but that may also happen with `has-text` I guess 🤔

frontend/src/js/components/deployments/deployment-wizard/ReleaseItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Typography variant="body2" className={`margin-bottom-x-small ${classes.linesOverflow}`}>
+          {description}
+        </Typography>
+      )}
+      <div className="margin-top-x-small">
+        {tags?.map(tag => (
+          <Chip className="margin-right-x-small" key={tag} label={tag} size="small" />
+        ))}
+      </div>
+      <Typography variant="caption">Updated {dayjs(modified).format('YYYY-MM-DD HH:mm')}</Typography>
```
mzedel:
```suggestion
      <Typography variant="caption">Updated {<Time value={modified} />}</Typography>
```

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { useAppDispatch } from '@northern.tech/store/store';
+import { getReleases } from '@northern.tech/store/thunks';
+import { useDebounce } from '@northern.tech/utils/debouncehook';
+
+import { ReleaseItem } from './ReleaseItem';
+
+const useStyles = makeStyles()(theme => ({
+  advancedFilterContainer: {
+    width: 'calc(50% - 8px)'
+  },
```
mzedel: that to me looks a lot like what our `.two-columns` class does/ what a grid implementation would be better suited for...

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    renderInput={params => <TextField {...params} label="Select type..." placeholder="Select a type" />}
+                  />
+                </div>
+              </div>
+            )}
+            <div className="flexbox column margin-top-small margin-bottom-small">
+              <Typography variant="subtitle2" className="margin-bottom-x-small">
+                Release name
+              </Typography>
+              <ControlledSearch onSearch={() => Promise.resolve()} name="searchTerm" placeholder="Search releases..." />
```
mzedel: not sure if the `onSearch` should outright be the fallback in case it's not provided as a prop...

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+            </div>
+          </form>
+        </FormProvider>
+        <Divider />
+        <Typography className="margin-top-small" variant="body1">
+          Results({releaseItems.length})
```
mzedel:
```suggestion
          Results ({releaseItems.length})
```

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }
+}));
+
+export const ReleaseArtifactFilter = props => {
+  const { open, onClose, onSelect } = props;
+  const { classes } = useStyles();
+  const [initialValues] = useState({ tags: [], type: null, searchTerm: '' });
+  const [filterCount, setFilterCount] = useState(0);
+  const dispatch = useAppDispatch();
+  const [advancedFilters, setAdvancedFilters] = useState(false);
```
mzedel: maybe `showAdvancedFilters` or something else pointing to the boolean character of the variable?

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+import { Close as CloseIcon, ExpandLess as ExpandLessIcon, ExpandMore as ExpandMoreIcon, FilterList as FilterListIcon } from '@mui/icons-material';
+import { Button, DialogContent, Divider, TextField, Typography } from '@mui/material';
+import { makeStyles } from 'tss-react/mui';
+
+import ChipSelect from '@northern.tech/common-ui/ChipSelect';
+import { ControlledSearch } from '@northern.tech/common-ui/Search';
+import { BaseDialog } from '@northern.tech/common-ui/dialogs/BaseDialog';
+import { ControlledAutoComplete } from '@northern.tech/common-ui/forms/Autocomplete';
+import { Release } from '@northern.tech/store/releasesSlice';
```
mzedel, line 26: What do you think: should we have a `@northern.tech/store/types` for the UI internal types?

frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
```diff
@@ -300,18 +282,25 @@ export const Software = ({ commonClasses, deploymentObject, releaseRef, releases
...
+              />
+              <Button
+                size="large"
+                color="neutral"
+                variant="outlined"
+                className={`text-overflow ${classes.releaseSelect}`}
+                endIcon={releaseFilterOpened ? <ExpandLessIcon /> : <ExpandMoreIcon />}
+                onClick={() => setReleaseFilterOpened(!releaseFilterOpened)}
+              >
+                <span className={`${classes.releaseSelectText} text-overflow`}>{deploymentRelease?.name ?? 'Select a release'}</span>
```
mzedel: I haven't looked into it, but the double `text-overflow` here looks a bit suspicious...

frontend/src/js/components/deployments/deployment-wizard/ReleaseItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+        {name}
+      </Typography>
+      {description && (
+        <Typography variant="body2" className={`margin-bottom-x-small ${classes.linesOverflow}`}>
+          {description}
+        </Typography>
+      )}
+      <div className="margin-top-x-small">
+        {tags?.map(tag => (
+          <Chip className="margin-right-x-small" key={tag} label={tag} size="small" />
```
mzedel, line 58: Strictly speaking that would have to be capped for overflow after 2 lines as well I guess, but 🤫

mzedel (dismissed):
Nicely done, a few nitpicks with only the two column layout being a bit uncomfortable, but nothing I couldn't live with/ we couldn't follow up in a coming PR 👍

frontend/tests/e2e_tests/utils/utils.ts
mineralsfree: Fair point. I've updated it to pick the first match in case other entries appear down the tree, which might be tags or description of the release

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
mineralsfree: I've added an optional property that makes the search behave as an RHF-controlled field. Let me know if you have a better suggestion

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
mineralsfree: That makes sense to me

frontend/src/js/components/deployments/deployment-wizard/ReleaseItem.tsx
mineralsfree: Let's wait for someone to complain about it. 👼 🤞

mzedel (approved):
🎉 looking forward to seeing this in use!

PR #1211, opened 2025-12-17
test(gui): Combine stderr and stdout streams when running commands

lluiscampos:


lluiscampos:
@mzedel This is the last one that I had. The motivation was to have combined stdout/stderr on the [client container log](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/12502130709/artifacts/external_file/frontend/logs/client.log) but I think it is good to have in general for all commands.

I you prefer to keep the rest of the commands as it was, I can add a new option `combineStdStreams` and then `resolve` with either stdout only or both depending of the option.

mzedel (approved):
Apparently I have to reassess my assumptions about the stderr usage 🫣. Thank you very much @lluiscampos !

PR #1214, opened 2025-12-18
MEN-8879 + MEN-8882: device usage and limits with tier support in header and widget

mineralsfree:


frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+  const viewPendingClick = () => {
+    handleClose();
+    navigate('/devices/pending');
+  };
+  const { counts: accepted } = useSelector(getAcceptedDevices);
+  const { isAdmin } = useSelector(getUserRoles);
+
+  const limits = useSelector(getDeviceLimits);
+  const mappedLimits = Object.entries(limits)
+    .filter(([, limit]) => limit > 0)
```
mzedel: I think this would at least need 
```suggestion
    .filter(([, limit]) => limit !== 0)
```
to allow us to show unlimited tiers (with a limit of `-1`), or more likely some additional logic, as the `standard` devices are currently still unlimited for both -1 and 0 😬...

frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+    navigate('/settings/subscription');
+  };
+  const viewPendingClick = () => {
+    handleClose();
+    navigate('/devices/pending');
+  };
+  const { counts: accepted } = useSelector(getAcceptedDevices);
+  const { isAdmin } = useSelector(getUserRoles);
+
+  const limits = useSelector(getDeviceLimits);
```
mzedel: purely for consistency in the code navigation consider moving these to the state definitions etc. (unlikely here, but that way we get less surprises about hook execution numbers - e.g. if there would be an early return for an empty state introduced)

frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+        disableTouchListener
+        disableHoverListener={!!anchorEl}
+        onMouseEnter={() => !anchorEl && setTooltip(true)}
+        onMouseLeave={() => setTooltip(false)}
+      >
+        <div>
+          <Badge color={severity} badgeContent={pending} anchorOrigin={{ horizontal: 'left' }}>
+            <Button
+              startIcon={<DeveloperBoardIcon className="margin-right-x-small" fontSize="small" />}
+              endIcon={<ArrowDropDownIcon fontSize="small" sx={{ color: 'text.secondary' }} className={classes.endIcon} />}
```
mzedel:
```suggestion
              endIcon={<ArrowDropDownIcon fontSize="small" className={classes.endIcon} />}
```
considering you have the `endIcon` class defined

frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+              className={`flexbox center-aligned ${classes[severity]} ${!!anchorEl && classes.pressed}`}
+              onClick={handleOpen}
+            >
+              <span>{total.toLocaleString('en-US')}</span>
```
mzedel: is the `span` needed here?

frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+      <div className="flexbox full-width space-between">
+        <div className="flexbox">
+          <Typography variant="subtitle2" className="capitalized-start">
+            {type}
+          </Typography>
+          {warning && <WarningAmberIcon fontSize="small" color="warning" className="margin-left-x-small" />}
+          {error && <ErrorOutlineIcon fontSize="small" color="error" className="margin-left-x-small" />}
+        </div>
+        <Typography variant="body2">
+          {total.toLocaleString('en-US')}/{limit.toLocaleString('en-US')}
```
mzedel: coming from a non `,`, but `.` separating country I find the US variant very disturbing 😜... - but could you define the target locale somewhere so that it is consistent across the file?

frontend/src/js/components/header/DeviceNotifications.tsx
```diff
@@ -11,86 +11,235 @@
...
+              </div>
+            </div>
+            <Divider />
+            <div className="margin-top-small">
+              {mappedLimits.map(limit => (
+                <DeviceLimit key={limit.type} {...limit} />
+              ))}
+            </div>
+            {isAdmin && (
+              <Button color="primary" variant="text" onClick={() => onManageDeviceLimit()}>
```
mzedel: to be more semantically correct I think:

```suggestion
              <Button color="primary" variant="text" component={Link} to="/settings/subscription" onClick={handleClose}>
```

_should_ also work, but the `onManageDeviceLimit` is also ok

frontend/src/js/components/header/DeviceNotifications.tsx
mzedel: looks like that wasn't included in the task design (only one of the related ones: https://www.figma.com/design/ZTGKMcKtHrZ0V7pfdws7KM/Zephyr-UX-changes?node-id=4524-2919&t=qSkPRQsL0cfGSUT4-4) 😞 - it will be present in on-prem setups though

mzedel (changes requested):
looks nice! a few nitpicks, only the unlimited case is really needed IMO - let me know if you'd rather push this to a separate task instead

frontend/src/js/components/header/DeviceNotifications.tsx
mineralsfree: I've added the commit that treats -1 as unlimited, but if we need to handle both the old and new approaches to 0, I think it deserves a separate task. We can wait with merging this one

mzedel: hmmm I _think_ we kept the `standard w/ limit 0 == unlimited` mainly due to the potential problems for the UI, but I'm not sure... I guess since -1 is _the future_ I would rather merge it and try to align the rest of the product to the unified understanding, then to pour more complexity to cover this up 🫣...

mzedel (approved):

PR #1218, opened 2025-12-18
test(gui): Misc fixes around `runCommand`

lluiscampos:
Fixes and improvements to help me debug QA-1417 aka "the QEMU issue"

### DIFF
--- a/frontend/tests/e2e_tests/run.js
+++ b/frontend/tests/e2e_tests/run.js
@@ -267,12 +267,13 @@ const composeLogs = async config =>
 
 const runCommand = (command, args = [], config, options = {}) =>
   new Promise((resolve, reject) => {
-    const { quiet = true, throwOnError = true, ...remainderOptions } = options;
+    const { quiet = true, throwOnError = true, shell = false, ...remainderOptions } = options
     let stdout = '';
     let stderr = '';
     const child = spawn(command, args, {
       stdio: ['inherit', 'pipe', 'pipe'],
       env: { ...process.env, TEST_ENVIRONMENT: config.environment },
+      shell,
       ...remainderOptions
     });
     child.stdout.on('data', data => {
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e_tests/run.js
```diff
@@ -313,7 +314,9 @@ const runCommand = (command, args = [], config, options = {}) =>
 
     child.on('error', error => {
       cleanup();
-      reject(error);
+      if (throwOnError) {
+        reject(error);
+      }
```
lluiscampos, line 319: @mzedel This is not (yet) the full fix, it still fails the cleanup with something like:
```
Cleanup💥 Failed to dump logs: Error: Command failed with exit code 1: (...)
cat: '/data/mender/deployment*.log': No such file or directory
```

I thought that the Promise won't throw errors, but just do the `resolve` or `reject` 🤔 

I can remove this whole "feature" about `throwOnError` if you want and find an alternate way to get the deployments logs.

lluiscampos: By the way, the above output is by forcing the tests to exit early and have no deployments, when the tests fail after a deployment, the logs get collected now with the other fix (the shell) 👍

frontend/tests/e2e_tests/run.js
```diff
@@ -257,7 +257,7 @@ const composeLogs = async config =>
     '📋 Collecting container logs...',
     async () => {
       console.log('getting em logs');
-      const result = await compose.logs(getComposeOptions(config));
+      const result = await compose.logs([], getComposeOptions(config));
```
mzedel, line 260: oh my 🤦 - thanks for catching that, I've seen the corresponding log output, but didn't look into it yet, as I primarily rely on the trace info alone 😬

frontend/tests/e2e_tests/run.js
mzedel: Hmm I assume that's an early rejection here, which will be avoided with the `throwOnError: false`, but it will probably still go through the `child.on('close',` further up and there get the non-0/ 1 exit code and then it gets thrown... so maybe both need the check for `throwOnError` (not sure if we can safely `resolve` here as an alternative, as I would imagine the child process could linger around otherwise) - what do you think?

mzedel (dismissed):
Thank you so much for diving in here as far as you did 🤿 🦈 🐟 🐋 !

frontend/tests/e2e_tests/run.js
lluiscampos: I don't understand.

`child.on('close',` further up already has the check for `throwOnError` -> it will only call `reject` if `throwOnError` is true and non-0 exit code.

So the problem must be somewhere else 🤔

lluiscampos: > what do you think?

I'll keep digging a bit on my free time 🪏

lluiscampos: @mzedel  :dart: https://github.com/mendersoftware/mender-server/pull/1218/changes/b62320ffc46e5ff84d5a6a49001aef51d3a23848

lluiscampos:
@mzedel I found the bug (see discussion above) and I rebased from main to have your related changes wrt collecting the logs always. Your review got stale so please re-review :eyes:

mzedel (approved):
🥇 amazing work @lluiscampos - thank you for sticking with it!

PR #1225, opened 2025-12-23
refactor(deviceconnect): consolidate websocket read error handling

alfrunes:


merlin-northern (approved):
very sweet thank you! I restarted the units, occasionally there is a failure in a deviceconnect test.
```
            	Test:       	TestManagementConnect/ok
            	Messages:   	The error websocket: close 1006 (abnormal closure): unexpected EOF is not a close error with status 1011
```
I wonder if that is something that we should acutally once take a look at.

PR #1227, opened 2025-12-31
feat(workflows): default value for input variables and gojsonq fix

merlin-northern:
With this change you can use the following syntax
to set a default value for an empty input variable:

${workflow.input.device.jsonInput.tier|standard}

There is also a required in some environments
fix for proper querying of the fields via jsonInput.

see for usage: https://github.com/mendersoftware/mender-server-enterprise/pull/865

Changelog: Commit
Ticket: MEN-9096


backend/services/workflows/app/processor/job.go
merlin-northern, line 1: @alfrunes I know it is hacky, but it is getting hard to move forward.

alfrunes (approved):

PR #1249, opened 2026-01-06
chore(pkg): initialized new package for type utils

alfrunes:
Initialized the package with a simple Pointer generic for dereferencing variables and constants.
One pointer func to rule them all!

(Cherry-picked from enterprise).


merlin-northern (approved):
it looks good, but I was expecting something that would replace the default sequence:
```go
someString:=p
return &someString
```
i.e.: you know all the dance to get the pointer to something.

PR #1251, opened 2026-01-07
fix(deviceconnect): Race condition sending websocket error to peer

alfrunes:
A race condition occur when the websocket handler terminates where it will sometimes close too soon, resulting in an unexpected EOF error observed from the writing go routine. This commit makes the writer the main go routine to ensure the writer always terminates before returning and closing the connection.

Ticket: QA-1445


bahaa-ghazal (approved):
LGTM! 
Only a typo in the commit msg (to soon -> too soon).

PR #1256, opened 2026-01-08
MEN-9205: MCU feature flag added

mineralsfree:
Until MCU is enabled, the onboarding modal will look like something in between the old and new approaches: 
image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,7 +11,7 @@
     "@mui/icons-material": "7.3.6",
     "@mui/material": "7.3.6",
     "@mui/x-date-pickers": "8.23.0",
-    "@northern.tech/store": "0.13.0",
+    "@northern.tech/store": "0.14.0",
     "@northern.tech/themes": "0.10.2",
     "@northern.tech/utils": "0.7.2",
     "@redux-devtools/extension": "3.3.0",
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -135,7 +135,7 @@
     "lint": "eslint src --ext .js,.jsx,.ts,.tsx && eslint tests --ext .js,.jsx,.ts,.tsx",
     "lint-fix": "eslint src --ext .js,.jsx,.ts,.tsx --fix && eslint tests --ext .js,.jsx,.ts,.tsx --fix",
     "serve": "rspack serve --mode development",
-    "test": "vitest --dir src/js",
+    "test": "vitest --dir src/js/components/devices/dialogs",
```
mzedel: that looks unintentional 😉 - just in case: I usually do either `npm run test devices/dialogs` or go straight to `npm run test DeviceConnectionDialog`

mzedel (changes requested):
👍 thanks for the rapid fix! just one tiny thing...

frontend/package.json
mineralsfree: Indeed 😅 Thank you

mzedel (approved):

PR #1257, opened 2026-01-08
QA-903: Enable some disabled gosec rules in mender-server project

bahaa-ghazal:


bahaa-ghazal:
Btw, I'll merge the changes with enterprise before OS

backend/services/workflows/model/task.go
```diff
@@ -22,6 +29,15 @@ const (
 	TaskTypeSMTP = "smtp"
 )
 
+var (
+	ErrCommandNotFound   = errors.New("command not found")
+	ErrCommandNotAllowed = errors.New("command not allowed")
+
+	allowedCommands = map[string]string{
+		"create-artifact": "/usr/bin/create-artifact",
+	}
```
alfrunes: I wonder if it would make sense to put this behind a configuration option. Different services have different requirements when it comes to executables. There is `create-artifact-worker` that has one binary, while `generate-delta-worker` require a different binary, while the plain `workflows-worker` does not require executing anything.
What do you think?

alfrunes (commented):
I'm not sure about the G204 fix, I'm thinking maybe the list of allowed executables should be a configuration option to make it somewhat easier to maintain for the services that require this permission.
The rest looks good :+1:

backend/services/workflows/model/task.go
bahaa-ghazal: That is a decent idea, and it will make it more secure for sure. I'll try to implement either per service white-list or per workflow.

backend/pkg/executor/executor.go
```diff
@@ -0,0 +1,39 @@
...
+	ErrCommandNotAllowed = errors.New("command not allowed")
+)
+
+//go:generate ../../utils/mockgen.sh
+type BinaryExecutor interface {
+	Command(ctx context.Context, command string, arg ...string) (*exec.Cmd, error)
+}
+
+type executor struct {
+	allowedBinaries map[string]string
```
alfrunes: Why not a string slice?
```suggestion
	allowedBinaries []string
```
Or a set?
```suggestion
	allowedBinaries map[string]struct{}
```

backend/services/workflows/app/processor/nats.go
```diff
@@ -12,20 +12,18 @@
...
 	"github.com/mendersoftware/mender-server/services/workflows/model"
 )
 
-func processNATSTask(
+func (jp *JobProcessor) ProcessNATSTask(
 	natsTask *model.NATSTask,
-	ps *processor.JobStringProcessor,
-	jp *processor.JobProcessor,
+	ps *JobStringProcessor,
 	nats nats.Client,
```
alfrunes: The `nats.Client` should now be in the `JobProcessor` state.

backend/services/workflows/config.yaml
```diff
@@ -126,3 +126,9 @@ workflows_path: "/etc/workflows/definitions"
 # Overwrite with environment variable: WORKFLOWS_CONCURRENCY
 
 # concurrency: 10
+
+# Allowed binaries that can be executed in the workers.
+# The key of the map is the command, and the value is the path to the binary.
+
+allowed_binaries:
+#  {cmd}: {path}
```
alfrunes: I think it's more preferable to have a list of whitelisted executables as this is kind of re-implementing PATH lookups.

backend/pkg/executor/executor.go
bahaa-ghazal: It's just for security reasons. Sometime in the worflows json scheme it uses the absolute path of the binary while other times uses just the command. Mainly to make sure that commands are translated to correct binaries.

alfrunes:
@bahaa-ghazal Is this PR ready for review? The acceptance tests are failing and there don't seem to be any change since last review.

backend/services/create-artifact-worker/workflows/generate_artifact.json
alfrunes, line 1: You also need to bump the `version` field of this job so that workflows reloads it (otherwise it will continue to use the definition it already has).

backend/pkg/executor/executor.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// New creates a new executor with a specific set of allowed binaries.
+func New(allowed []string) BinaryExecutor {
+	return &executor{
+		allowedBinaries: allowed,
+	}
+}
+
+func (e *executor) Command(ctx context.Context, binarie string, arg ...string) (*exec.Cmd, error) {
```
alfrunes:
```suggestion
func (e *executor) Command(ctx context.Context, binary string, arg ...string) (*exec.Cmd, error) {
```

backend/services/workflows/app/processor/nats.go
alfrunes: This comment was marked resolved, but the `nats` argument is still there despite being part of `JobProcessor` state.

alfrunes (changes requested):
Just a couple more minor things. Most importantly is to bump the workflow version.

backend/services/workflows/app/processor/nats.go
bahaa-ghazal: uhh i think i forgot to save some changes to this branch while i was testing geminiCLI

alfrunes (approved):

PR #1260, opened 2026-01-12
ci: implement consistent changelog and PR body generation

oldgiova:
Refactored .gitlab/generate_changelog.sh to properly handle different release types (stable, saas, RC) with correct tag filtering. Fixed --ignore-tags pattern escaping that was causing RC/saas tags to appear in stable changelogs.

Added optional PR body generation as 4th parameter, using same range and filtering logic as changelog to ensure consistency between PR description and changelog files.

Added safeguard to truncate oversized PR bodies while preserving markdown structure, preventing GitHub API "Body is too long" errors.

Updated .gitlab-ci.yml to use unified changelog generation for both changelog files and PR bodies, replacing separate git-cliff invocations.

Co-authored with Claude


kjaskiewiczz (approved):
looks good; thanks!

PR #1282, opened 2026-01-21
MEN-9095: email verification and 2fa nudging

mineralsfree:


frontend/src/js/common-ui/ConfirmModal.tsx
```diff
@@ -28,7 +29,7 @@ interface ConfirmModalProps {
   toType: string;
 }
 export const ConfirmModal = (props: ConfirmModalProps) => {
-  const { close, onConfirm, className = '', toType, header, description, open, maxWidth = 'xs' } = props;
+  const { close, onConfirm, className = '', toType, header, description, open, maxWidth = 'xs', confirmButtonText = '' } = props;
```
mzedel:
```suggestion
  const { close, onConfirm, className = '', toType, header, description, open, maxWidth = 'xs', confirmButtonText = 'Confirm' } = props;
```
in light of the `{confirmButtonText || 'Confirm'}` further down

frontend/src/js/components/dashboard/Dashboard.tsx
```diff
@@ -74,9 +97,45 @@ export const Dashboard = () => {
...
+            Your Mender account is not using recommended security settings. Improve your account security in{' '}
+            <Link to="/settings/my-profile" color="inherit">
+              My profile
+            </Link>
+            .
+          </div>
+          <IconButton onClick={dismissAlert}>
+            <CloseIcon />
+          </IconButton>
+        </Alert>
```
mzedel:
```suggestion
        <Alert
          className="margin-bottom-small margin-right"
          onClose={dismissAlert}
          severity="warning"
        >
          <>
            Your Mender account is not using recommended security settings. Improve your account security in{' '}
            <Link to="/settings/my-profile" color="inherit">
              My profile
            </Link>
            .
          </>
        </Alert>
```
afaict that would be the mui standard dismiss setup for the alert (and I haven't tried the classes, but I would assume that's how it would work - especially with the `div` removed that would likely be enough)

---
tried it now, works with the trimmed down version, just had to add a `margin-right` + the alert goes `full-width` by default 👌

frontend/src/js/components/dashboard/Dashboard.tsx
```diff
@@ -74,9 +97,45 @@ export const Dashboard = () => {
...
+  };
+
+  useEffect(() => {
+    if (currentUser && !has2FA && canHave2FA && userSettingInitialized) {
+      const today = dayjs();
+      const dismissedMonthAgo = today.diff(secDismissedTimestamp, 'months');
+      if (secDismissedTimestamp === null || dismissedMonthAgo > 0) {
+        setShowSecurityAlert(true);
+      }
+    }
```
mzedel: considering the dependency on the different aspect of the store state and `secDismissedTimestamp` not being used anywhere else - would this not be better in a selector?

frontend/src/js/components/settings/user-management/twofactorauth-steps/AuthSetup.tsx
```diff
@@ -47,71 +61,55 @@ export const AuthSetup = ({ currentUser, handle2FAState, has2FA, onClose, qrImag
...
-            or{' '}
-            <a href="https://support.google.com/accounts/answer/1066447" target="_blank" rel="noopener noreferrer">
-              Google Authenticator
-            </a>
-            .
-          </li>
-          <li>Scan the QR code on the right with the authenticator app you just downloaded on your device.</li>
+      <div className="flexbox">
+        <ol className={classes.orderedList}>
+          <li className="margin-top-none margin-bottom-small">Download a third party authentication app such as Authy or Google authenticator.</li>
```
mzedel: I assume the instructions should be wrapped in `Typography` as well 😬

frontend/src/js/components/settings/user-management/twofactorauth-steps/AuthSetup.tsx
```diff
@@ -47,71 +61,55 @@ export const AuthSetup = ({ currentUser, handle2FAState, has2FA, onClose, qrImag
...
-              Type in the generated code in the input field below and click Verify.
-              {validated2fa ? (
-                <div className="flexbox space-between centered margin-top margin-right margin-bottom" style={{ justifyContent: 'flex-end' }}>
-                  <CheckCircleIcon className="green" />
-                  <h3 className="green margin-left-small" style={{ textTransform: 'uppercase' }}>
-                    Verified
-                  </h3>
+              Once the code has been scanned, type in the generated code from the app to verify.
+              <Form
+                className="margin-text-small"
```
mzedel: pre-existing, but I only saw this now: that class doesn't exist 🫣

frontend/src/js/components/settings/user-management/SelfUserManagement.tsx
```diff
@@ -79,21 +90,71 @@ export const SelfUserManagement = () => {
     dispatch(saveUserSettings({ mode: newMode }));
   };
 
+  const startVerification = () => {
+    dispatch(verifyEmailStart()).then(() => setConfirmationShown(true));
```
mzedel:
```suggestion
    dispatch(verifyEmailStart()).unwrap().then(() => setConfirmationShown(true));
```
I think...

frontend/src/js/components/settings/user-management/SelfUserManagement.tsx
```diff
@@ -79,21 +90,71 @@ export const SelfUserManagement = () => {
...
+  };
   const handlePass = () => setEditPass(toggle);
 
+  useEffect(() => {
+    if (currentUser.email && !currentUser.verified) {
+      setNeedsVerification(true);
+    }
+  }, [currentUser.email, currentUser.verified, setNeedsVerification]);
```
mzedel: More general: effects are typically be located below the local state definitions, to prevent the number of effects causing problems if a component were to exit early (due to mismatched numbers of effects running during different renders) + regular state setters are not needed in the dependency array

Here: the effect is not needed at all, as the state is only tied to the `currentUser` attributes, thus:

```suggestion
  const needsVerification = currentUser.email && !currentUser.verified;
```

frontend/src/js/components/settings/user-management/SelfUserManagement.tsx
```diff
@@ -79,21 +90,71 @@ export const SelfUserManagement = () => {
...
+          slotProps={{
+            message: {
+              className: 'flexbox center-aligned space-between full-width padding-none'
+            }
+          }}
+        >
+          <Typography>{showNotice}</Typography>
+          <IconButton onClick={() => setShowNotice('')}>
+            <CloseIcon />
+          </IconButton>
```
mzedel: similar to the definition on the Dashboard - the `onClose` prop sets the icon & alignment for us already 👌

frontend/src/js/components/settings/user-management/TwoFactorAuthSetup.tsx
```diff
@@ -106,21 +109,45 @@ export const TwoFactorAuthSetup = () => {
...
+        ))}
+      <ConfirmModal
+        open={confirmDisable}
+        description="Are you sure you want to turn off 2FA? This will make your account less secure."
+        header="Disable two-factor authentication"
+        close={() => setConfirmDisable(false)}
+        confirmButtonText="Disable 2FA"
+        onConfirm={onToggle2FAClick}
+      />
+      {showEmailVerification && <EmailVerification email={currentUser.email} verifyEmailStart={() => dispatch(verifyEmailStart())} />}
```
mzedel:
```suggestion
      {showEmailVerification && <EmailVerification email={currentUser.email} verifyEmailStart={() => dispatch(verifyEmailStart()).unwrap()} />}
```
I think... (to take the burden of knowing this is a thunk off the consuming component)

mzedel (changes requested):
Looks good - some nitpicks, I think only the thunk unwrapping could end up showing unexpected behaviour... + aligning the alerts with the MUI built in functionality would likely save a little work down the line..

mzedel (approved):

PR #1290, opened 2026-01-23
fix(iot-manager): Fix re-encrypt to encrypt ALL secrets

alfrunes:


backend/services/iot-manager/store/mongo/datastore_mongo.go
```diff
@@ -259,10 +269,53 @@ func (db *DataStoreMongo) GetIntegrations(
...
 	}
-	if err = cur.All(ctx, &results); err != nil {
-		return nil, errors.Wrap(err, "error retrieving integrations collection results")
+	return cur, nil
+}
+
+func (db *DataStoreMongo) GetIntegrationsIter(
+	ctx context.Context,
+	fltr model.IntegrationFilter,
+) iter.Seq2[*model.Integration, error] {
```
merlin-northern, line 278: nice usage of iterator to not alloc, copu, and return all the results!

backend/services/iot-manager/cmd/reencrypt.go
```diff
@@ -16,42 +16,31 @@ package cmd
...
+	for integration, err := range dataStore.GetIntegrationsIter(
+		ctx, model.IntegrationFilter{Limit: -1},
+	) {
+		if err != nil {
+			return fmt.Errorf("error retrieving integrations: %w", err)
 		}
-		integrations, err := dataStore.GetIntegrations(ctx, filter)
+		l.Infof("Re-encrypting credentials for integration %s", integration.ID)
+		err = dataStore.SetIntegrationCredentials(ctx,
+			integration.ID, integration.Credentials)
```
merlin-northern, line 40: was wondering where does the encryption happens? more like from the nomenclature perspective. does `SetIntegrationCredentials` actually encrypts something?
how do we know that after this operation completes with err=nil the integration still works? how do we know that after the command completes all the integration work?

backend/services/iot-manager/main.go
```diff
@@ -86,8 +91,18 @@ func doMain(args []string) {
 			},
 			{
 				Name:   "re-encrypt",
-				Usage:  "Re-encrypt the secrets using the (new) encryption key",
+				Usage:  helpReencrypt,
 				Action: cmdReencrypt,
+				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "to-key",
+						Usage: "Override for the updated encryption key",
```
merlin-northern, line 99:
```suggestion
						Usage: "Override for the updated encryption key",
						Required: true,
```

backend/services/iot-manager/main.go
```diff
@@ -86,8 +91,18 @@ func doMain(args []string) {
...
+					},
+					cli.StringFlag{
+						Name:  "from-key",
+						Usage: "Override for the from encryption key.",
```
merlin-northern, line 103:
```suggestion
						Usage: "Override for the from encryption key.",
						Required: true,
```

merlin-northern (approved):

backend/services/iot-manager/cmd/reencrypt.go
alfrunes: I had to scratch my head too, but it happens transparently in when decoding (tries primary and fallback key) and encoding BSON (encrypts using the primary key). That's why it looks like we're just reading the document and write the same document back, but it if it was encrypted with the fallback key, it will be re-encrypted using the primary key.

backend/services/iot-manager/main.go
alfrunes: I'm actually more inclined to removing the override option, but it might be more useful in tests.
The keys are taken from the configuration environment and it should not be necessary to specify the key on the command line.

PR #1292, opened 2026-01-24
chore(gui): keep editors open to prevent user input from getting lost

mzedel:


frontend/src/js/components/settings/user-management/UserManagement.tsx
```diff
@@ -112,25 +112,27 @@ export const UserManagement = () => {
     setRemoveDialog(false);
   };
 
-  const submit = (userData, type, id, passwordResetEmail) => {
+  const submit = async (userData, type, id, passwordResetEmail) => {
```
mineralsfree, line 115: maybe 
```
  const submit = async (userData, type, id, passwordResetEmail) => {
    try {
      if (userData) {
        if (id) {
          await props[actions[type]](id, userData).unwrap();
        } else {
          await props[actions[type]](userData).unwrap();
        }
      }
      if (passwordResetEmail) {
        dispatch(passwordResetStart(passwordResetEmail));
      }
      dialogDismiss();
    } catch {
      // error already handled in thunk - leave open
    }
  };
```

To remove remove duplicate logic

mineralsfree (dismissed):

mineralsfree:
@mzedel something happened to dependencies / package-lock I think 🦺

mzedel:
> @mzedel something happened to dependencies / package-lock I think 🦺

argh, sorry for not checking CI before the request @mineralsfree - too many branches in flight 😬

mineralsfree (approved):

PR #1293, opened 2026-01-26
test(inventory): PATCH /tenants/:tenant_id/device/:device_id/attributes

merlin-northern:
test(inventory): PATCH /tenants/:tenant_id/device/:device_id/attributes
    
Ticket: MEN-9143

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -405,6 +405,7 @@ test:backend:integration:
     when: always
     paths:
       - ${GOCOVERDIR}/integration
+      - backend/logs.*
       - backend/tests/integration/tests/logs.*
       - backend/tests/integration/tests/results_integration_*.xml
       - backend/tests/integration/tests/report_integration_*.html
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -17,10 +17,24 @@ from urllib.parse import urlparse
 
 import pytest
 
+from mender_client import ApiClient
 from testutils.infra.mongo import MongoClient
 from testutils.common import wait_until_healthy
 from testutils.infra.container_manager.kubernetes_manager import isK8S
 from testutils.api.client import get_free_tcp_port, wait_for_port
+
+original_deserialize = ApiClient.deserialize
+
+
+def new_deserialize(self, response_text, response_type, content_type):
+    # the generated client does not support the application/jwt content type
+    if content_type == "application/jwt":
+        return response_text
+    return original_deserialize(self, response_text, response_type, content_type)
+
+
+ApiClient.deserialize = new_deserialize
+
 
 wait_until_healthy("backend-tests")
 
--- a/backend/tests/integration/test_inventory_internal.py
+++ b/backend/tests/integration/test_inventory_internal.py
@@ -1,0 +1,155 @@
+# Copyright 2026 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import pytest
+from mender_client import Configuration, ApiClient, api
+from mender_client.models.attribute_v2 import AttributeV2
+import uuid
+import redo
+
+from testutils.api.client import ApiClient as ApiClientOld
+import testutils.api.deviceauth as deviceauth
+
+from testutils.common import (
+    create_user,
+    make_accepted_devices,
+)
+
+from testutils.infra.cli import CliUseradm, CliDeviceauth
+
+
+@pytest.fixture(scope="function")
+def clean_migrated_mongo(clean_mongo):
+    deviceauth_cli = CliDeviceauth()
+    useradm_cli = CliUseradm()
+
+    deviceauth_cli.migrate()
+    useradm_cli.migrate()
+
+    yield clean_mongo
+
+
+class TestPatchAttributes:
+    def test_inventory_patch_attributes_internal(self, clean_migrated_mongo):
+        login = "user-foo@acme.com"
+        password = "correcthorse"
+        user = create_user(login, password)
+
+        configuration = Configuration.get_default()
+        host = configuration.host
+        configuration.username = login
+        configuration.password = password
+        api_client_management = ApiClient(configuration)
+        user_token = api.UserAdministrationManagementAPIApi(
+            api_client_management
+        ).login()
+
+        tenant_token = ""
+        devauthd = ApiClientOld(deviceauth.URL_DEVICES)
+        devauthm = ApiClientOld(deviceauth.URL_MGMT)
+        make_accepted_devices(devauthd, devauthm, user_token, tenant_token, 2)
+
+        configuration.access_token = user_token
+        inventory_management = api.DeviceInventoryManagementAPIApi(
+            api_client_management
+        )
+        for _ in redo.retrier(attempts=3, sleeptime=1):
+            devices = inventory_management.list_device_inventories(
+                _headers={"Authorization": f"Bearer {user_token}"}, per_page=32, page=1
+            )
+            if len(devices) > 0:
+                break
+        device_id_to_update = devices[0].id
+
+        inventory_internal = api.DeviceInventoryInternalAPIApi(
+            ApiClient(Configuration(host="http://inventory:8080"))
+        )
+        ts = [
+            AttributeV2.from_dict(
+                {"name": "mac", "scope": "identity", "value": "de:ad:be:ef:06:12"}
+            )
+        ]
+        inventory_internal.update_inventory_for_a_device_scope_wise(
+            tenant_id="", device_id=device_id_to_update, attribute_v2=ts
+        )
+
+        current_devices = inventory_management.list_device_inventories(
+            _headers={"Authorization": f"Bearer {user_token}"}, per_page=32, page=1
+        )
+
+        found = False
+        for device in current_devices:
+            if device.id == device_id_to_update:
+                found = any(
+                    [
+                        attribute.name == "mac"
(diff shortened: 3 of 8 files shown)


merlin-northern:
@alfrunes  since I am running behind schedule, I am updating the make_devices* functions to the new client separately, but it will take time, so I kindly request review -- this is the best I could do. before planing I should be able to get rid of the ApiClientOld in this test. I do think this should be enough to close the task and make it lovely before tomorrow.

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,130 @@
...
+        tenant_token = ""
+        useradmm = ApiClientOld(useradm.URL_MGMT)
+        devauthd = ApiClientOld(deviceauth.URL_DEVICES)
+        devauthm = ApiClientOld(deviceauth.URL_MGMT)
+        make_accepted_devices(devauthd, devauthm, user_token, tenant_token, 2)
+
+        configuration = Configuration(host="https://traefik")
+        configuration.access_token = user_token
+        configuration.debug = True
+        configuration.verify_ssl = False
```
alfrunes: The default configuration is set in the root conftest.py:
https://github.com/mendersoftware/mender-server/blob/4946901a8c044a2145c278de7a90a03cce5df9be/backend/tests/conftest.py#L40-L45

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,130 @@
...
+        configuration.username = login
+        configuration.password = password
+        api_client_useradm = ApiClient(configuration)
+        host = api_client_useradm.configuration.host
+        credentials = login + ":" + password
+        credentials_bytes = credentials.encode("ascii")
+        user_token = api.UserAdministrationManagementAPIApi(api_client_useradm).login()
+
+        tenant_token = ""
+        useradmm = ApiClientOld(useradm.URL_MGMT)
```
alfrunes: `useradmm` seems to be unused.

backend/tests/integration/conftest.py
```diff
@@ -17,11 +17,25 @@
...
+
+
+def new_deserialize(self, response_text, response_type, content_type):
+    # the generated client does not support the application/jwt content type
+    if content_type == "application/jwt":
+        return response_text
+    return original_deserialize(self, response_text, response_type, content_type)
+
+
+ApiClient.deserialize = new_deserialize
```
alfrunes, line 36: Not a big fan of such workarounds that rely on import order. Could work as a short term, but I think we probably need to export the templates for the clients and maintain our own templates.

merlin-northern: I understand. but you know, this is basically like overriding Marshal/Unmarshal in go :)

backend/tests/integration/test_inventory_internal.py
merlin-northern: right added there, split the commits, used get_default for configure in the test

backend/tests/integration/test_inventory_internal.py
merlin-northern: right. removed unused things, including this one. also added a new test scenario (new attribute add)

.gitlab-ci.yml
```diff
@@ -405,6 +405,7 @@ test:backend:integration:
     when: always
     paths:
       - ${GOCOVERDIR}/integration
+      - backend/logs.*
```
alfrunes, line 408: I think this should be in the tests directory if I read the compose mounts correctly:
```suggestion
      - backend/tests/logs.*
```
I think we also need to adjust the other artifact paths as well :point_down:

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,146 @@
...
+            devices = am.list_device_inventories(
+                _headers={"Authorization": f"Bearer {user_token}"}, per_page=32, page=1
+            )
+            if len(devices) > 0:
+                break
+        device_id_to_update = devices[0].id
+
+        api_client_inventory = ApiClient()
+        api_client_inventory.configuration.host = "http://inventory:8080"
+        a = api.DeviceInventoryInternalAPIApi(api_client_inventory)
```
alfrunes: REQUIRED: This is doing the same mistake as test_health.py
```suggestion
        inventory_internal = api.DeviceInventoryInternalAPIApi(
            ApiClient(Configuration(host="http://inventory:8080"))
        )
```

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,146 @@
...
+        ts = [
+            AttributeV2.from_dict(
+                {"name": "mac", "scope": "identity", "value": "de:ad:be:ef:06:12"}
+            )
+        ]
+        a.update_inventory_for_a_device_scope_wise(
```
alfrunes: :telescope:  :mage:

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,146 @@
...
+        devauthd = ApiClientOld(deviceauth.URL_DEVICES)
+        devauthm = ApiClientOld(deviceauth.URL_MGMT)
+        make_accepted_devices(devauthd, devauthm, user_token, tenant_token, 2)
+
+        configuration = Configuration.get_default()
+        configuration.access_token = user_token
+        configuration.debug = True
+        configuration.verify_ssl = False
+        api_client_inventory_management = ApiClient(configuration=configuration)
+        am = api.DeviceInventoryManagementAPIApi(api_client_inventory_management)
```
alfrunes: What does "am" mean?
```suggestion
        inventory_management = api.DeviceInventoryManagementAPIApi(api_client_inventory_management)
```

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,146 @@
...
+                {"name": "mac", "scope": "identity", "value": "de:ad:be:ef:06:12"}
+            )
+        ]
+        a.update_inventory_for_a_device_scope_wise(
+            tenant_id="", device_id=device_id_to_update, attribute_v2=ts
+        )
+
+        current_devices = am.list_device_inventories(
+            _headers={"Authorization": f"Bearer {user_token}"}, per_page=32, page=1
+        )
```
alfrunes: Why not get the device inventory by ID?

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,146 @@
...
+
+        found = False
+        for device in current_devices:
+            if device.id == device_id_to_update:
+                for attribute in device.attributes:
+                    if attribute.name == "mac":
+                        assert attribute.value.actual_instance == "de:ad:be:ef:06:12"
+                        assert attribute.scope == "identity"
+                        found = True
```
alfrunes: Can be compressed to:
```python
any(
    [
        attribute.name == "mac"
        and attribute.value.actual_instance == "de:ad:be:ef:06:12"
        and attribute.scope == "identity"
        for attribute in device.attributes
    ]
)
```

alfrunes (changes requested):
Thanks for spotting and fixing the global configuration change in test_health.py
There is one more instance in the new test that needs to be updated. Apart from that I dropped a few suggestions here and there.

.gitlab-ci.yml
merlin-northern: it is not in the tests directory:

Screenshot 2026-01-27 at 11 29 33

alfrunes: Sorry, I misunderstood and thought it was a pytest artifact, but it is from our own script
https://github.com/mendersoftware/mender-server/blob/7705d4491ba7f1bc00a937cd4220ccffd4f51fb5/backend/tests/run-integration.bash#L270-L274

backend/tests/integration/test_inventory_internal.py
merlin-northern: python master! changed. although it is longer and more unreadable :>

backend/tests/integration/test_inventory_internal.py
merlin-northern: because I like to iterate? are you adamant here? perhaps, you know, 🥺

backend/tests/integration/test_inventory_internal.py
merlin-northern: ante meridiem? fixed.

backend/tests/integration/test_inventory_internal.py
merlin-northern: 😘  fixed.

backend/tests/integration/test_inventory_internal.py
merlin-northern: you are right. it is a total mess.  fixed.

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,159 @@
...
+class TestPatchAttributes:
+    def test_inventory_patch_attributes_internal(self, clean_migrated_mongo):
+        login = "user-foo@acme.com"
+        password = "correcthorse"
+        user = create_user(login, password)
+
+        configuration = Configuration.get_default()
+        host = configuration.host
+        configuration.debug = True
+        configuration.verify_ssl = False
```
alfrunes: Can't we remove this since it's already set in conftest.py?
```suggestion
```

backend/tests/integration/test_inventory_internal.py
```diff
@@ -0,0 +1,159 @@
...
+        make_accepted_devices(devauthd, devauthm, user_token, tenant_token, 2)
+
+        configuration = Configuration.get_default()
+        configuration.access_token = user_token
+        configuration.debug = True
+        configuration.verify_ssl = False
+        api_client_inventory_management = ApiClient(configuration=configuration)
+        inventory_management = api.DeviceInventoryManagementAPIApi(
+            api_client_inventory_management
+        )
```
alfrunes: Can reuse the same api client as useradm.
```suggestion
        configuration.access_token = user_token
        inventory_management = api.DeviceInventoryManagementAPIApi(
            api_client_useradm
        )
```

alfrunes (dismissed):

backend/tests/integration/test_inventory_internal.py
merlin-northern: sure. fixed.

backend/tests/integration/test_inventory_internal.py
merlin-northern: sure. fixed.

alfrunes (approved):

PR #1296, opened 2026-01-26
ci: added jq fall back when no sync pr is found

mzedel:
not 100% sure if that's the reason for the failure to sync...


.gitlab/merge-enterprise.yml
```diff
@@ -36,7 +36,7 @@ merge-to-enterprise:
         --header "Authorization: Bearer $GITHUB_BOT_TOKEN_REPO_FULL" \
         --header "Accept: application/vnd.github+json" \
         --header "X-GitHub-Api-Version: 2022-11-28" \
-        -O- | jq -r '.[]' > pull.json
+        -O- | jq -r 'first // {}' > pull.json
```
alfrunes: If `first` fails, I think we can suppress the error and return the empty object.
```suggestion
        -O- | jq -r 'first? // {}' > pull.json
```

alfrunes (approved):

PR #1297, opened 2026-01-26
ME-614 fix(gui): fixed an issue that would prevent importing config data into a partly populated editor

mzedel:


frontend/src/js/common-ui/forms/KeyValueEditor.tsx
```diff
@@ -171,7 +171,12 @@ export const KeyValueEditor = ({ disabled, errortext, initialInput = {}, inputHe
       ? Object.entries(initialInput).map(([key, value]) => ({ helptip: inputHelpTipsMap[key.toLowerCase()], key, value }))
       : [{ ...emptyInput }]
   };
-  const [initialValues] = useState(defaultValues);
+  const [initialValues, setInitialValues] = useState(defaultValues);
+
+  useEffect(() => {
+    setInitialValues(defaultValues);
```
mineralsfree, line 183: When I use the file upload config, the changes in the file are still not reflected in the inputs, the changes are only available after the config save... Is that a different isssue?

mineralsfree (approved):

PR #1300, opened 2026-01-26
Fix/multi device links

mzedel:


frontend/src/js/components/devices/DeviceGroups.test.tsx
```diff
@@ -35,33 +44,22 @@ const preloadedState = {
...
 
-vi.mock('react-router-dom', async importOriginal => {
-  const actual = await importOriginal();
-  return {
-    ...actual,
-    useLocation: vi.fn(),
-    useSearchParams: vi.fn()
-  };
-});
+const renderWithRouter = (ui: React.ReactElement, { route = '/', preloadedState: state = preloadedState } = {}) => {
```
mineralsfree, line 47: Maybe this deserves to be in a tesing package? 👍

mineralsfree (approved):
Looks good

frontend/src/js/components/devices/DeviceGroups.test.tsx
mzedel: Indeed, we already have a couple of similar setups... will follow up in nt-gui 👌

PR #1309, opened 2026-01-27
MEN-9060 - feat(gui): enabled dark mode support for saml + other editors

mzedel:


mineralsfree:
needs rebase

frontend/src/js/components/devices/device-details/DeviceTwin.tsx
mineralsfree: add theme here?

mineralsfree (changes requested):

frontend/src/js/common-ui/CodeEditor.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const editorThemeName = `mender-${muiTheme.palette.mode}${isReadOnly ? '-ro' : ''}`;
+
+  const defineEditorTheme = useCallback(
+    (monaco: Monaco) => {
+      monacoRef.current = monaco;
+      monaco.editor.defineTheme(editorThemeName, {
+        base: isDarkMode ? 'vs-dark' : 'vs',
+        inherit: true,
+        rules: [],
+        colors: { 'editor.foreground': makeColorMonacoSafe(muiTheme.palette.text.primary) }
```
mineralsfree, line 90: Will it still work if `text.primary` changes from rgb to hex in theme definition?

mineralsfree (approved):
only one concern, looks very nice

frontend/src/js/common-ui/CodeEditor.tsx
mzedel: based on the `rgbToHex` util from mui here: https://github.com/mui/material-ui/blob/master/packages/mui-system/src/colorManipulator/colorManipulator.js#L166-L168 - yes... the _making monaco safe_ adds a simple check to catch the shortform hex (`#fff`) as that was too fancy for the editor theming 😬

PR #1313, opened 2026-01-28
test(gui): adjusted e2e test to a new device limit widget

mineralsfree:


mzedel (dismissed):

mzedel:
@mineralsfree the test id needs to be set in the snapshots too 😬

mzedel (approved):

PR #1314, opened 2026-01-28
fix(gui): fixed the email verification flow for oauth2 users

mineralsfree:


frontend/src/js/components/settings/user-management/SelfUserManagement.tsx
```diff
@@ -95,6 +96,16 @@ export const SelfUserManagement = () => {
       .then(() => setConfirmationShown(true));
   };
   const handlePass = () => setEditPass(toggle);
+  useEffect(() => {
+    if (activationCode) {
+      dispatch(verifyEmailComplete(activationCode))
+        .unwrap()
+        .catch(() => {
+          console.log('Error verifying email');
+        })
```
mzedel, line 105: users should get the snackbar already, but OTOH we could add this to the `commonErrorHandler` and get a clearer indication through sentry this way...

mzedel (dismissed):

frontend/src/js/components/settings/user-management/SelfUserManagement.tsx
mineralsfree: I think, it's already covered in the thunk, removing the `console.log`

mzedel (approved):

PR #1346, opened 2026-02-02
test: welcome go based integration tests

merlin-northern:
test: welcome go based integration tests

Ticket: MEN-8571


merlin-northern:
@bahaa-ghazal you are missing some context, as I did not demo it to you -- this brings the new framework of backend integration tests. I want to write the tests for device tiers in go. so I need it here to bring it into the ent.

backend/tests/runner/scripts/run.sh
alfrunes, line 1: I was hoping we could get rid of boilerplate scripts like this.

backend/go.mod
```diff
@@ -80,23 +86,31 @@ require (
...
 	github.com/google/go-tpm v0.9.8 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.18.3 // indirect
 	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
 	github.com/klauspost/pgzip v1.2.6 // indirect
+	github.com/labstack/gommon v0.4.2 // indirect
 	github.com/leodido/go-urn v1.4.0 // indirect
+	github.com/mailru/easyjson v0.7.7 // indirect
```
alfrunes: I don't want to get too political, but those are quite some additional dependencies...
And you might want to give this one a try in your favorite search engine.

backend/tests/runner/client/client.go
alfrunes, line 1: This is going to have severe consequences on our coverage economy.

alfrunes: Also, can we please add a `_gen.go` suffix to all generated files?

alfrunes (commented):
Some initial thoughts...

Pros:
* Sticking to Golang means one less language / one less technicality to learn
* Less dependencies (assuming we get rid of Python from our test environments)

Cons:
* Lack of standardized framework / clear structure
  * No clear way of how to create reusable fixtures
  * Maybe we should explore existing frameworks for Golang?
* Python has a lower entry bar making integration tests a good starting point for onboarding / getting to know the product
* An interpreted language is arguably easier to debug

backend/go.mod
merlin-northern: this is oapi-codegen dep I have not added it. Edit: I have no idea where it came from actually, I guess it was after I generated the client from oapi-codegen. Edit2: oh yes I was right the first time:
```
backend/vendor/github.com/go-openapi/swag/yaml.go:27: "github.com/mailru/easyjson/jwriter"
```
EditFinal:
```
#go mod why -m github.com/mailru/easyjson
# github.com/mailru/easyjson
github.com/mendersoftware/mender-server/tests/runner/client
github.com/getkin/kin-openapi/openapi3
github.com/perimeterx/marshmallow
github.com/mailru/easyjson/jlexer
```

backend/tests/runner/scripts/run.sh
merlin-northern: I understand, but there is a purpose for this stage of the tests.  treat as temporary -- I need something to start with.

backend/go.mod
merlin-northern: fixed.

backend/tests/runner/client/client.go
merlin-northern: I switched to the well known `openapitools/openapi-generator-cli` lovely generator: 
```
docker run --rm -v ${PWD}/backend:/backend openapitools/openapi-generator-cli generate -i /backend/docs/api/dist/openapi.yaml  -g go -o /backend/pkg/apis/client
```
I currently do not know how to change all the suffixes, could you swallow this as is, since I will be fixing things around it anyways.

frodeha:
I'm very, very positive and interested in (at least starting to) move towards writing integration/acceptance tests in Go so a massive "hurray!!!" for the initiative and the PR, Peter!

Before I go further I would like to chime in with my opinions who are (maybe a bit contrary) to Alf's above (I largely agree with the Pro's so I'll focus on the Con's). I would argue that Python (or more specifically pytest (or at least our pytests)) is actually quite difficult to wrangle with - especially if you've done little or no Python before (which is the case with me). 

The sheer amount of "magic disguised as convenience" in pytest combined with the dynamic nature of Python makes it quite hard to understand what's going on indeed (speaking from experience), and is as such (imo) both less easy to get into and more difficult to debug. It's actually one of the few things I do consult AI about from time to time  - which is not exactly a strength in my view. I can of course respect that opinions on this can vary and will depend on a number of factors, so I'll just leave mine (subjective as they are) at that.

When it comes to the PR itself though I have some comments/suggestions. I realize that this is a starting point so please do treat these as suggestions for further improvements should we continue down this path (which I hope we will!).

**Test organization**
I don't think we should roll our own setup like this. It's not compatible with the `-run` argument of `go test` (I think) and there are preexisting solutions to this already (so we don't need to make another). In the past I've had good experience using the [Suite package from testify](https://pkg.go.dev/github.com/stretchr/testify/suite) to structure exactly Go integration tests and share setup/tear-down and other functionality (or fixtures in `pytest` lingo).  

I took the liberty of a quick and dirty refactor of the tests in this PR to use the Suite package as a poc/demo. It's not necessarily very well thought through and I'm not sure it produces green results, but it might illustrate what I'm thinking at least. 

<details>
<summary>Expand for code 👨‍💻</summary>

```go
// ---------------
// ./integration_test.go

type IntegrationTests struct {
	suite.Suite
	settings *TestSettings
}

func (i *IntegrationTests) SetupSuite() {
	// TODO: Setup compose environment here?

    // copy/paste from setup in this PR
	var (
		serverURL       string
		username        string
		password        string
		usernamePattern string
		passwordPattern string
	)

	flag.StringVar(&serverURL, "server-url", "https://localhost", "Mender Server URL")
	flag.StringVar(&username, "username", "demo@mender.io", "Mender Server login username")
	flag.StringVar(&password, "password", "password123", "Mender Server user password")
	// the patterns are used to inform the test about what users and passwords were created during the setup phase
	// that is: what usernames and password had been created before the go test started
	flag.StringVar(&usernamePattern, "username-pattern", "", "Mender Server login username pattern e.g: test-user-%d@mender.io")
	flag.StringVar(&passwordPattern, "password-pattern", "", "Mender Server user password pattern e.g.: test-user-password-%d")
	flag.Parse()

	if len(passwordPattern) > 0 && len(usernamePattern) > 0 {
		username = strings.ReplaceAll(usernamePattern, "%d", "1")
		password = strings.ReplaceAll(passwordPattern, "%d", "1")
	}
	config := openapi.NewConfiguration()
	config.Host = ServerURL
	config.Scheme = "https"
	config.HTTPClient = &http.Client{
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true,
			},
		},
	}

	i.settings = &TestSettings{
		ServerURL:     ServerURL,
		Username:      username,
		Password:      password,
		configuration: config,
		client:        openapi.NewAPIClient(config),
	}

	// Other stuff?
}

func (i *IntegrationTests) TearDownSuite() {
	// TODO: Tear down compose environment
}

func TestIntegration(t *testing.T) {
	suite.Run(t, new(IntegrationTests))
}

//----------------
// ./user_management_test.go

type UserManagement struct {
	suite.Suite // Or we can embed the IntegrationTests suite maybe? :shrug:
	settings *TestSettings
}

func (u *UserManagement) TestLogin() {
	ctx := u.login()
	v := ctx.Value(openapi.ContextAccessToken)
	token, ok := v.(string)
	assert.True(u.T(), ok, "access token in context is of invalid type")
	assert.NotEmpty(u.T(), token)
}

func (u *UserManagement) TestSelf() {
	ctx := u.login()
	body, r, err := u.useradm().ShowOwnUserData(ctx).Execute()

	require.NoError(u.T(), err)
	require.NotNil(u.T(), r)
	assert.Equal(u.T(), http.StatusOK, r.StatusCode)
	assert.Equal(u.T(), u.settings.Username, body.Email)

	u.T().Logf("test passed with user data: %+v\n", body)
}

func (u *UserManagement) TestMe() {
	ctx := u.login()
	body, r, err := u.useradm().ShowMyUserSettings(ctx).Execute()
	require.NoError(u.T(), err)
	require.NotNil(u.T(), r)
	require.NotNil(u.T(), body)
	assert.Equal(u.T(), http.StatusOK, r.StatusCode)
	u.T().Logf("test passed with data: %+v\n", body)
}

func (u *UserManagement) TestRemove() {
	ctx := u.login()
	r, err := u.useradm().RemoveUser(ctx, "id").Execute()
	require.NoError(u.T(), err)
	require.NotNil(u.T(), r)
	assert.Equal(u.T(), http.StatusNoContent, r.StatusCode)
}

func (u *UserManagement) login() context.Context {
	if u.settings.jwt == "" {
		auth := openapi.BasicAuth{
			UserName: u.settings.Username,
			Password: u.settings.Password,
		}

		ctx := context.WithValue(u.T().Context(), openapi.ContextBasicAuth, auth)
		token, r, err := u.useradm().Login(ctx).Execute()

		require := require.New(u.T())
		require.NoError(err)
		require.NotNil(r)
		require.NotEmpty(token)
		require.Equal(http.StatusOK, r.StatusCode)

		u.settings.jwt = token
	}

	return context.WithValue(
		u.T().Context(),
		openapi.ContextAccessToken,
		u.settings.jwt,
	)
}

func (u *UserManagement) useradm() *openapi.UserAdministrationManagementAPIAPIService {
	return u.settings.client.UserAdministrationManagementAPIAPI
}

// TestUserManagement registers the UserManagement suite on the main IntegrationTest suite.
func (i *IntegrationTests) TestUserManagement() {
	suite.Run(i.T(), &UserManagement{settings: i.settings})
}
```
</details>

I'd like to re-iterate that this is the 30 minute attempt to describe a concept - not a final solution.


**Boilerplate scripts**
As Alf, I also would love to get rid of the boilerplate setup script(s). To that end I've had a look at the Docker Compose Go SDK that was released recently as part of [Docker Compose v5](https://docs.docker.com/compose/compose-sdk/) and it looks like we can do all of [this in plain Go](https://pkg.go.dev/github.com/docker/compose/v5@v5.0.2/pkg/api#Compose) now. I was thinking that we could do the required docker-compose setup and tear-down in `SetupSuite()` and `TearDownSuite()` of the IntegrationTests suite above (if we go with that pattern). I haven't tried this myself, but I'd say it's definitely worth a look (and I'd be very happy to have a look myself).

All that being said, I like the general direction towards Go integration tests a lot, so big 👍 from me!

bahaa-ghazal (dismissed):
For me this looks good. The golang tests look clearer and simpler then the current once. My only concerns is that with these go tests, we might need to create many testing tools to simulate what we have currently. I would rather use something that already exists and know it works fine, like we have currently, instead of putting large workload to simulate what these tools do (like workflows 🙃).  
Ofc, I'm just guessing  🤷‍♂️, so I would like that we test it with writing couple of complex integration tests to make sure it will work fine before we fully replace the python tester.
I will approve this pr, but it will be nice if we got more more feedback from others.

merlin-northern:
> For me this looks good. The golang tests look clearer and simpler then the current once. My only concerns is that with these go tests, we might need to create many testing tools to simulate what we have currently. I would rather use something that already exists and know it works fine, like we have currently, instead of putting large workload to simulate what these tools do (like workflows 🙃). Ofc, I'm just guessing 🤷‍♂️, so I would like that we test it with writing couple of complex integration tests to make sure it will work fine before we fully replace the python tester. I will approve this pr, but it will be nice if we got more more feedback from others.

thank you! I adopted to the github.com/stretchr/testify/suite package and I think it looks also fun. gh dismissed your review, I am sorry for that.

merlin-northern:
> I'm very, very positive and interested in (at least starting to) move towards writing integration/acceptance tests in Go so a massive "hurray!!!" for the initiative and the PR, Peter!
>... 
> All that being said, I like the general direction towards Go integration tests a lot, so big 👍 from me!

thank you! and amazin suggestion. so I moved to github.com/stretchr/testify/suite making some design decisions of my own, based on your comment. I really would like an approval because I will never make it on time otherwise.

frodeha (approved):
Let's give it a try! 👍 Looking forward to seeing where this goes (hopefully someplace nice and type safe 😁)

PR #1350, opened 2026-02-03
MEN-9222: Apply size limits on deplyoment and device group names

bahaa-ghazal:


backend/pkg/rules/patterns.go
```diff
@@ -0,0 +1,15 @@
+package rules
+
+import (
+	"regexp"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+)
+
+var (
+	groupNameRegex = regexp.MustCompile("^[A-Za-z0-9_-]*$")
```
alfrunes, line 10: I suppose we don't allow empty group names...
```suggestion
	groupNameRegex = regexp.MustCompile("^[A-Za-z0-9_-]+$")
```

backend/pkg/rules/patterns.go
```diff
@@ -0,0 +1,15 @@
...
+
+	deviceGroupPattern = validation.Match(groupNameRegex).Error(
+		"Group name can only contain: upper/lowercase " +
```
alfrunes:
```suggestion
		"group name can only contain: upper/lowercase " +
```

backend/pkg/rules/sizes.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+)
+
+var (
+	legacyDeviceGroupNameSize = validation.Length(0, 1024)
+	deviceGroupNameSize       = validation.Length(0, 256)
+
+	deploymentNameSize = validation.Length(0, 256)
```
alfrunes, line 11:
```suggestion
	legacyDeviceGroupNameSize = validation.Length(1, 1024)
	deviceGroupNameSize       = validation.Length(1, 256)

	deploymentNameSize = validation.Length(1, 256)
```

backend/pkg/rules/validators.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func DeviceGroupeName(value interface{}) error {
+	s := fmt.Sprintf("%v", value)
+	return validation.Validate(s,
+		validation.Required,
+		deviceGroupNameSize,
+		deviceGroupPattern,
+	)
+}
```
alfrunes: We should make a type assertion here. If you pass a pointer to an invalid string, it will validate as correct under these conditions.
```suggestion
func DeviceGroupeName(value any) error {
	if s, ok := value.(string); ok {
		return validation.Validate(s,
			validation.Required,
			deviceGroupNameSize,
			deviceGroupPattern,
		)
	}
	return fmt.Errorf("invalid type %T for device group", value)
}
```

backend/pkg/rules/validators.go
```diff
@@ -0,0 +1,33 @@
...
+
+func DeploymentName(value interface{}) error {
+	s := fmt.Sprintf("%v", value)
+	return validation.Validate(s,
+		validation.Required,
+		deploymentNameSize,
+	)
+}
```
alfrunes: Same for these two rules.

alfrunes (changes requested):

backend/pkg/rules/patterns.go
bahaa-ghazal: `validation.Required` makes sure that groupName is not empty

backend/pkg/rules/sizes.go
bahaa-ghazal: That does not actually do anything. Even if we have `validation.Length(1, 1024)` and we pass in something empty it will pass. if something is empty the validatior does not consider the length and `validation.Length()` will pass anyways. That way we need `validation.Required`.

backend/pkg/rules/validators.go
bahaa-ghazal: Yeah... I have done this cheat because the value can be either string or `GroupName`

backend/pkg/rules/validators.go
```diff
@@ -0,0 +1,42 @@
...
+		return false
+	}
+
+	return v.Kind() == kind
+}
+
+func DeviceGroupName(value interface{}) error {
+	if !HasUnderlyingType(value, reflect.String) {
+		return fmt.Errorf("invalid type %T for device group", value)
+	}
```
alfrunes, line 23: Why not use built-in type assertion?
```suggestion
	if _, ok := value.(string); ok {
		return fmt.Errorf("invalid type %T for device group", value)
	}
```

bahaa-ghazal: As I said sometimes this value could be of type `GroupName` which will not work

backend/pkg/rules/sizes.go
alfrunes: Ahh, ok. Thanks for the clarification :ok_hand:

backend/pkg/rules/validators.go
```diff
@@ -0,0 +1,42 @@
...
+
+	return validation.Validate(value,
+		validation.Required,
```
alfrunes, line 26:
```suggestion
		validation.Required.Error("group name is required"),
```

alfrunes: In the general case, whether a field is required might be contextual.
I like the shared validation rules you've created, but I'm wondering whether it would make sense to split it into individual rules (e.g. one for validating the character set and length of a specific type, etc.).

alfrunes (approved):
Looks fine to me. But I wonder if we should leave out the Required rule from the shared rules as it is contextual (e.g. in a `PATCH` API, you'd probably don't want attributes to be required although they might be required when creating the object).

backend/pkg/rules/validators.go
bahaa-ghazal: yeah I have tried removing the Required out of the rule set, but it made sens  to me to package everything you need to one set of rules. 
We can also make the individual rules public for more custom rules.

PR #1351, opened 2026-02-04
MEN-9277: backend based tiers in subscripiton page

mineralsfree:
Needs https://github.com/NorthernTechHQ/nt-gui/pull/519


frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -54,50 +58,27 @@ const useStyles = makeStyles()(() => ({
     }
   }
 }));
-type DeviceTypeId = 'micro' | 'standard';
 
-type DeviceTypeConfig = {
-  id: string;
+//With hope in enum from backend types
+type DeviceTypeId = string;
```
mzedel, line 63: we now have https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/types/src/types.gen.ts#L3388-L3397 after https://github.com/mendersoftware/mender-server-enterprise/pull/966 👌

frontend/src/js/components/subscription/SubscriptionSummaryItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  isPreviewLoading: boolean;
+  previewPriceItem: PlanPreviewPriceItem;
+  summaryLabel: string;
+};
+export const SubscriptionSummaryItem = (props: SubscriptionSummaryItemProps) => {
+  const { addons, previewPriceItem, isPreviewLoading, summaryLabel } = props;
+  const disabled = previewPriceItem?.price === 0;
+  const textColor = disabled ? 'text.disabled' : 'text.primary';
+  return (
+    previewPriceItem && (
```
mzedel, line 34:
```suggestion
  if (!previewPriceItem) {
    return null;
  }
  return (
```

just to avoid the extra nesting, but fine as is too. OTOH this would also allow you to destructure the `previewPriceItem` afterwards to avoid some of the checks...
```
const { addons = [], price = 0, quantity } = previewPriceItem;
```

frontend/src/js/components/subscription/SubscriptionSummaryItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const NumberSkeleton = () => <Skeleton width={35} height={26} />;
+
+type SubscriptionSummaryItemProps = {
+  addons: AvailableAddon[];
+  isPreviewLoading: boolean;
+  previewPriceItem: PlanPreviewPriceItem;
+  summaryLabel: string;
+};
+export const SubscriptionSummaryItem = (props: SubscriptionSummaryItemProps) => {
```
mzedel, line 29: thanks for following up with this 🙏!

frontend/src/js/components/subscription/SubscriptionPage.test.tsx
```diff
@@ -12,17 +12,33 @@
...
+export const standardDeviceTier: Partial<DeviceTypes> = {
+  standard: { id: 'standard', stripeProductName: 'mender_standard', label: 'Standard devices', summaryLabel: 'Standard', tooltipId: 'standardDevice' }
+};
+export const microDeviceTier: Partial<DeviceTypes> = {
+  micro: {
+    id: 'micro',
+    stripeProductName: 'mender_micro',
+    label: 'Micro devices',
+    summaryLabel: 'Micro',
+    tooltipId: 'mcuDevice',
```
mzedel:
```suggestion
    tooltipId: 'microDevice',
```

not that it's very really relevant for the test here

frontend/src/js/components/subscription/SubscriptionSummaryItem.test.tsx
```diff
@@ -0,0 +1,49 @@
+// Copyright 2025 Northern.tech AS
```
mzedel:
```suggestion
// Copyright 2026 Northern.tech AS
```

frontend/src/js/components/subscription/SubscriptionSummaryItem.tsx
mzedel: additionally, you have the check in the `.map`:
```
!!previewPrice.items[id] && (
  <SubscriptionSummaryItem
    ...
    previewPriceItem={previewPrice.items[id]}
```
(just for the record: that would be yet another good use for reduce 😜) in `SubscriptionSummary` - so this should never be falsy right?

mzedel (dismissed):
Very well done! some tiny nitpicks, but nothing really worth chasing (except maybe the copyright year for legal reasons 😬)

mzedel (approved):

PR #1354, opened 2026-02-05
feat(gui): added alert for RBAC not available in non-enterprise plans

mineralsfree:
Ticket: MEN-8610

### DIFF
--- a/frontend/src/js/components/settings/user-management/UserForm.tsx
+++ b/frontend/src/js/components/settings/user-management/UserForm.tsx
@@ -13,9 +13,11 @@
 //    limitations under the License.
 import { useEffect, useMemo, useState } from 'react';
 import { useWatch } from 'react-hook-form';
+import { useSelector } from 'react-redux';
 
 import { InfoOutlined } from '@mui/icons-material';
 import {
+  Alert,
   Checkbox,
   Collapse,
   DialogActions,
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/settings/user-management/UserForm.tsx
```diff
@@ -96,37 +100,44 @@ export const UserRolesSelect = ({ currentUser, disabled, onSelect, roles, user }
...
+              <br />
+              Consider adding the <i>{rolesById[rolesByName.readOnly].name}</i> role as well.
+            </FormHelperText>
+          )}
+        </FormControl>
+        <EnterpriseNotification className="margin-left-small" id={BENEFITS.rbac.id} />
+      </div>
+      {!isEnterprise && (
+        <Alert className="margin-top-small" severity="warning">
+          Role-base access control (RBAC) is not available in your current plan. All users will have full administrative access.
```
mzedel:
```suggestion
          Role-based access control (RBAC) is not available in your current plan. All users will have full administrative access.
```

mzedel (changes requested):
tiny bit 🫣

frontend/src/js/components/settings/user-management/UserForm.tsx
mineralsfree: Thanks for the catch! have to double-check what I copy-paste from the figma next time 😄

mzedel (approved):

PR #1357, opened 2026-02-08
MEN-9407 - feat(deployments): aligned deployment filtering w/ ux expectations

mzedel:
to go with https://github.com/NorthernTechHQ/nt-gui/pull/521

fyi @chiachenglu & @michaelntech 

The combination of this and the related nt-gui PR will allow the following:

https://github.com/user-attachments/assets/54db3085-4ba2-4d1b-81a4-c74d804cf379


chiachenglu:
Thanks for the PR. Can you illustrate more the expect behavior with this changes? @mzedel

mzedel:
> Thanks for the PR. Can you illustrate more the expect behavior with this changes? @mzedel

@chiachenglu the closest description would be the comment in here: https://github.com/mendersoftware/mender-server/pull/1357/changes#diff-15c270ef76c97953ee1b794980d20e0dfc995d6d34b8f9ae5a7574b097af2de8R1909-R1911 - the key differences to what we talked about are: 
- the UI will provide the setting for the identity attribute to save the backend an internal request
- I didn't remove the undocumented `artifact_name` from the search just yet, since removals are less likely to go through...

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -1705,6 +1705,37 @@ func ParseDeploymentLookupQueryV1(vals url.Values) (model.Query, error) {
...
+		)
+	}
+	if idAttribute != "" {
+		if idScope == "" {
+			if idAttribute == "name" {
+				idScope = "inventory"
+			} else {
+				idScope = "identity"
+			}
+		}
```
alfrunes, line 1746: This seems to make a bit of complex assumptions, could we simply make the scope required?

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -1705,6 +1705,37 @@ func ParseDeploymentLookupQueryV1(vals url.Values) (model.Query, error) {
...
+			"invalid id_scope %q: must be \"identity\" or \"inventory\"",
+			idScope,
+		)
+	}
+	if idScope == "inventory" && idAttribute != "name" {
+		return query, errors.Errorf(
+			"id_scope \"inventory\" only supports id_attribute \"name\", got %q",
+			idAttribute,
+		)
+	}
```
alfrunes, line 1738: To simplify the validation and make it easier to explain in OAS specs, can we compress these three parameters to one?
For example:
`search=<scope>/<attribute>=<value>`

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -2754,6 +2754,17 @@ func (db *DataStoreMongo) buildDeploymentsQuery(
...
 	}
 
+	// filter by device IDs (matches single-device deployments by name
+	// and multi-device deployments by device_list)
+	if match.DeviceIDs != nil {
+		andq = append(andq, bson.M{
+			"$or": []bson.M{
+				{StorageKeyDeploymentName: bson.M{"$in": match.DeviceIDs}},
+				{StorageKeyDeploymentDeviceList: bson.M{"$in": match.DeviceIDs}},
+			},
```
alfrunes, line 2741: For single device deployments the device list should always be populated.
```suggestion
			StorageKeyDeploymentDeviceList: bson.M{"$in": match.DeviceIDs},
```
Unless there's something I'm missing, can we remove this filter condition?

backend/services/deployments/app/app.go
```diff
@@ -1905,6 +1906,56 @@ func (d *Deployments) LookupDeployment(ctx context.Context,
...
+				Type:      "$eq",
+				Value:     query.SearchText,
+			}},
+		})
+		if err != nil {
+			return nil, 0, errors.Wrap(err,
+				"searching inventory for device")
+		}
+		if len(devices) > 0 {
+			deviceID := devices[0].ID
```
alfrunes, line 1937: This seems a bit unpredictable when multiple devices are matching the search term.

alfrunes (commented):
I'm a bit out of context here, do you have a ticket reference?

I made a few suggestions, but there also needs to be an update to the OAS specifications.

backend/services/deployments/app/app.go
mzedel: That's the point, otherwise we wouldn't know which ID to search for in the target devices - the identity attribute is supposed to be unique enough for the tenant, but if it's not, we could end up showing more wrong deployments...

backend/services/deployments/api/http/api_deployments.go
mzedel: That would change the behaviour of the `search` param - here that extension is only triggered if those are supplied and the `search` failed.

backend/services/deployments/api/http/api_deployments.go
mzedel: That's in line with the identity attribute as present in the UI - so unless that's going to change I would keep this as a safeguard against unintended searches or settings.

mzedel:
> there also needs to be an update to the OAS specifications.

This follows the established practice of not showing the details of that endpoint in line with the search for `artifact_name` (this in particular is both undocumented and unintended)

As to the context, that's the topic we talked about on Friday - the UI offering deployment search (via the otherwise unused `deployment.name`), but the tenant configured device identity attribute won't be searchable, only the device id - this would make this possible without having the backend retrieve the settings from useradm, but instead get it supplied by the UI.

alfrunes:
> This follows the established practice of not showing the details of that endpoint in line with the search for artifact_name (this in particular is both undocumented and unintended)

I was not aware of this "established practice", as you say it was probably unintended. But if I'm not mistaken, the OAS spec [seems to be there](https://github.com/mendersoftware/mender-server/blob/v4.1.0/backend/docs/api/deployments/management_v1.yaml#L42-L46), but the endpoint is marked as deprecated.

> As to the context, that's the topic we talked about on Friday - the UI offering deployment search (via the otherwise unused deployment.name), but the tenant configured device identity attribute won't be searchable, only the device id - this would make this possible without having the backend retrieve the settings from useradm, but instead get it supplied by the UI.

Ahh, ok. Sorry, I did not recall. But still, shouldn't this be tracked in JIRA?

backend/services/deployments/api/http/api_deployments.go
alfrunes: You're right. I was mainly thinking about the documentation part. But how about only merging the scope into the prefix of the `id_attribute` parameter?

mzedel:
> I was not aware of this "established practice", as you say it was probably unintended. But if I'm not mistaken, the OAS spec [seems to be there](https://github.com/mendersoftware/mender-server/blob/v4.1.0/backend/docs/api/deployments/management_v1.yaml#L42-L46), but the endpoint is marked as deprecated.

established practice in so far as the endpoint docs say it matches on `Deployment name or description` - but the query model says [here](https://github.com/mendersoftware/mender-server/blob/main/backend/services/deployments/model/deployment.go#L342) `deployment name and artifact name`. IOW the discrepancy between docs & implementation usually isn't a hurdle 😉.

As to the deprecation notice: apparently that didn't reach the gui, as until yesterday that's what was used - adjusted now via: https://github.com/NorthernTechHQ/nt-gui/pull/543) 👌.

> shouldn't this be tracked in JIRA?

I'm adding a task there next 👌.

backend/services/deployments/api/http/api_deployments.go
mzedel: wouldn't this also assume customers won't have an attribute starting with `identity` or limit their search values (e.g. we have no control how they name deployments).
The explicit query params are supposed to avoid making these assumptions and considering the usage in only the management endpoint it shouldn't cause any significant strain AFAICT.

backend/services/deployments/store/mongo/datastore_mongo.go
alfrunes: This query is still filtering on deployment name, is this intentional?

alfrunes (commented):
I'd prefer having a `Ticket` reference upfront in the commit footer of the first commit.
Otherwise looks good to me.

mzedel:
Thank you for taking another look!

> I'd prefer having a `Ticket` reference upfront in the commit footer of the first commit. Otherwise looks good to me.

Done 👌 and for:

> ... still filtering on deployment name, is this intentional?

yes, as I wasn't sure if the device list is always populated (not all of my old local deployments have it), but the name is (but that again doesn't work well with multiple device ids 😕)

alfrunes (approved):
ACK, thanks for clarifying. Let's proceed with it as is :+1:

PR #1372, opened 2026-02-09
refactor(useradm): Properly enforce required useradm CLI flags

LudvigAnderson:
Previously, the useradm commands did not have required flags marked as such. If missing multiple flags, the user would only be notified of one. Now, all mising flags will be shown immediately to the user.

Changelog: useradm commands now properly inform the user of missing required flags.


alfrunes (approved):
Thank you!

PR #1374, opened 2026-02-10
MEN-8459 - deployments page design alignment (+ dashboard deployment widget and deployment report related changes)

mzedel:


frontend/src/js/common-ui/forms/Filters.tsx
```diff
@@ -14,29 +14,31 @@
...
-import { Button } from '@mui/material';
+import { Button, Typography } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 
 import { TIMEOUTS } from '@northern.tech/store/constants';
 import { useDebounce } from '@northern.tech/utils/debouncehook';
 
 const useStyles = makeStyles()(theme => ({
+  container: {
+    backgroundColor: theme.palette.grey[50],
```
mineralsfree: we need a different color for dark theme here 🔦

frontend/src/js/common-ui/Time.tsx
```diff
@@ -75,8 +100,8 @@ export const RelativeTime = ({ className, shouldCount = 'both', updateTime }) =>
 };
 
 const cutoffDays = 14;
-export const ApproximateRelativeDate = ({ className, updateTime }) => {
-  const [updatedTime, setUpdatedTime] = useState();
+export const ApproximateRelativeDate = ({ className, updateTime }: { className?: string; updateTime?: string | Date }) => {
+  const [updatedTime, setUpdatedTime] = useState<Dayjs>();
 
   useEffect(() => {
     setUpdatedTime(updatedTime => (updateTime !== updatedTime ? dayjs(updateTime) : updatedTime));
```
mineralsfree, line 107: Not related to the changes, but why this useEffect even here? 🤔

frontend/src/js/common-ui/forms/Filters.tsx
```diff
@@ -76,17 +78,24 @@ export const Filters = ({ className = '', defaultValues, filters = [], initialVa
...
         {isDirty && (
-          <Button className={`absolute ${classes.filterReset}`} color="info" variant="outlined" onClick={() => reset()}>
-            Clear filter
-          </Button>
+          <div>
+            <Typography className="margin-top-small margin-bottom-small" color="text.secondary" variant="subtitle2">{`\u{200B}`}</Typography>
+            <Button className={`nowrap ${classes.filterReset}`} color="info" onClick={() => reset()} variant="outlined">
+              Clear filter
+            </Button>
+          </div>
```
mineralsfree, line 99: if Typography is only for layout, this could be a cleaner solution: 
```
<div style="
    display: flex;
    flex-direction: column;
    justify-content: end;
"><button>Clear filter</button>
</div>
```
specified in classes 😄

frontend/src/js/common-ui/forms/Filters.tsx
```diff
@@ -14,29 +14,31 @@
...
       gridTemplateRows: 'minmax(50px, max-content) 1fr'
-    },
-    '.filter-item > div': {
-      alignSelf: 'baseline'
     }
   },
-  filterReset: { right: theme.spacing(3) }
+  filterReset: {
+    marginBottom: 2,
+    marginTop: 2
```
mineralsfree, line 41: Do those `2px` margins really make a visible difference?

frontend/src/js/common-ui/Confirm.tsx
```diff
@@ -14,7 +14,19 @@
...
+import { Button, IconButton, Typography } from '@mui/material';
+import { makeStyles } from 'tss-react/mui';
+
+const useStyles = makeStyles()(theme => ({
+  nudgeInward: { marginRight: 6 },
+  wrapper: {
+    zIndex: 1,
+    background: theme.palette.info.light,
+    opacity: 1,
+    height: '100%',
```
mineralsfree, line 28: setting height to `100%` here needs to be addressed in the decommissioning device drawer. (Potentially other places?) 

Image

frontend/src/js/common-ui/Confirm.tsx
```diff
@@ -14,7 +14,19 @@
...
 import { Cancel as CancelIcon, CheckCircle as CheckCircleIcon, Check as CheckIcon, Close as CloseIcon, Edit as EditIcon } from '@mui/icons-material';
-import { Button, IconButton } from '@mui/material';
+import { Button, IconButton, Typography } from '@mui/material';
+import { makeStyles } from 'tss-react/mui';
+
+const useStyles = makeStyles()(theme => ({
+  nudgeInward: { marginRight: 6 },
+  wrapper: {
+    zIndex: 1,
+    background: theme.palette.info.light,
```
mineralsfree: Dark theme doesn't look right (background should be `theme.palette.info.dark` in dark theme)

frontend/src/js/components/deployments/progress/RolloutProgressBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+        </Tooltip>
+      </Typography>
+    ) : (
+      <Typography variant="caption">{`Current phase: ${currentPhaseIndex + 1} of ${phasesCount}`}</Typography>
+    )}
+  </div>
+);
+
+const Side = ({ totalFailureCount }: { totalFailureCount: number }) => (
+  <Typography variant="body2" className={`flexbox center-aligned ${totalFailureCount ? 'warning' : ''}`} style={{ justifyContent: 'flex-end' }}>
```
mineralsfree: I think we should add `flex-end` to `main.less`

frontend/tests/e2e_tests/integration/03-advanced/04-deployments.spec.ts
```diff
@@ -84,10 +84,11 @@ test.describe('Deployments', () => {
     await deviceGroupSelect.focus();
     await deviceGroupSelect.fill('All');
     await page.click(`#deployment-device-group-selection-listbox li:has-text('All devices')`);
-    await triggerDeploymentCreation(page, page.waitForSelector(selectors.deploymentListItem, { timeout: timeouts.tenSeconds }));
+    await triggerDeploymentCreation(page, page.getByRole('listitem').first().waitFor({ timeout: timeouts.tenSeconds }));
     await page.getByRole('tab', { name: /finished/i }).click();
-    await page.waitForSelector(selectors.deploymentListItemContent, { timeout: timeouts.sixtySeconds });
-    const datetime = await page.getAttribute(`${selectors.deploymentListItemContent} time`, 'datetime');
+    const listItem = page.getByRole('listitem').first();
```
mineralsfree: isn't it too fragile when used on page?

mineralsfree (commented):
Found a few things that could be followed up on later. The confirmation component is the only thing causing a real issue.

frontend/src/js/common-ui/Time.tsx
mzedel: That's for counting time differences between a point in time (e.g. deployment starting time, device check in time) and right now (e.g. _a few seconds ago_ or  _in 53s_) - I think the most prominent use case was the device check in time back when this was much more frequent... now it's used e.g. for deployment creation time or user & role changes.
Since the `updateTime` will be stable the component won't re-render on it's own in many situations, thus the effect.

frontend/src/js/common-ui/forms/Filters.tsx
mzedel: my idea was to mimic the regular filter items, which get their `top` defined by the `Typography` and `gridTemplateRows: 'minmax(50px, max-content) 1fr'` - I guess I could put this in a shared class for the filter items...
Putting the button at the "flex-end" will not always align if the filters have a different height (e.g. because of a notification shown underneath... so I plucked the "zero width space" approach from the MUI implementation for the labels on outlined inputs that do a somewhat similar thing 😅

frontend/src/js/components/deployments/progress/RolloutProgressBar.tsx
mzedel: https://github.com/NorthernTechHQ/nt-gui/pull/542 maybe?

frontend/tests/e2e_tests/integration/03-advanced/04-deployments.spec.ts
mzedel: IMO in the deployments case it should be the only one from the prominent page content - I adjusted the selector to reflect this a bit better... with the devices page it would need a bit more granular selection

frontend/src/js/common-ui/forms/Filters.tsx
mzedel: that's to vertically center the button with the inputs to the side - unlike the inputs the button shouldn't wrap (or grow to the same height), so the margins get the button to approximately align in total height with the regular input height.. the top margin does the main alignment and the bottom one is mostly to satisfy the "total height" approximation... 🤷

frontend/src/js/components/deployments/Report.tsx
```diff
@@ -79,6 +80,26 @@ export const DeploymentAbortButton = ({ abort, deployment }) => {
...
+  const finished = deployment.finished || deployment.status === DEPLOYMENT_STATES.finished;
+  const { failures, successes } = groupDeploymentStats(deployment);
+  const deploymentState = getDeploymentState(deployment);
+  let content = deploymentState;
+  let color = 'info';
+  if (finished && (!!successes || !failures)) {
+    content = successes === totalDeviceCount && totalDeviceCount > 1 ? `All ${successes}` : `${successes}`;
+    content = `${content} ${pluralize('devices', successes)} updated successfully`;
+    content = failures ? `${content} ${pluralize('devices', failures)} failed to update` : content;
+    color = failures ? 'error' : 'success';
```
mineralsfree: maybe 
```
  const allLabel = successes === totalDeviceCount && totalDeviceCount > 1 ? 'All ' : '';

  const successPart = `${allLabel}${successes} ${pluralize('device', successes)} updated successfully`;
  const failurePart = failures ? `. ${failures} ${pluralize('device', failures)} failed` : '';

  const content = finished ? `${successPart}${failurePart}` : deploymentState;
  const color = finished ? (failures ? 'error' : 'success') : 'info';
```

frontend/src/js/common-ui/forms/Filters.tsx
mineralsfree: Now I get your idea, but maybe it's ok just to have it in the bottom right corner using flexbox? Given some users with smaller screens will see it like this:
image

mineralsfree (dismissed):
Great job finishing the marathon task! 🏃‍♂️ 
Some last-mile adjustments are needed for the confirmation component. My other comment is just a minor suggestion.

frontend/src/js/components/deployments/Report.tsx
mzedel: looking at https://www.figma.com/design/vewvWAMm9vPomw4YZOHiGt/Follow-up-after-design-theme-update?node-id=6195-8400&t=LjK6Y3STQZyWRJcE-4 I would propose:
```js
  let content = deploymentState;
  let color = 'info';
  if (finished) {
    if (!!successes || !failures) {
      const allLabel = successes === totalDeviceCount && totalDeviceCount > 1 ? 'All ' : '';
      content = `${allLabel}${successes} ${pluralize('devices', successes)} updated successfully`;
    }
    content = failures ? `${failures} ${pluralize('devices', failures)} failed to update` : content;
    color = failures ? 'error' : 'success';
  }
```
as a slight mix 🍹

frontend/src/js/common-ui/forms/Filters.tsx
mzedel: It's not shown on the deployments page, but on the auditlog it's shown quite similar to your screenshot: https://www.figma.com/design/vewvWAMm9vPomw4YZOHiGt/Follow-up-after-design-theme-update?node-id=6685-27730&t=LjK6Y3STQZyWRJcE-4

mineralsfree: Wow, so it's intentional... 😅 I take my words back

mineralsfree (approved):

PR #1375, opened 2026-02-10
fix(store): let subscription page handle products retrieval to prevent erroneous on-prem requests

mzedel:
needs https://github.com/NorthernTechHQ/nt-gui/pull/532


mineralsfree (approved):

mzedel:
picked the commit into https://github.com/mendersoftware/mender-server/pull/1366 - to allow that to pass + get the updated store package in along the way 👌

PR #1382, opened 2026-02-12
QA-1457: Incorporate mail server in end-to-end test environment

alfrunes:
* Backported `compose/docker-compose.smtp4dev.yml` from enterprise
* Created an EmailClient interface with Gmail and smtp4dev implementations for staging and local environments respectively.
* Updated test for changing password to check email notification is received if emailClient is configured.


frontend/tests/e2e_tests/package.json
```diff
@@ -39,5 +39,10 @@
     "uuid": "^13.0.0",
     "validator": "^13.15.26",
     "yaml": "^2.8.2"
+  },
+  "dependencies": {
+    "@google-cloud/local-auth": "^2.1.0",
+    "google-auth-library": "^10.5.0",
+    "googleapis": "^105.0.0"
   }
```
mzedel: since none of the dependencies go "into the product" we only use (& also only monitor "devDependencies") here - can you please move these up and do `npm i` (to update the lockfile)?

alfrunes:
I need to do a little bit of refactoring of the GMail API part to use OAuth2 credentials.

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+
+            // Now that we have the code, use that to acquire tokens.
+            const r = await oAuth2Client.getToken(code);
+            // Make sure to set the credentials on the OAuth2 client.
+            oAuth2Client.setCredentials(r.tokens);
+            console.info(`Tokens acquired: ${JSON.stringify(r.tokens, null, 2)}`);
+            console.info(`To use gmail with the end-to-end tests, set the following environment variable:`);
+            console.info('```sh');
+            console.info(`export GOOGLE_REFRESH_TOKEN=${r.tokens.refresh_token}`);
+            console.info('```');
```
mzedel: Is there a specific reason why this would be markdown formatted?

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    // Open an http server to accept the oauth callback. In this simple example, the
+    // only request to our webserver is to /oauth2callback?code=<code>
+    const server = http
+      .createServer(async (req, res) => {
+        try {
+          if (req.url.indexOf('/oauth2callback') > -1) {
+            // acquire the code from the querystring, and close the web server.
+            const qs = new url.URL(req.url, 'http://localhost:3000').searchParams;
```
mzedel:
```suggestion
            const params = new url.URL(req.url, 'http://localhost:3000').searchParams;
```
or `queryString` if that's what's meant here...

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+            const code = qs.get('code');
+            res.end('Authentication successful! Please return to the console.');
+            server.destroy();
+
+            // Now that we have the code, use that to acquire tokens.
+            const r = await oAuth2Client.getToken(code);
```
mzedel: `r`? considering we have `res` and `req` taken already - maybe: `tokenResponse` or...?

frontend/tests/e2e_tests/fixtures/fixtures.ts
```diff
@@ -88,7 +90,20 @@ const test = (process.env.TEST_ENVIRONMENT === 'staging' ? nonCoveredTest : cove
...
+    if (environment == 'staging' && process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET && process.env.GOOGLE_REFRESH_TOKEN) {
+      // strip off opaque string in username: `username+opaque@domain.com`
+      const userId = username.replace(/(^[^+]+)(\+[^@]+)?(@.*)$/, '$1$3');
+      const gmail = new GmailEmailClient(process.env.GOOGLE_CLIENT_ID, process.env.GOOGLE_CLIENT_SECRET, process.env.GOOGLE_REFRESH_TOKEN, userId);
+      await use(gmail);
+    } else if (environment == 'enterprise') {
+      const emailUrl = process.env.SMTP4DEV_URL ? process.env.SMTP4DEV_URL : 'http://localhost:8025';
+      await use(new Smtp4devEmailClient({ baseUrl: emailUrl }));
+    } else {
+      await use(null);
```
mzedel: this is likely the path that is run the most and additionally the email functionality is not used for any other test and as such this is not really suitable as a fixture...

frontend/tests/e2e_tests/integration/09-potentially-destructive/01-settings.spec.ts
```diff
@@ -152,6 +152,10 @@ test.describe('Settings', () => {
       await page.screenshot({ path: './test-results/logout.png' });
       await page.getByRole('button', { name: /next/i }).waitFor({ timeout: timeouts.fiveSeconds });
       await expect(page.getByRole('button', { name: /next/i })).toBeVisible();
+      if (emailClient) {
+        const emails = await emailClient.getEmails({ to: username, unread: true });
+        expect(emails.find(email => email.subject === 'Your password has changed')).toBeDefined();
+      }
     });
```
mzedel, line 166:
```suggestion
});
test('password change triggers email notification', async ({ environment, emailClient, page, username }) => {
      test.skip(!isEnterpriseOrStaging(environment));
      const emails = await emailClient.getEmails({ to: username, unread: true });
      expect(emails.find(email => email.subject === 'Your password has changed')).toBeDefined();
});
```

or to avoid the `emailClient` fixture:

```
const emailClient = setupEmailClient(environment);
```

frontend/tests/e2e_tests/utils/email.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (textPart?.body?.data) {
+      return Buffer.from(textPart.body.data, 'base64url').toString('utf-8');
+    }
+
+    return '';
+  }
+
+  private async fetchMessage(id: string): Promise<Email> {
+    const res = await this.gmail.users.messages.get({
+      userId: 'me',
```
mzedel: should this maybe also be:
```suggestion
      userId: this.userId,
```
?

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+          }
+        } catch (e) {
+          reject(e);
+        }
+      })
+      .listen(3000, () => {
+        // open the browser to the authorize url to start the workflow
+        open(authorizeUrl, { wait: false, app: 'firefox' }).then(cp => cp.unref());
+      });
+    destroyer(server);
```
mzedel: The `server-destroyer` for just this is a bit much... nowadays there is:
```suggestion
    server.close();
    server.closeAllConnections();
```
which you could also wrap in a `setTimeout` if you want to give this some time to complete the workflow

frontend/tests/e2e_tests/utils/email.ts
```diff
@@ -0,0 +1,206 @@
...
+  from?: string;
+  to?: string;
+  unread?: boolean; // filter only unread messages
+}
+
+export interface EmailClient {
+  getEmails(params?: SearchParams): Promise<Email[]>;
+}
+
+export class GmailEmailClient implements EmailClient {
```
mzedel, line 36: while not really a problem, it does stick out as very odd considering the rest of the frontend codebase that is much more functional oriented - I assume your prompt didn't specify to align with the surrounding codebase 🫣

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+            console.info('```');
+            resolve(oAuth2Client);
+          }
+        } catch (e) {
+          reject(e);
+        }
+      })
+      .listen(3000, () => {
+        // open the browser to the authorize url to start the workflow
+        open(authorizeUrl, { wait: false, app: 'firefox' }).then(cp => cp.unref());
```
mzedel:
```suggestion
        open(authorizeUrl).then(cp => cp.unref());
```
to not tie this to any browser

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
...
+        'https://www.googleapis.com/auth/gmail.readonly' // To read messages
+      ]
+    });
+
+    // Open an http server to accept the oauth callback. In this simple example, the
+    // only request to our webserver is to /oauth2callback?code=<code>
+    const server = http
+      .createServer(async (req, res) => {
+        try {
+          if (req.url.indexOf('/oauth2callback') > -1) {
```
mzedel: since there is no handling of the alternative path here we can improve readability by avoiding indentation:
```suggestion
          if (req.url.indexOf('/oauth2callback') > -1) {
            return;
          }
```

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
```diff
@@ -0,0 +1,74 @@
+import { OAuth2Client } from 'google-auth-library';
+import http from 'http';
+import open from 'open';
+import destroyer from 'server-destroy';
```
mzedel: these might just happen to exist in the `node_modules` already, but aren't declared in the `package.json`

mzedel (changes requested):
I'm not sure what the status of this is, since I couldn't find any defined `GOOGLE_CLIENT_ID` or secret in the gitlab config to allow this to pass staging tests...?

In general it looks usable, albeit the email client is quite alone in the codebase - they might only know the `ErrorBoundary` as a fellow class.

The dependency categorization has to be addressed for the pipeline and ideally the fixture setup, take whatever you find reasonable from the rest.

alfrunes:
Thanks for the feedback!
I marked it as ready for review as this is not an area I'm very confident when it comes to structuring my implementation. I will take a look at your suggestions and try to incorporate as many as possible.

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
alfrunes: It was primarily to highlight that this is the only snippet that needs to be pasted into the shell session.

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
alfrunes: Good catch. I forgot to clean up this workaround when testing :face_with_peeking_eye:

frontend/tests/e2e_tests/utils/email.ts
alfrunes: It's more my Gopher :brain: preferring interfaces, and thus didn't really consider a functional implementation. If it's not too bad, I'd prefer leaving it as is for now.

frontend/tests/e2e_tests/integration/09-potentially-destructive/01-settings.spec.ts
alfrunes: My thinking was primarily to use fixtures to make the setup worker (/session) scoped, and then tests can dynamically detect whether the environment supports reading emails. But I can see this is breaking with the general paradigm. Will fix...

frontend/tests/e2e_tests/package.json
alfrunes: Like this b8947167996e459844e2821ed52c5aaf2f8912fa?

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
alfrunes: Removed the markdown snippet markers.

frontend/tests/e2e_tests/scripts/oauth2authorize.ts
alfrunes: (Also inverted the condition)

frontend/tests/e2e_tests/integration/09-potentially-destructive/01-settings.spec.ts
alfrunes: I moved the last assertion to it's own test here: 64e327bd6e2b95ddf1189243d5f7e9522f1ea2c3
And refactored the fixture here: ffc1d535725bf57bb2d0fb997a8a2a83b8949c0a

alfrunes:
There were quite a few changes since this PR was opened, but I put the last changes in separate commits that I will squash later.

frontend/tests/e2e_tests/package.json
mzedel: 🎯 exactly! now there's just https://github.com/mendersoftware/mender-server/pull/1382#discussion_r2892705403 IOW the `open` dependency missing in that list

frontend/tests/e2e_tests/integration/09-potentially-destructive/01-settings.spec.ts
mzedel: Thank you!

frontend/tests/e2e_tests/utils/email.ts
mzedel: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#interfaces these might be even more flexible 🎉, but don't worry and leave it as is 👌

mzedel (commented):
👌 thanks for the adjustments - I protected the branch in Gitlab so that the enterprise tests run, ideally you could give the staging tests a try with the service account a try to... to give them a try I usually do something like the following:
```diff
diff --git a/.gitlab-ci.yml b/.gitlab-ci.yml
index 803144dc..7d3efe07 100644
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -480,9 +480,6 @@ test:prep:
   stage: test
   extends: .build:base
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker
-  rules:
-    - if: $CI_COMMIT_BRANCH == "main"
-      when: on_success
   services:
     - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-dind
       alias: docker
@@ -497,7 +494,7 @@ test:prep:
 
 .template:test:staging-deployment:
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/mendersoftware/mender-test-containers:gui-e2e-testing
-  stage: .post
+  stage: test
   services:
     - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:${DOCKER_VERSION}-dind
       alias: docker
@@ -507,7 +504,6 @@ test:prep:
   needs:
     - job: test:prep
       artifacts: true
-    - job: mender-helm-version-bump:staging
   variables:
     CI: 1
     DEVICE_TYPE: qemux86-64
@@ -537,20 +533,6 @@ test:staging-deployment:chrome:
   script:
     - npx playwright install chromium
     - npm run script -- --environment=staging --project=chromium
-  rules:
-    - if: $CI_COMMIT_BRANCH == "main"
-      when: delayed
-      start_in: 30 minutes
-
-test:staging-deployment:firefox:
-  extends: .template:test:staging-deployment
-  script:
-    - npx playwright install firefox
-    - npm run script -- --environment=staging --project=firefox
-  rules:
-    - if: $CI_COMMIT_BRANCH == "main"
-      when: delayed
-      start_in: 40 minutes

--- a/frontend/tests/e2e_tests/integration/session.setup.ts
+++ b/frontend/tests/e2e_tests/integration/session.setup.ts
@@ -60,10 +60,6 @@ test.describe('Test setup', () => {
   });
   test('allows account creation', async ({ baseUrl, context, environment, page, pas
sword, request, username }) => {
     test.skip(environment !== 'staging');
-    test.setTimeout(6 * timeouts.sixtySeconds);
-    await pollDeployment({ location: baseUrl, attempt: 1, maxAttempts: 6, delay: 3 
* timeouts.tenSeconds }); // give max 6 * 30s to see a version update, same as curre
nt CI job delay
-    // wait a little extra to allow rollout to complete if we happen to poll the fi
rst replica in a deployment
-    await new Promise(resolve => setTimeout(resolve, 3 * timeouts.tenSeconds));
```

frontend/tests/e2e_tests/utils/email.ts
alfrunes: Thanks for the pointers! I'll leave it as is for now, but I'll prefer type aliases for future reference.

alfrunes:
@mzedel At this point, the tests don't really work for the staging environment as it depends on the `example.com` email domain to  skip the email verification check. Thus changing the email address tends to break the tests in several places. I think changing this should be a task of it's own as I did not consider this when making the task.
What do you think?

mzedel:
> @mzedel At this point, the tests don't really work for the staging environment as it depends on the `example.com` email domain to skip the email verification check. Thus changing the email address tends to break the tests in several places. I think changing this should be a task of it's own as I did not consider this when making the task. What do you think?

My assumption was that one of the reasons for this change would be to set the different setups to use the email verification? For staging the only change to use the staging user would probably be to either provide a username via: `--username <uniqueStagingUser>@northern.tech` or change https://github.com/mendersoftware/mender-server/blob/main/frontend/tests/e2e_tests/runner/config.js#L40 to:

```js
credentials.username = config.username ?? `username+${uuid()}@northern.tech`;
```

the rest of the `@example.com` users shouldn't run into any limitations (they are either only consider for the on-prem tests or only used to check if SSO users can log in)

alfrunes:
It seems like Gitlab is not able to post status here, but here's the latest pipeline run: https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/2433964354
The commit linter fails on purpose as a reminder to remove the debug commit.

compose/docker-compose.smtp4dev.yml
```diff
@@ -0,0 +1,21 @@
+# Include this overlay to test SMTP features like email verifications.
+# smtp4dev acts like a web interface for all email addresses.
+# Any email sent by the backend (to any recipient) will end up in the smtp4dev inbox.
+# A web interface for the indbox can be accessed on port 8025 on your machine.
```
mzedel:
```suggestion
# A web interface for the inbox can be accessed on port 8025 on your machine.
```

frontend/tests/e2e_tests/integration/02-baseline/02-settings.spec.ts
```diff
@@ -81,6 +84,34 @@ test.describe('Settings', () => {
       }
       await userCreationButton.waitFor();
     });
+
+    test('verify email address', async ({ baseUrl, environment, page, username }) => {
+      test.skip(environment !== 'staging' && environment !== 'enterprise');
```
mzedel: here and in the other places

```suggestion
      test.skip(!isEnterpriseOrStaging(environment));
```

frontend/tests/e2e_tests/integration/02-baseline/02-settings.spec.ts
```diff
@@ -81,6 +84,34 @@ test.describe('Settings', () => {
...
+      const emailClient = setupEmailClient(username, environment);
+      test.skip(!emailClient, 'test requires configuring a mailbox');
+      await page.goto(`${baseUrl}ui/settings/my-profile`);
+
+      test.skip(!(await page.getByText('Not verified').isVisible()), 'email is already verified');
+
+      await page.getByRole('button', { name: /Verify/i }).click();
+
+      // Wait five seconds for the email to be sent
+      await new Promise(resolve => setTimeout(resolve, timeouts.oneSecond * 3));
```
mzedel:
```suggestion
      await page.waitForTimeout(timeouts.default);
```

frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
```diff
@@ -115,10 +115,12 @@ test.describe('RBAC functionality', () => {
         await page.getByText(/Create user/i).click();
         await page.getByText('The user was created successfully.').waitFor();
       }
-      for (let j = 0; j < userCreations.length; j++) {
+      for (const { user } of userCreations) {
```
mzedel, line 118: 👍 thank you!

frontend/tests/e2e_tests/integration/session.setup.ts
```diff
@@ -55,7 +28,13 @@ test.describe('Test setup', () => {
   test('allows account creation', async ({ baseUrl, context, environment, page, password, request, username }) => {
     test.skip(environment !== 'staging');
     test.setTimeout(6 * timeouts.sixtySeconds);
-    await pollDeployment({ location: baseUrl, attempt: 1, maxAttempts: 6, delay: 3 * timeouts.tenSeconds }); // give max 6 * 30s to see a version update, same as current CI job delay
+    await poll({
+      callback: async (): Promise<boolean> => Promise.resolve(true), // FIXME: Bypass version check
```
mzedel: I'm tending to drop this entirely, since the check has failed to resolve due to slow rollouts and succeeded for rollouts of older gui images (but newer backend services) - so it's not ideal anyway... WDYT?

mzedel (commented):
consider this approved, just minor nitpicks I can follow up separately - the staging related changes in the ci file need to be reverted, but once that's done I'll ✅ asap

mzedel (approved):

PR #1383, opened 2026-02-12
test: give saml editor some time to load to not paste text into an unprepared void

mzedel:


mineralsfree (approved):
Should work, but I wonder if checking if `window.monaco` is defined is more reliable... 🤔

mzedel:
> Should work, but I wonder if checking if `window.monaco` is defined is more reliable... 🤔

Not necessarily, since the editor will be initialized and then extra modules will be loaded via https://github.com/suren-atoyan/monaco-loader?tab=readme-ov-file#init (there the `window.monaco` instance should be available while the loader will only begin loading)...

PR #1384, opened 2026-02-12
MEN-8645: Speed Dial adjustments

mineralsfree:


mzedel (approved):
thanks for the slots change + the e2e masking!! - oh and the change of course!

PR #1385, opened 2026-02-12
QA-1448(deviceconfig): Migrate mongo-driver to v2

bahaa-ghazal:


bahaa-ghazal:
if you want a sneak peek on other services that i migrated https://github.com/mendersoftware/mender-server/pull/1387

backend/pkg/mongo/v2/dbtest/db.go
```diff
@@ -0,0 +1,248 @@
...
+	Ctx      context.Context
+	timeout  time.Duration
+	client   *mongo.Client
+	output   bytes.Buffer
+	server   *exec.Cmd
+	dbpath   string
+	host     string
+	tomb     tomb.Tomb
+	registry *bson.Registry
+}
```
alfrunes, line 82: Do you think we can remove this one i v2 in favor of the `testing.dbClientFromEnv`?
In my experience, this implementation occasionally leave zombie processes after runs and requires you to install `mongod` on your system.

backend/pkg/mongo/v2/migrate/db.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	mopts "go.mongodb.org/mongo-driver/v2/mongo/options"
+
+	storev1 "github.com/mendersoftware/mender-server/pkg/store"
```
alfrunes, line 25:
```suggestion
	"github.com/mendersoftware/mender-server/pkg/store"
```

backend/pkg/mongo/v2/migrate/mocks/Migration.go
alfrunes, line 1: I don't think we need this mock interface. Usually when testing migrations we want to test them against the database.

backend/pkg/mongo/v2/migrate/migration.go
```diff
@@ -0,0 +1,24 @@
...
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
+package migrate
+
+// Migration defines an incremental mongo migration step, with a concrete
+// version. Current version of DB is passed as `from` parameter. Once migration
+// completes, DB will be in version Version().
+
+//go:generate ../../../../utils/mockgen.sh
```
alfrunes, line 20: I don't think this is really needed.
```suggestion
```

backend/pkg/store/utils.go
alfrunes, line 1: Store v1 and v2 make different assumptions on the multi-tenancy model. The v1 model is currently only used in deployments and inventory service.

backend/pkg/store/v2/utils.go
```diff
@@ -86,7 +86,13 @@ func ArrayWithTenantID(ctx context.Context, doc bson.A) bson.A {
...
 // from context and original database name
 func DbFromContext(ctx context.Context, origDbName string) string {
-	return origDbName
+	identity := identity.FromContext(ctx)
+	tenant := ""
+	if identity != nil {
+		tenant = identity.Tenant
+	}
+
+	return DbNameForTenant(tenant, origDbName)
```
alfrunes: This is turning `store/v2` into `store` (v1). This will break the multi tenancy model for most services.

alfrunes (changes requested):
You're off to a good start, well done! 👏 

Unfortunately, I believe there are some unintended breaking changes with the changes to the `pkg/store*` packages. When we migrated the multi-tenancy model to a single database. We took a similar approach like you did with the `pkg/mongo/v2/*` packages and since we still have 2 services that is using the old model the v1 package has stuck around.

I'm not sure what's the best approach here. Seeing that only `pkg/store/v2` depends on mongo, maybe you could try to refactor out the dependency to `pkg/mongo` and `pkg/mongo/v2`?

The rest of the comments are more nitpicks.

backend/services/deviceauth/store/mongo/datastore_mongo.go
```diff
@@ -816,7 +816,8 @@ func (db *DataStoreMongo) UpdateAuthSetById(
 	mod model.AuthSetUpdate,
 ) error {
 	c := db.client.Database(DbName).Collection(DbAuthSetColl)
-	res, err := c.UpdateOne(ctx, mongostore.WithTenantID(ctx, bson.M{"_id": id}), bson.M{"$set": mod})
+	res, err := c.UpdateOne(ctx, mongostore.WithTenantID(ctx, bson.M{"_id": id}),
+		bson.M{"$set": mod})
```
alfrunes, line 820: Nitpick: These formatting changes does not match the commit message 38620fb27d5b94910d04c62a519e250efee8b197.

backend/services/deviceconfig/store/mongo/migrations.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/mendersoftware/mender-server/pkg/log"
-	"github.com/mendersoftware/mender-server/pkg/mongo/migrate"
+	"github.com/mendersoftware/mender-server/pkg/mongo/v2/migrate"
 	mstore "github.com/mendersoftware/mender-server/pkg/store"
```
alfrunes, line 24: That's strange. It seems like the migration environment and the internal provision tenant API is still using the v1 multi-tenancy model, but is never used. We should clean this up. But that's beyond the scope of this task.

alfrunes (approved):
Looks good to me :rocket:

PR #1386, opened 2026-02-12
Test/golang integrations

mzedel:


merlin-northern (commented):
thank you! I have it on queue. I took a different path with the tests I am writing, but we will come to a cool conclusion (CC) for sure.

merlin-northern:
it is still on my queueI just need some spare time.

PR #1388, opened 2026-02-12
docs: required fix for the OpenAPI3 specs

merlin-northern:
Contains a required fix for the go client generated based on specs.
(We cant use NewDevice since it collides with the default constructor generated)

Changelog: None
Ticket: QA-1097

### DIFF
--- a/backend/docs/api/dist/openapi.yaml
+++ b/backend/docs/api/dist/openapi.yaml
@@ -5691,7 +5691,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/NewDevice'
+              $ref: '#/components/schemas/NewDeviceInternalProvision'
         required: true
       responses:
         '202':
@@ -9737,7 +9737,7 @@ components:
           type: number
         system:
           type: number
-    NewDevice:
+    NewDeviceInternalProvision:
       type: object
       properties:
         id:
--- a/backend/docs/api/iot-manager/internal_v1.yaml
+++ b/backend/docs/api/iot-manager/internal_v1.yaml
@@ -47,7 +47,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/NewDevice'
+              $ref: '#/components/schemas/NewDeviceInternalProvision'
         required: true
       responses:
         202:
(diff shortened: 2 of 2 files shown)


merlin-northern:
@alfrunes I need this to move forward with the rest of the things, I am making little steps easy to review, based on all the work we all did as a team.

What will follow is: sync of this to ENT with additional fixes in the ENT along these lines, then the generated clients dance OS and ENT (via sync and extra fixes, similar dance), then the Tiers test with the MEN-8571, and in the meanwhile the global use of the newly generated go client, together with the new PR from Manuel -- long path as you can see. I am making small delicious bites 😘  please review/approve.

_Edit_: it ended up being super small, this message is longer than the diff.

mzedel (changes requested):
@merlin-northern just a heads up: my PR there is quite a bit out of date, so it apparently lacks the statistics endpoint + the attribute values are also needed with the different types (instead of only strings)...

merlin-northern:
> @merlin-northern just a heads up: my PR there is quite a bit out of date, so it apparently lacks the statistics endpoint + the attribute values are also needed with the different types (instead of only strings)...

yeah I rebased it (or so I hope)

merlin-northern:
> > @merlin-northern just a heads up: my PR there is quite a bit out of date, so it apparently lacks the statistics endpoint + the attribute values are also needed with the different types (instead of only strings)...
> 
> yeah I rebased it (or so I hope)

ok I see it, I woke up wrong side of the diff.

merlin-northern:
> @merlin-northern just a heads up: my PR there is quite a bit out of date, so it apparently lacks the statistics endpoint + the attribute values are also needed with the different types (instead of only strings)...

apparently most of the (with regards to the docs only) https://github.com/mendersoftware/mender-server/pull/1178 is already in, so I checked file-by-file and this is it, one fix, one thing to carry over.

backend/docs/api/workflows/internal_v1.yaml
```diff
@@ -26,7 +26,7 @@ paths:
                 status: ok
         "500":
           $ref: '../common/responses.yaml#/components/responses/InternalServerError'
-  /api/v1/health:
+  /health:
```
alfrunes: Is this correct?
https://github.com/merlin-northern/mender-server/blob/b01afedeaed1ba8130aa3b205faff4434acd839e/backend/services/workflows/api/http/router.go#L30

merlin-northern: fixed. wow. ended up to be only one line.

mzedel (approved):

alfrunes (approved):

PR #1392, opened 2026-02-13
chore: generate client from the OpenAPI spec

merlin-northern:
Pipeline with triggered run:https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/2324579112

Ticket: QA-1097

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -253,6 +253,17 @@ test:backend:api-spec:
     - git diff --exit-code backend/tests/mender_client
   allow_failure: false
 
+test:backend:api-client:
+  extends: .template:combine-api
+  rules:
+    - changes:
+        paths: ["backend/docs/api/*.yaml", "backend/docs/api/**/*.yaml"]
+        compare_to: "${RULES_CHANGES_COMPARE_TO_REF}"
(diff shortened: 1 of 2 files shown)


merlin-northern:
Chapter 3: let's have a single client for a start, I need this to unblock the porting of all the places (tenantadm+fixes!, backendtests-ng) to one single generated client. If approved this will first go to ENT, then I will move forward. if need be I will follow the full excellent work by Manuel in https://github.com/mendersoftware/mender-server/pull/1178 with separate clients and models. For now let's take off!

mzedel (commented):
what is not ok about the one in https://github.com/mendersoftware/mender-server/tree/main/backend/tests/runner/client ? or would you want to replace that one with the one generated in here and use it in both product and tests?

merlin-northern:
> what is not ok about the one in https://github.com/mendersoftware/mender-server/tree/main/backend/tests/runner/client ? or would you want to replace that one with the one generated in here and use it in both product and tests?

exactly. I want to replace _all_ calls with the one in pkg. (nothing is wrong with it, it is the same as this one here)

mzedel (approved):

PR #1393, opened 2026-02-13
chore: enter the go client based on backend/docs/api/dist/openapi.yaml 

merlin-northern:
chore: enter the go client based on backend/docs/api/dist/openapi.yaml

this one adds a generated go client (based on https://github.com/mendersoftware/mender-server/pull/1392)
does also:
* do not generate go.sum and go.mod as it is not for us
* do not generate tests since for now we are not using them

Ticket: QA-1097


mzedel (approved):
I was looking frantically for the client already when I wanted to adjust the integration tests pr...

merlin-northern:
> I was looking frantically for the client already when I wanted to adjust the integration tests pr...
😂 

yes, same here, when I wanted to use it in the tests-ng. sorry, I am a bit slow on the uptake.

PR #1394, opened 2026-02-13
ci: made use of shared nt ci base images

oldgiova:


mzedel (approved):
Thank you for cleaning up after me 🫣

PR #1398, opened 2026-02-16
fix(gui): show alert if dynamic group is removed after deployment

mineralsfree:
Ticket: ME-550

Requires : 

- [x] https://github.com/NorthernTechHQ/nt-gui/pull/540

### DIFF
--- a/frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
+++ b/frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
@@ -48,10 +48,10 @@ const useStyles = makeStyles()(theme => ({
   releaseSelectText: { minWidth: 0, flexGrow: 1 }
 }));
 
-export const getDevicesLink = ({ devices, filters = [], group, name }) => {
+export const getDevicesLink = ({ devices, filters = [], group, name, groupId }) => {
   let devicesLink = '/devices';
   if (filters.length) {
-    return `${devicesLink}?${formatDeviceSearch({ pageState: {}, filters, selectedGroup: group })}`;
+    return `${devicesLink}?${formatDeviceSearch({ pageState: {}, filters, selectedGroup: group, groupId })}`;
   }
   // older deployments won't have the filter set so we have to try to guess their targets based on other information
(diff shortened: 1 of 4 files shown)


frontend/src/js/components/devices/DeviceGroups.tsx
```diff
@@ -147,6 +149,20 @@ export const DeviceGroups = () => {
...
+      dispatch(getDynamicGroup(locationParams.groupId))
+        .unwrap()
+        .catch(error => {
+          if (error.message && error.message.includes('404')) {
+            // Remove group selection in case group is removed / modified
+            dispatch(selectGroup({ group: locationParams.groupId }));
+            setGroupRemoved(locationParams.groupName);
+          }
+        });
+    }
```
mzedel:
```suggestion
    if (locationParams.groupId) {
      return;
    }
      dispatch(getDynamicGroup(locationParams.groupId))
        .unwrap()
        .catch(error => {
          if (error.message && error.message.includes('404')) {
            // Remove group selection in case group is removed / modified
            dispatch(selectGroup({ group: locationParams.groupId }));
            setGroupRemoved(locationParams.groupName);
          }
        });
```

just maybe

frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
```diff
@@ -142,7 +142,7 @@ export const Devices = ({
   const dispatch = useAppDispatch();
 
   const { deploymentDeviceCount = 0, devices = [], filter, group = null } = deploymentObject;
-  const device = devices.length === 1 ? devices[0] : {};
+  const device = useMemo(()=>devices.length === 1 ? devices[0] : {}, [devices]);
```
mzedel:
```suggestion
  const device = useMemo(() => devices.length === 1 ? devices[0] : {}, [devices]);
```

Thanks for adding this, and annoying that the prettier config seems to be broken enough to not catch this/ take care of it for us even before 😞...

mzedel (commented):
that looks usable, just some irrelevant nitpicks while waiting for the packages to be out & included...

mzedel (approved):

PR #1399, opened 2026-02-16
chore: format generated client with go fmt

merlin-northern:
chore: format generated client with go fmt

Ticket: QA-1097


mzedel (commented):
Could we instead ignore that client entirely from formatting, linting etc.? We should never need to touch it...

merlin-northern:
> Could we instead ignore that client entirely from formatting, linting etc.? We should never need to touch it...

I was torn a bit when I noticed that it is not formatted, let's hear others opinion. somehow the linter knows it is generated and does not complain. (let me check maybe there is a flag for it) (ok, I have another way) (I wanted to use -e GO_POST_PROCESS_FILE="gofmt -w" but there is no gofmt in the container we are running the generator from, so that one failed)

mzedel:
@merlin-northern this has been sitting stale for a while without a reaction, please reopen if you still intend to do something with this

PR #1401, opened 2026-02-17
Qa -1448(part1): Migrate mongo-driver to v2

bahaa-ghazal:
I decided to migrate services in batches instead of one by one, because I would then create ~20 PRs, one for every service in OS and ENT


kjaskiewiczz (approved):
looks good

backend/services/deployments/store/mongo/datastore_mongo.go
```diff
@@ -319,10 +288,8 @@ var (
 			{Key: StorageKeyDeploymentName, Value: 1},
 			{Key: StorageKeyDeploymentCreated, Value: 1},
 		},
-		Options: &mopts.IndexOptions{
-			Background: &_true,
-			Name:       &IndexNameDeploymentName,
-		},
+		Options: mopts.Index().
+			SetName(IndexNameDeploymentName),
```
merlin-northern, line 292: background is lost

bahaa-ghazal: It was removed after v4.2 : https://www.mongodb.com/community/forums/t/after-4-2-background-createindexes/15136

PR #1402, opened 2026-02-17
QA-1472: Relax e2e match criteria for `mender-demo-artifact` version

lluiscampos:
With upcoming changes from the demo artifact, we will not anymore have a predictable version but it is generated on each build and publish.

This commit just expects the artifact name to be `mender-demo-artifact-<something>` and verify that the same version is also provided as `data-partition.mender-demo-artifact.version`.

See:
* https://github.com/mendersoftware/mender-demo-artifact/pull/369

### DIFF
--- a/frontend/tests/e2e_tests/fixtures/fixtures.ts
+++ b/frontend/tests/e2e_tests/fixtures/fixtures.ts
@@ -18,17 +18,11 @@ import { test as nonCoveredTest } from '@playwright/test';
 import { getPeristentLoginInfo } from '../utils/commands.ts';
 import { timeouts } from '../utils/constants.ts';
 
-type DemoArtifactVersionInfo = {
-  artifactVersion: string;
-  updateVersion: string;
-};
-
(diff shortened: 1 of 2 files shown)


lluiscampos:
@mzedel :warning: Untested change. I was not able to make the test pass locally - it failed in a precondition of finding the Artifact :man_shrugging:

mzedel (dismissed):
👍 that looks reasonable to me - thank you!

looking at the test I see that the demo artifact will be downloaded from `https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender` - will this continue to work?

lluiscampos:
> looking at the test I see that the demo artifact will be downloaded from `https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender` - will this continue to work?

@mzedel Yes! It must continue to work because Mender Server uses it (also the e2e tests - which is great)

mzedel (approved):

PR #1405, opened 2026-02-17
chore(gui): user creation timeout e2e test

mineralsfree:
Ticket: QA-1353


mzedel (commented):
that should trigger it - if it is in any way still present. Maybe this too was due to the extra load on the staging cluster...

mzedel:
@mineralsfree if you drop the debug commit we can get this in right away 👌

mzedel (approved):

PR #1407, opened 2026-02-18
QA-1448(part2): Migrate mongo-driver to v2

bahaa-ghazal:


backend/pkg/mongo/v2/oid/objectid.go
```diff
@@ -126,6 +129,7 @@ func (oid ObjectID) MarshalBSONValue() (byte, []byte, error) {
 
 // UnmarshalBSONValue provides the bson.ValueUnmarshaler interace.
 func (oid *ObjectID) UnmarshalBSONValue(t byte, b []byte) error {
+
```
merlin-northern: is this one needed?

backend/pkg/mongo/v2/doc/bson.go
```diff
@@ -58,6 +58,12 @@ func DocumentFromStruct(
 		return nil
 	}
 
+	// If a field already exists, don't append a duplicate field but override it
+	overrides := make(map[string]bool)
+	for _, e := range appendElements {
+		overrides[e.Key] = true
+	}
+
```
merlin-northern, line 66: how come this is needed by the v2 driver?

backend/services/deviceauth/model/limit.go
```diff
@@ -22,10 +24,10 @@ var (
 )
 
 type Limit struct {
-	Id       string `json:"-" bson:"_id,omitempty"`
-	Name     string `json:"name" bson:"name"`
-	Value    uint64 `json:"value" bson:"value"`
-	TenantID string `json:"-" bson:"tenant_id"`
+	Id       oid.ObjectID `json:"-" bson:"_id,omitempty"`
```
merlin-northern, line 27: is this change required part of the port to the new driver version?

backend/services/inventory/model/device.go
```diff
@@ -249,6 +247,18 @@ func (gn GroupName) Validate() error {
...
+func (da DeviceAttributes) ToMap() map[string]DeviceAttribute {
+	// Create map with capacity equal to slice length
+	m := make(map[string]DeviceAttribute, len(da))
+	for _, attr := range da {
+		// Construct the unique key: <scope>-<name>
+		key := attr.Scope + "-" + attr.Name
+		m[key] = attr
+	}
+	return m
+}
```
merlin-northern: where is that function  used? is it required part of the port to the v2?

backend/pkg/mongo/v2/doc/bson.go
```diff
@@ -67,7 +73,16 @@ func DocumentFromStruct(
...
+			inlineDoc := DocumentFromStruct(valFace)
+			for _, elem := range inlineDoc {
+				if !overrides[elem.Key] {
+					doc = append(doc, elem)
+				}
+			}
+			continue
+		}
+		// Skip the struct field if it is in the overrides list
+		if overrides[key] {
```
merlin-northern, line 85: I do not get it, why is this needed by the port to v2 driver?

merlin-northern (commented):
looks good, my comments follow from the fact that I have not looked at the v2 driver at all, so I know nothing.

backend/pkg/mongo/v2/oid/objectid.go
bahaa-ghazal: Yes ... For my sanity 😔

backend/pkg/mongo/v2/doc/bson.go
bahaa-ghazal: Long store my friend...
So this is used when calling WithTenantID() with a struct as a parameter. The problem is when the struct already has tenant field like struct `model.Device`. I faced an insane bug after the migration in the test `TestStoreDeleteDevice` where you insert `model.Device`-s and you look at the db (used compass) it looks totally fine! But the moment you try to get a device by the tenant id nothing comes back! Trying to fetch all the docs in collection will pass, but only when you filter by tenant id it fails. At the end after losing my sanity it turned out that the stored bson doc had 2 `tenant_id` fields, one empty and one with the correct value, and when you filter it checks the input against he empty value... 
Sorry for my rant 😪
Anyways I faced multiple cases where the test shouldn't have passed at all but fails only with the new driver.
Now that I'm thinking about it, we need to check that the production DB is not corrupted with hidden tenant_id field, else this will be a problem

backend/services/deviceauth/model/limit.go
bahaa-ghazal: Yes, the new driver does not unmarshal ObjectID to string by default. There is a work around to make this happen, but if it stored in the db as  ObjectID I don't see why not to unmarshal it as ObjectID 🤷

backend/services/inventory/model/device.go
bahaa-ghazal: Here: https://github.com/bahaa-ghazal/mender-server/blob/2e4e49dbf40ebd2a60d3aba666af88f4f4ef9ba6/backend/services/inventory/store/mongo/datastore_mongo_test.go#L2965

What this do is turn the `DeviceAttributes` from a slice of `DeviceAttribute` to `map[string]DeviceAttribute`, the same way we stored. It is just quality of life func.

I need this because test that should've fails actually failed, and that we used to compare if 2 slices are similar. I swapped them into the map form to not make them position dependent.

backend/pkg/mongo/v2/doc/bson.go
bahaa-ghazal: I hope you get it now 🧞‍♂️

backend/services/inventory/model/device.go
merlin-northern: I could not find it because gh as hiding the changes. sorry for that. so if it is used only in the tests, why isnt it defined in the tests?

backend/services/inventory/model/device.go
```diff
@@ -285,43 +295,43 @@ func GetDeviceAttributeNameReplacer() *strings.Replacer {
...
-		err = elem.Value().Unmarshal(&(*d)[i])
-		if err != nil {
-			return err
-		}
+
+	*da = make(DeviceAttributes, 0, len(tempMap))
+
+	// Convert map -> slice
+	for _, attr := range tempMap {
+		// We just append the value. The scope and name are already inside DeviceAttribute
```
merlin-northern: I am a bit mislead about this comment. and just in passing, could you double check that you do not have to make a deep copy with regards to `Description *string` and `Timestamp   *time.Time` here?

backend/services/inventory/model/device.go
```diff
@@ -285,43 +295,43 @@ func GetDeviceAttributeNameReplacer() *strings.Replacer {
...
-			Scope:       d[i].Scope,
-			Timestamp:   d[i].Timestamp,
-		}
-		attrs[i].Key = attr.Scope + "-" + replacer.Replace(d[i].Name)
-		attrs[i].Value = &attr
+
+	// Convert slice -> map
+	for _, attr := range da {
+		key := attr.Scope + "-" + replacer.Replace(attr.Name)
+		output[key] = attr
```
merlin-northern: I think this should be
```suggestion
	for k, attr := range da {
		key := attr.Scope + "-" + replacer.Replace(attr.Name)
		a := DeviceAttribute{
			Name:        da[k].Name,
			Description: da[k].Description,
			Value:       da[k].Value,
			Scope:       da[k].Scope,
			Timestamp:   da[k].Timestamp,
		}
		output[key] = a
```
just to be on a safe side

backend/services/deviceauth/model/limit.go
merlin-northern: thanks for the explanation.

backend/pkg/mongo/v2/doc/bson.go
merlin-northern: I am trying :)

backend/pkg/mongo/v2/doc/bson.go
merlin-northern: wow. I super understand what you were going through. perhaps it is a good topic for the MSTM.

merlin-northern (commented):
looks good I just have last doubts about the bson Unmarshal and Marshall

backend/services/inventory/model/device.go
bahaa-ghazal: So looking at the db we don't store `Description` or `Timestamp`, well for `Timestamp` we store attributes `system-created_ts` and `system-updated_ts` where the value is the timestamp.

backend/services/inventory/model/device.go
bahaa-ghazal: I thought it could be useful and reusable.

backend/pkg/mongo/v2/doc/bson.go
bahaa-ghazal: I'll test it locally, if the old driver is corrupting the db, then we need to clean it up.

bahaa-ghazal: so I have looked all over the repo and the only place we misused the `WithTenantID` is in the test : https://github.com/mendersoftware/mender-server/blob/e7907c19bde119d533558b8a0758810c2dcd6b08/backend/services/deviceauth/store/mongo/datastore_mongo_test.go#L1669.
It should be fine to merge it given that there was no misusage from before.

backend/services/inventory/model/device.go
```diff
@@ -285,43 +295,49 @@ func GetDeviceAttributeNameReplacer() *strings.Replacer {
...
-		if err != nil {
-			return err
-		}
+
+	*da = make(DeviceAttributes, 0, len(tempMap))
+
+	// Convert map -> slice
+	for _, attr := range tempMap {
+		*da = append(*da, attr)
 	}
```
merlin-northern: I leave it up to you but I would drop the append and use the indexing since we know exactly the sizes here.

merlin-northern (dismissed):
you are my hero with all this migration. on a side: have you tried to upgrade some ancient installation and run all those migrations on real data?

backend/pkg/mongo/v2/doc/bson.go
```diff
@@ -58,6 +58,12 @@ func DocumentFromStruct(
 		return nil
 	}
 
+	// If a field already exists, don't append a duplicate field but override it
+	overrides := make(map[string]bool)
```
merlin-northern, line 62: nit, we know the max len so could 
```go
overrides := make(map[string]bool,len(appendElements))
```

backend/pkg/mongo/v2/doc/bson.go
```diff
@@ -67,7 +73,16 @@ func DocumentFromStruct(
 		value := s.Field(i)
 		key, valFace, set := valueFromStructField(field, value)
 		if key == "inline" {
-			doc = append(doc, DocumentFromStruct(valFace)...)
+			inlineDoc := DocumentFromStruct(valFace)
+			for _, elem := range inlineDoc {
+				if !overrides[elem.Key] {
```
merlin-northern, line 78: this key may not exist in the `overrides` map, so the nice way is to check if it exists, for instance:
```
if _,exists:=overrides[elem.Key];!exists {
```

backend/pkg/mongo/v2/doc/bson.go
```diff
@@ -67,7 +73,16 @@ func DocumentFromStruct(
...
+					doc = append(doc, elem)
+				}
+			}
+			continue
+		}
+		// Skip the struct field if it is in the overrides list
```
merlin-northern, line 84: same comment for the line below
```
if _,exists:=overrides[key];exists {
```

backend/services/deviceauth/store/mongo/migration_1_10_0_test.go
```diff
@@ -47,31 +47,24 @@ func TestMigration_1_10_0(t *testing.T) {
...
 				Keys: bson.D{
 					{Key: model.DevKeyStatus, Value: 1},
 					{Key: model.DevKeyId, Value: 1},
 				},
-				Options: &options.IndexOptions{
-					Background: &_false,
-					Name:       &indexDevices_Status,
-				},
+				Options: options.Index().
+					SetName(indexDevices_Status),
```
merlin-northern, line 67: okey just to make me calm, the Background does not exist anymore in v2?

merlin-northern (approved):
just left two comments that were nagging me. you are my hero with those migrations to v2.

backend/services/deviceauth/store/mongo/migration_1_10_0_test.go
bahaa-ghazal: this explain it https://www.mongodb.com/community/forums/t/after-4-2-background-createindexes/15136

bahaa-ghazal:
@merlin-northern The reason why there are an merge commit is that I'm merging the OS changes manually with my enterprise branch. It is a merge commit.
Thanks for the review 😘

PR #1454, opened 2026-03-03
test(gui): added retries to a file download part of the file transfer test

mineralsfree:


mzedel (dismissed):
looks good already, but can you expand this to cover the `Files -> downloads -> allows artifact downloads` test as well?

mineralsfree:
Not sure what the best way to accommodate this in an artifact test. Lmk if it's fine or can be improved @mzedel  
UPD: looks like in case of an artifact, it fails consistently... Will take a look tomorrow

mzedel:
> UPD: looks like in case of an artifact, it fails consistently... Will take a look tomorrow

I could imagine the fallback `request` based download might have been the norm there...

mzedel (approved):

PR #1485, opened 2026-03-06
ME-631 - fix(create-artifact-worker): hardened single file generation input validations

mzedel:
- this should limit path traversal possibilities and reduce possibilities for malicious user inputs

### DIFF
--- a/backend/services/create-artifact-worker/cmd/single-file.go
+++ b/backend/services/create-artifact-worker/cmd/single-file.go
@@ -32,15 +32,17 @@ import (
 )
 
 const (
-	argToken          = "token"
-	argArtifactName   = "artifact-name"
-	argDescription    = "description"
-	argDeviceType     = "device-type"
-	argArtifactId     = "artifact-id"
-	argGetArtifactUri = "get-artifact-uri"
-	argDelArtifactUri = "delete-artifact-uri"
-	argTenantId       = "tenant-id"
-	argArgs           = "args"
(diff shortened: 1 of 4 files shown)


backend/services/create-artifact-worker/cmd/single-file.go
```diff
@@ -231,14 +245,64 @@ func (c *SingleFileCmd) Validate() error {
...
-	if c.FileName == "" {
-		return errors.New("destination filename can't be empty")
+	// Validate required fields
+	requiredFields := []struct {
+		name   string
+		value  string
+		maxLen int
+	}{
+		{"filename", c.FileName, maxInputLength},
+		{"artifact-name", c.ArtifactName, 4 * maxInputLength},
```
kjaskiewiczz: why it's 4 times more than maxInputLength?

kjaskiewiczz (dismissed):
looks good; thanks!

backend/services/create-artifact-worker/cmd/single-file.go
mzedel: 🤦 right, I assume that should be 4096 to match the `lengthIn1To4096` validator used in the deployments service

kjaskiewiczz: I have no idea tbh :D

kjaskiewiczz (approved):

PR #1488, opened 2026-03-09
MEN-9321 - feat(gui): add DocsTextLink variant for inline documentation links

mzedel:


frontend/src/js/common-ui/DocsLink.tsx
```diff
@@ -120,6 +120,26 @@ export const DocsTooltip = ({ anchor = {}, id = '', ...props }) => {
...
 
+interface DocsTextLinkProps {
+  [key: string]: unknown;
+  capitalizedStart?: boolean;
+  children?: ReactNode;
+  id: keyof typeof DOCSTIPS;
+}
+
+export const DocsTextLink = ({ capitalizedStart = true, children, id, ...props }: DocsTextLinkProps) => {
+  const { path } = DOCSTIPS[id] || {};
```
mineralsfree: defaulting to {} can cause target in DocsLink to be `'https://docs.mender.io/undefined'`

frontend/src/js/common-ui/DocsLink.tsx
```diff
@@ -120,6 +120,26 @@ export const DocsTooltip = ({ anchor = {}, id = '', ...props }) => {
...
+  return (
+    <DocsLink path={path} {...props}>
+      {children ?? (
```
mineralsfree: I think `||` instead of `??` here will be safer

mineralsfree (changes requested):
I think we shouldn't return anything if DOCSTIPS id is wrong

mineralsfree (approved):

PR #1507, opened 2026-03-11
Fix/copy column values more

mzedel:
the prior change didn't take the inventory table into account - to prevent this both now reuse a component

### DIFF
--- a/frontend/src/js/common-ui/CopyableText.tsx
+++ b/frontend/src/js/common-ui/CopyableText.tsx
@@ -1,0 +1,77 @@
+// Copyright 2026 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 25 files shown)


frontend/src/js/common-ui/CopyableText.tsx
```diff
@@ -0,0 +1,77 @@
...
+  children: ReactNode;
+  onCopy?: () => void;
+  textClasses?: string;
+  title?: string;
+}
+
+export const CopyableText = ({ children, onCopy, textClasses = '', title }: CopyableTextProps) => {
+  const { classes } = useStyles();
+  const [isHovered, setIsHovered] = useState(false);
+  const isComponent = React.isValidElement(children);
```
mineralsfree: Maybe   `React.isValidElement(children) ||
  (Array.isArray(children) &&
    children.some(child => React.isValidElement(child)))`;

since array of React Elements is not a valid Element => Typography wil use `<p>{children}</p>`

mineralsfree (dismissed):
One small improvement suggestion to stay on a safer side

frontend/src/js/common-ui/CopyableText.tsx
mzedel: Good point - adjusted 👌

mineralsfree (approved):

PR #1513, opened 2026-03-12
QA-1538: Add Mender Client 6 to backend tests `test_compat`

lluiscampos:


lluiscampos:
I believe that the test is not running in CI. I don't have much more time left myself to follow-up, so I would appreciate someone from the Mender Server team to take over :innocent:

mzedel (approved):
with the tests now apparently running again I'd be happy to use this as is and follow up with potential adjustments to the test setup at a later time 👌

mzedel:
merging this as the typo fix is needed to have the tests pass in the ENT repo

PR #1516, opened 2026-03-12
feat(gui): added filtering by tiers on device list page

mineralsfree:
- [x] Neeeds https://github.com/NorthernTechHQ/nt-gui/pull/589

### DIFF
--- a/frontend/src/js/components/devices/AuthorizedDevices.tsx
+++ b/frontend/src/js/components/devices/AuthorizedDevices.tsx
@@ -23,11 +23,21 @@ import { makeStyles } from 'tss-react/mui';
 import { defaultTextRender, getDeviceIdentityText } from '@northern.tech/common-ui/DeviceIdentity';
 import Loader from '@northern.tech/common-ui/Loader';
 import storeActions from '@northern.tech/store/actions';
-import { ALL_DEVICES, DEVICE_ISSUE_OPTIONS, DEVICE_STATES, SORTING_OPTIONS, TIMEOUTS, UNGROUPED_GROUP, onboardingSteps } from '@northern.tech/store/constants';
+import {
+  ALL_DEVICES,
+  DEVICE_FILTERING_OPTIONS,
+  DEVICE_ISSUE_OPTIONS,
+  DEVICE_STATES,
+  SORTING_OPTIONS,
+  TIMEOUTS,
+  UNGROUPED_GROUP,
+  onboardingSteps
+} from '@northern.tech/store/constants';
 import {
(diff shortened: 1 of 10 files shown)


frontend/src/js/components/devices/AuthorizedDevices.tsx
```diff
@@ -360,6 +373,23 @@ export const Authorized = ({
 
   const onDeviceIssuesSelectionChange = ({ target: { value: selectedIssues } }) =>
     dispatchDeviceListState({ selectedIssues, page: 1, refreshTrigger: !refreshTrigger });
+  useEffect(() => {
+    const tierFilter = filters.find(f => f.key === 'tier' && f.scope === 'system');
```
mzedel:
```suggestion
    const tierFilter = filters.find(({ key, scope }) => key === 'tier' && scope === 'system');
```
just since the `f` doesn't add any real context - but really just a suggestion

mzedel: and additionally, since for this we do have constants:
```ts
import { Scope } from '@northern.tech/types/MenderTypes';
...
    const tierFilter = filters.find(f => f.key === 'tier' && f.scope === Scope.SYSTEM);
```
and I'll take a look at getting the same available from the different tiers...

frontend/src/js/components/devices/widgets/TierSelection.tsx
```diff
@@ -0,0 +1,28 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+import { MenuItem, Select } from '@mui/material';
+
+export const TierSelection = ({ className = '', onChange, selectedTier = '', enabledTiers }) => (
+  <div className="flexbox align-items-center margin-left">
+    Tier:
```
mzedel:
```suggestion
    <Typography variant="body2">Tier:</Typography>
```
but OTOH that would then have to be aligned with the others as well...

mzedel (dismissed):
LGTM - thank you!

If you feel like it, I'd be happy to also see the device stuff selectors aligned to use Typography + the slightly adjusted gap between title & select (to allow this task be fully aligned with the design without sticking out), but we can also have this in a follow up as it's not what the related epic wants to get done 👌

mzedel (approved):
✨ 😄

PR #1522, opened 2026-03-13
🍒 4.0.x - chore: Revert Use Virtual Device for OS updates in docker composition"

mzedel:
This reverts commit 9142d7aab449c043000d65b77cd66390d8ace461.

Once that we have branch out 4.0.x with the old Virtual Device, we can bring back the Virtual Device for Application Updates to `main` \o/

Ticket: MEN-7789

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -389,12 +389,25 @@ services:
         aliases: [mender-nats]
 
   client:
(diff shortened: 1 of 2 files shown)


merlin-northern (approved):
thanks for working on it

PR #1525, opened 2026-03-16
MEN-9223: validate the length of role and group name definitions

mineralsfree:


frontend/src/js/common-ui/forms/Form.tsx
```diff
@@ -47,6 +47,11 @@ const getErrorMsg = (validateMethod, args) => {
       } else {
         return 'This please only enter valid characters. Valid characters are a-z, A-Z, 0-9, _ and -';
       }
+    case 'maxLength':
```
mzedel: since the validations rely heavily on https://github.com/validatorjs/validator.js one of the ideas was to not replicate functionality present in there, so I thought relying on the `isLength` would be possible (like https://github.com/validatorjs/validator.js/blob/master/src/lib/isLength.js#L9-L15) - but as long as we don't keep adding more and more custom validation methods here I guess that's good enough too!

frontend/src/js/components/devices/group-management/GroupDefinition.tsx
```diff
@@ -27,33 +27,36 @@ const filter = createFilterOptions();
...
-  let errortext = null;
+  let errorText = '';
   const isModification = name.length && groups.some(group => decodeURIComponent(group) === name);
   if (!name && !isModification) {
     invalid = true;
   } else if (!validator.isWhitelisted(name, 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-.')) {
     invalid = true;
-    errortext = 'Valid characters are a-z, A-Z, 0-9, ., _ and -';
+    errorText = 'Valid characters are a-z, A-Z, 0-9, ., _ and -';
+  } else if (name.length > 256) {
```
mzedel: can you put the 256 in a constant somewhere in the file?

mzedel (dismissed):
just the constant definition would be nice to have, the other one is just to keep in mind for the future

frontend/src/js/common-ui/forms/Form.tsx
mineralsfree: I don't see a nice way of doing it in the current approach. I think we should move away from string parsing in general. Created a ticket https://northerntech.atlassian.net/browse/MEN-9487

mzedel: that's what I meant with the link to the implementation - you can use `isLength:1:256` straight away and then it would require adjusting the `getErrorMsg` in the `Form` component...

as to getting away from string parsing, that would be nice and I guess is getting more useful with the incoming variations in when what validation should be run... ideally I'd just like to get most of the forms to RHF first, to keep the variations at bay

mzedel (approved):

PR #1526, opened 2026-03-16
refactor: Move user model to common pkg

bahaa-ghazal:
This is a change preparing for ticket: https://northerntech.atlassian.net/browse/MEN-9458

This will enable us to decouple models in pkg instead of importing models strength from the services.
This will make refactoring/migrating thing a lot safer.

@alfrunes what do you think about this idea?


backend/services/useradm/model/user.go
```diff
@@ -148,7 +116,7 @@ func (u User) NextETag() (ret ETag) {
 
 func (u User) Validate() error {
 	if err := validation.ValidateStruct(&u,
-		validation.Field(&u.Email, validation.Required),
+		validation.Field(&u.Email, validation.By(rules.Email), validation.Required),
```
alfrunes, line 119: `rules.Email` is called implicitly by the Email.Validate type (called implicitly if the value is non-empty).
```suggestion
		validation.Field(&u.Email, validation.Required),
```

backend/services/useradm/store/mongo/migration_2_0_0_test.go
```diff
@@ -48,28 +49,36 @@ func TestMigration_2_0_0(t *testing.T) {
 				"foo": {
 					inUsers: []model.User{
 						{
-							ID:    "foo",
-							Email: "foo@bar.com",
+							User: user.User{
```
alfrunes, line 52: Nitpick: this reads quite repetitive, there are 3 levels of User here...

alfrunes (approved):
Generally positive to this change. Centralizing definitions is one of the goals we wanted to achieve after migrating to the mono repo.
I'm a little bit conflicted as to where OAS generated types fits into this, but this seems like a change in the right direction although in an ideal world the types should even be generated.
Anyway, it's a :+1: from me, but I'd like to hear from at least one more @mendersoftware/backend-dependabot-reviewers.

kjaskiewiczz (changes requested):
I know we have similar patterns here and there, but you have to be aware that what we do in other places is unfortunately not always what we should actually do.
We want to move common things to common package. Both useradm and tenantadm will be synchronizing data with HubSpot, so it make sense to have hubspot client in the common package, but I think having all models from all services in common package is a bit of oversharing and I'd rather create representation of hubspot user object in hubspot client.
User model IMO should stay in useradm service. If you are not convinced we can have a call with Frode or Alf (or both if you like).

bahaa-ghazal:
> I know we have similar patterns here and there, but you have to be aware that what we do in other places is unfortunately not always what we should actually do. We want to move common things to common package. Both useradm and tenantadm will be synchronizing data with HubSpot, so it make sense to have hubspot client in the common package, but I think having all models from all services in common package is a bit of oversharing and I'd rather create representation of hubspot user object in hubspot client. User model IMO should stay in useradm service. If you are not convinced we can have a call with Frode or Alf (or both if you like).

@kjaskiewiczz Maybe you have misunderstood the intention of this change, I'm not proposing putting all the models in every service into the common pkg. Let my clarify:
What I'm proposing is to move **some** of the models that is used by multiple services, like user and tenant model, into common pkg. The reason for this decision is to decouple the shared packages in `backend/pkg` form the services (e.g. not importing services in pkg), and to not have the same definition in multiple places. 
I'm also proposing to move just **parts** of the models to the common pkg and not the whole model. You can think of the models in pkg/common as the base for the model, and each services can import it and extend the functionality of it. For example the user struct has `Etag` and `Password` fields, but they are only relevant in the context of useradm, so useradm extend the user struct in the scope of the service.

There is another option to solve this problem which is to use the generated api models as common models. The drawback of this that we would not have the flexibility of extending these models, like adding methods or editing the json tags, and fields in generated models does not have bson tag.

I would also like to hear you thougths @frodeha about this change

frodeha:
> I would also like to hear you thougths @frodeha about this change

Happy to oblige! 

TLDR: I'm generally speaking quite skeptical to this 🙃 😅 

The definition of what a Mender User and Mender Tenant _is_ belongs squarely in `useradm` and `tenantadm` in my opinion. Moving parts of/entire models to a shared location doesn't sound appealing to me at all. I can easily imagine situations where we will loose track of the side-effects of working with to such shared definitions. Making an extreme example; an innocent change in how we store user data can take down all of Marketings HubSpot automation or (even worse) a tweak in how we send data to HubSpot can easily result in invalid data being stored to the database (which would cause real havoc).

Further more, if the HubSpot client (for whatever reason) needs to inherit the domain types of useradm / tenantadm (although I'm also skeptical to this), it should be HubSpot depending on useradm / tenantadm - not vice versa. If something needs to go down unintentionally, it should be our internal Marketing stuff - not Mender production.

In my opinion, the preferred solution is to have separate types in useradm and the HubSpot client (e.g what KJ has done [here](https://github.com/mendersoftware/mender-server-enterprise/pull/1078/changes#diff-97546519fffe9d6353daf4c8f38f25474a1833cd806a0503ae617c95087d9922)), precisely because they are _not_ the same definitions (at least not semantically) as I see it. They have very different reasons to change  and should therefore be allowed to do so without impacting each other (e.g the "S" in SOLID). Translating from one to the other (which would be the job of useradm and tenantadm btw) is perhaps tedious (but hardly difficult) and will (imo) result in better separation of concerns and a more robust solution in the end.

bahaa-ghazal:
Alright, I do agree with most of what you said and I don't have hard feelings about this change. Let's do it normally then. I'll wait for https://github.com/mendersoftware/mender-server-enterprise/pull/1078 to be merged before copying the client.

PR #1529, opened 2026-03-20
fix(gui): made ai analysis depend on ai feature activation

mzedel:


oldgiova (approved):
This is for not showing china AI banner I guess. LGTM for what I can understand here

PR #1530, opened 2026-03-20
MEN-9504 - plan notification adjustment

mzedel:
Screenshot 2026-03-20 at 14 54 03
Screenshot 2026-03-20 at 14 54 19

### DIFF
--- a/frontend/src/js/common-ui/EnterpriseNotification.tsx
+++ b/frontend/src/js/common-ui/EnterpriseNotification.tsx
@@ -15,8 +15,8 @@ import { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { Link } from 'react-router-dom';
 
-import { Chip } from '@mui/material';
-import { withStyles } from 'tss-react/mui';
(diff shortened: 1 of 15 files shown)


chiachenglu:
@mzedel Thank you for the quick act on this! Does this chip has the enable, hover, focus and pressed state definitions (like [here](https://www.figma.com/design/e2WPEGusQtx3JnKtOO2san/%F0%9F%8E%B8-UI--Mender-Design-System?node-id=11033-144922&t=dJ4v54xr418wTPMw-4))? 

Screenshot 2026-03-23 at 09 47 58

mzedel:
Thanks for taking a look @chiachenglu! This changes the `Chip` component used to show this to the standard one, so it uses the same settings as the theme, IOW it should look as defined in the theme 👌.

chiachenglu (approved):
Looks good to me! 
https://github.com/user-attachments/assets/8974649d-94cd-4373-ab72-d320bfbc5e0c

PR #1534, opened 2026-03-20
MEN-9413 - releases are now software

mzedel:


frontend/src/js/config/routes.tsx
```diff
@@ -33,6 +33,15 @@ import { TenantPage } from '../components/tenants/TenantPage';
...
 type RouteConfigs = Record<string, RouteConfig>;
 
+const ReleaseFallback = () => {
+  const { search } = useLocation();
+  const { selectedRelease } = useParams();
+  const navigate = useNavigate();
+
+  navigate(`/software/${selectedRelease}${search}`);
+  return null;
+};
```
mineralsfree, line 43: maybe 
```const ReleaseFallback = () => {
  const { search } = useLocation();
  const { selectedRelease } = useParams();
  return <Navigate to={`/software/${selectedRelease}${search}`} replace />;
};

File: frontend/src/js/components/deployments/__snapshots__/Report.test.tsx.snap [hunk 2]
```diff
@@ -1358,7 +1358,7 @@ Devices that are in the middle of the deployment at the time of abort will finis
                 <a
                   class="flexbox centered"
                   data-discover="true"
-                  href="/releases/undefined"
```
mineralsfree, line 1361: 😄

mineralsfree (dismissed):
Looks good!

frontend/src/js/config/routes.tsx
mzedel: Indeed I though so too, but followed the router docs: https://reactrouter.com/api/components/Navigate 🤷 - although they are opinionated sometimes 😄

mineralsfree: Then maybe it's worth wrapping the `navigate` in useEffect?

mineralsfree (approved):

PR #1538, opened 2026-03-23
MEN-9306 - device identity attribute settings

mzedel:
Screenshot 2026-03-23 at 13 34 07
Screenshot 2026-03-23 at 13 34 26

### DIFF
--- a/frontend/src/js/common-ui/DocsLink.tsx
+++ b/frontend/src/js/common-ui/DocsLink.tsx
@@ -15,7 +15,7 @@ import { ReactNode, forwardRef, useState } from 'react';
 import { useSelector } from 'react-redux';
 
 import { Description as DescriptionIcon, Launch as LaunchIcon } from '@mui/icons-material';
-import { Chip, Collapse, Typography, chipClasses } from '@mui/material';
+import { Chip, Collapse, Typography, TypographyProps, chipClasses } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 
 import { TIMEOUTS } from '@northern.tech/store/constants';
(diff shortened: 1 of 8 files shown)


frontend/src/js/common-ui/DocsLink.tsx
```diff
@@ -122,20 +122,21 @@ interface DocsTextLinkProps {
...
   const { path } = DOCSTIPS[id];
   return (
     <DocsLink path={path} {...props}>
-      {children || (
-        <Typography className={`inline ${capitalizedStart ? 'capitalized-start' : ''}`} color="primary">
-          learn more
-        </Typography>
-      )}
+      <Typography className={`inline ${capitalizedStart ? 'capitalized-start' : ''}`} color="primary" {...typographyProps}>
+        {children || 'learn more'}
```
mineralsfree, line 138: Maybe we should restrict `children` to string (in TS) to avoid rendering non-inline elements here? Or even rely on `title` prop in `DocsLink` that seems to be a good candidate. Since it's a `DocsTextLink` I think we shouldn't render anything by text?

mineralsfree (approved):

frontend/src/js/common-ui/DocsLink.tsx
mzedel: That's more due to the changing ways the link is used - sometimes it's just text and sometimes it's with an icon... I don't yet see a clear direction here 😕 - the distinction between regular `DocsLink` and `DocsTextLink` is more an _old_ vs _new_ (with old being the "majority icon based" one, whereas new is "majority text based")

PR #1540, opened 2026-03-23
Setup/teardown compose environment in next generation integration tests

frodeha:
## Description

This PR takes inspiration from https://github.com/mendersoftware/mender-server/pull/1386 (credit to Manuel for the original initiative 🙏) and introduces setup and teardown functionality in the next generation integration tests that don't depend on external scripts (bash or otherwise). The express goal of this is to enable a seamless and robust experience running the tests, reduce external dependencies and establish a practical and efficient way of working with and expanding them - also when it comes to the os -> enterprise sync.

There's, as always, many more things we can do, but I'm at the point where I really would love some input and feedback on what's here already and maybe we can build upon the general structure established here as we move forward.

_**Note: I'm aware that the merge to enterprise is going to be painful for this one. I am happy to take that on though!**_

### Points of note
The integration tests now use the Docker Compose- and Docker CLI Go SDKs to manage the docker compose environment used by the tests - e.g we are no longer executing `docker compose ...` (from bash or Go). Instead we are interacting directly with the Docker Engine API (which is what `docker compose` does). As far as possible I have tried to not depend on string commands being passed to Exec and clumsy stdout parsing. 

The integration tests now run outside the docker compose environment (not inside it as it did before). This comes with several benefits:
* The tests and the server environment is de-coupled. You can trivially run these tests against _any_ Mender server environment (in fact, I added an argument for doing just that). This is immensely helpful when writing the tests, because you can simply run them against the same docker compose environment where you did your implementation (no need for a rebuild).
* The tests execute _a lot_ faster. Taking the time consuming docker build of the test container (and therefore forced docker compose setup process) out of the process speeds up execution time (and therefore iteration time) massively.
* Debugging. You can now easily debug the tests using a normal debugger (eg. delve or what have you). It's so useful to be able to attach a debugger and step through every aspect of the test code. This has been (in my experience at least) _very_ clumsy (I've actually never been able to get it to work) before and is (at least in my view) a game changer.

There's also some downsides to it of course:
* Internal APIs are now essentially inaccessible. I have been thinking for quite a while on whether or not this is a deal breaker and I've come to the conclusion that I think it's not a deal breaker. My thinking is essentially that we shouldn't really need to use the internal APIs for integration tests. In my view, integration tests are external by nature and if we need to use an internal API to facilitate a test, then we're not testing the system in a way that anyone external will ever use it (e.g because they can't use the internal APIs) and the test case should probably be reconsidered in that case. Further more, if it turns out that I'm wrong about this assumption, I can think of ways to make this method work even so, but I would rather cross that bridge if we get to it.
* Portability is "technically" lowered (e.g, you now require docker _and_ Go to run the tests, not just docker). Although true, I don't particularly care about it. Anyone working on Mender needs to have Go installed anyway and if someone, for whatever reason, _just_ wants to run the integration tests I'm okay with them needing to install Go to do so. The upsides to this approach more than justifies it (imo).

I have attempted to establish a structure that should reduce open source -> enterprise merge issues. Essentially, open source specific implementation is put in dedicated `*_opensource_test.go` files. The thinking is that we will add `*_enterprise_test.go` versions over in enterprise. This also goes to the tests themselves, open source suites should be in `_opensource_test.go` files (as was already the case). I had an idea of using build tags to compile different implementations of the same functions, but that became a bit more clumsy to work with over in enterprise so I went for the approach in this PR. Happy to hear other / better solutions though!

There's several custom test flags parsed in TestMain. Every time I needed something a bit custom during the implementation I added a flag for it as I suppose it could be useful for others. Some of these flags are sharable between opensource and enterprise, but some are not (external tenants is a good example). I would love suggestions on how to do that better as right now it's not really workable with enterprise without substantially changing the flag parsing over in enterprise (which will result in more conflicts during sync, and I'd like to reduce that as much as we can).

### Examples of running the tests
```bash
$ cd backend

# Run the tests normally
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests
=== RUN   TestBackendIntegrationSuite
    backend_integration_test.go:80: Creating docker compose environment
    backend_integration_test.go:106: Successfully created docker compose environment
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf
=== NAME  TestBackendIntegrationSuite
    backend_integration_test.go:228: Tearing down project: backend-tests
--- PASS: TestBackendIntegrationSuite (14.91s)
    --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1 (0.11s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin (0.05s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf (0.00s)
PASS
ok      github.com/mendersoftware/mender-server/tests/runner/tests      15.445s

# Run a single test 
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests -run '^TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin$'
=== RUN   TestBackendIntegrationSuite
    backend_integration_test.go:80: Creating docker compose environment
    backend_integration_test.go:106: Successfully created docker compose environment
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin
=== NAME  TestBackendIntegrationSuite
    backend_integration_test.go:228: Tearing down project: backend-tests
--- PASS: TestBackendIntegrationSuite (14.94s)
    --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1 (0.10s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin (0.05s)
PASS
ok      github.com/mendersoftware/mender-server/tests/runner/tests      15.506s

# Run the tests but don't tear down compose environment
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests -skip-cleanup
=== RUN   TestBackendIntegrationSuite
    backend_integration_test.go:80: Creating docker compose environment
    backend_integration_test.go:106: Successfully created docker compose environment
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf
--- PASS: TestBackendIntegrationSuite (13.43s)
    --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1 (0.11s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin (0.05s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf (0.00s)
PASS
ok      github.com/mendersoftware/mender-server/tests/runner/tests      14.098s

$ docker ps --format 'table {{.Names}}'
NAMES
backend-tests-deployments-1
backend-tests-s3-1
backend-tests-s3-filer-1
backend-tests-create-artifact-worker-1
backend-tests-workflows-worker-1
backend-tests-deviceauth-1
backend-tests-s3-volume-1
backend-tests-workflows-1
backend-tests-deviceconnect-1
backend-tests-iot-manager-1
backend-tests-inventory-1
backend-tests-deviceconfig-1
backend-tests-useradm-1
backend-tests-traefik-1
backend-tests-nats-1
backend-tests-s3-master-1
backend-tests-mongo-1
backend-tests-gui-1

# Run the tests and don't tear down compose environment if the tests fail
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests -skip-cleanup-on-failure
... more or less same as above...

# Run the tests and inspect containers and print logs of services on failure
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests -print-service-status-on-failure=useradm,s3 -print-service-logs-on-failure="*"
... more or less same as above...

# Run the tests with externally managed server
$ go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests -server-url=docker.mender.io -usernames=user@docker.mender.io -passwords=$PASSWORD
=== RUN   TestBackendIntegrationSuite
    backend_integration_test.go:70: Running tests against external server: docker.mender.io
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser
=== RUN   TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf
--- PASS: TestBackendIntegrationSuite (0.20s)
    --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1 (0.20s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestLogin (0.05s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestMe (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestRemoveUser (0.00s)
        --- PASS: TestBackendIntegrationSuite/TestUseradmManagementV1/TestSelf (0.00s)
PASS
ok      github.com/mendersoftware/mender-server/tests/runner/tests      0.607s
```

Ticket: None


backend/tests/runner/tests/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+		var tenant Tenant
+		for idx := range names {
+			tenant.Users = append(tenant.Users, User{
+				Username: names[idx],
+				Password: passwords[idx],
+			})
+		}
+		externalTenants = []Tenant{tenant}
+	}
```
frodeha, line 100: Would love some input on how to make this part better in terms of the os -> ent sync

mzedel: I'm not quite sure what part of the sync you see as problematic here, but my 2 cents on the creation would be to (theoretically) support multiple tenants in the OS configuration already, since you rely on them being a list (in `i.settings.Tenants[0].Users`).
This way you define the users not as a list of usernames & passwords, but as a list of tenant configurations (maybe `<orgName>,<username>,<password>`) or go straight to a config/ seed file to define each of them.

But maybe I also miss the problem you foresaw entirely 😅.

frodeha: Yeah, that's one option. We're also going to need service provider tenants, but we could also just support that in open source. 

I suppose where I would love some input is:
a) Is having this (e.g both parsing of enterprise only entities and the entities themselves - Tenant in this case) in open source something we want / are ok with? I would love to not have this out of principle, but maybe this is where principle should/could give way to practice.
b) If we are okay with just having all of this in open source, then the os -> enterprise sync shouldn't really pose any problems. However, if that's not the case,  I suppose I'm looking for practical and clever suggestions on how to have two separate flag and parsing logics that don't step on each other in terms of sync.
c) Irrespective of a) and b), the reading of the seed arguments is going to get awkward in enterprise (it already is really, but it gets worse with multiple tenants, service provider -> children relationships and users). You indeed touch upon it above:
> "(maybe `<orgName>,<username>,<password>`) or go straight to a config/ seed file to define each of them". 

Defining one entity per flag instead of one entity property per flag I admit is better. That being said, I get a sense that maybe a config/seed file is the way to go. It's really the only way I can see of defining a hierarchy structure in a somewhat sane way... Would we be okay with something like that here? Or do you have better suggestions?

backend/tests/runner/tests/backend_integration_opensource_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"runtime/debug"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/cli/cli"
+	"github.com/docker/compose/v5/pkg/api"
+	"github.com/google/uuid"
+)
+
+func getDockerComposeFilesOpenSource() []string {
+	return []string{"../../../../docker-compose.yml"}
```
merlin-northern: magic string, this should be configurable either cmd arg or env with a default eq to `paths.Join(ROOTDIR,"docker-compose.yml")`. reasons:
* ../ relative paths in this case is not portable
* the location of docker compose file may be different

backend/tests/runner/tests/backend_integration_opensource_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func seedComposeEnvironmentOpenSource(ctx context.Context, compose api.Compose, project *types.Project) ([]Tenant, error) {
```
merlin-northern: naming: all this is on purpose OpneSource? so you envision renaming in enterprise? or just rewritting? I belive that syncing this is pointless. let's have one for open source and one for enterprise, in which case the interface should be without os/ent distinction.

backend/tests/runner/tests/backend_integration_opensource_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func isOpenSource() bool {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return false
+	}
+
+	return info.Main.Path == "github.com/mendersoftware/mender-server"
+}
```
merlin-northern: is there a reason you want to detect it? I find it pointless. ent has its own version, and os its own. this file has opensource in the name.

backend/tests/runner/tests/main_test.go
mzedel: IMO we have a somewhat single tenant setup in OS, even some of the addons we often only consider as optional in the plans ([troubleshoot](https://github.com/mendersoftware/deviceconnect) + [configure](https://github.com/mendersoftware/deviceconfig) both started with an OS core/ are available for in both OS + ENT).
So I'd say: seed file with a format that's helpful for both OS and ENT sounds good to me! I guess we could even use it in the e2e tests...

backend/tests/runner/tests/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if serverURL != "" {
+		externalServerURL = serverURL
+		names := strings.Split(usernames, ",")
+		passwords := strings.Split(passwords, ",")
+
+		if len(names) == 0 {
+			log.Printf("At least one username and password is required")
+			os.Exit(1)
+		}
```
merlin-northern, line 85: you use create-org so why do we require one user/pass pair?

backend/tests/runner/tests/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+		var tenant Tenant
+		for idx := range names {
+			tenant.Users = append(tenant.Users, User{
+				Username: names[idx],
+				Password: passwords[idx],
+			})
+		}
+		externalTenants = []Tenant{tenant}
```
merlin-northern, line 99: please consider just separate ent framework. we can separate the interface and replace the implementation of the calls, but these are tests, lets not get crazy. the setup for os is just different than in ent, so let's just keep two separate implementations.

.gitlab-ci.yml
```diff
@@ -561,6 +561,10 @@ test:backend:integration:ng:
   extends: .template:test:backend:integration
   variables:
     MANTRA_PROJECT_NAME: "backend_integration_ng_open_source"
+  before_script:
+    - !reference [.template:test:backend:integration, before_script]
+    - apk add go
```
merlin-northern, line 589: version considerations: let's think for a second what go version we are getting? (that's why I used the wget-tar approach so we can control the version. something to consider perhaps at one point.

backend/Makefile
```diff
@@ -116,7 +116,7 @@ test-integration:
 
 .PHONY: test-integration-ng
 test-integration-ng:
-	./tests/runner/scripts/run.sh $(args)
+	go test -count=1 -v github.com/mendersoftware/mender-server/tests/runner/tests
```
merlin-northern, line 119: I always run with count=1 😘

backend/go.mod
```diff
@@ -14,13 +14,17 @@ require (
...
 	github.com/aws/smithy-go v1.24.1
+	github.com/compose-spec/compose-go/v2 v2.10.1
+	github.com/docker/cli v29.2.1+incompatible
 	github.com/gin-gonic/gin v1.12.0
 	github.com/go-ozzo/ozzo-validation/v4 v4.3.0
 	github.com/go-viper/mapstructure/v2 v2.5.0
 	github.com/golang-jwt/jwt/v4 v4.5.2
 	github.com/google/uuid v1.6.0
-	github.com/gorilla/websocket v1.5.3
+	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674
```
merlin-northern, line 25: why has this changed?

backend/Makefile
merlin-northern: I also was thinking about this: (out of scope perhaps here) we have already did download the whole vendor in the previous phases. how could we re-use this cache? this will speed up greatly the execution. this step will run go mod vendor and get everything (again). this is a wider discussion.

backend/go.mod
merlin-northern, line 1: please run `go mod vendor; go mod tidy`
when I did:
```diff
diff --git a/backend/go.mod b/backend/go.mod
index 5d1dc75e..4c233e6b 100644
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/aws/smithy-go v1.24.1
 	github.com/compose-spec/compose-go/v2 v2.10.1
 	github.com/docker/cli v29.2.1+incompatible
+	github.com/docker/compose/v5 v5.1.1
 	github.com/gin-gonic/gin v1.12.0
 	github.com/go-ozzo/ozzo-validation/v4 v4.3.0
 	github.com/go-viper/mapstructure/v2 v2.5.0
@@ -93,7 +94,6 @@ require (
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/distribution/reference v0.6.0 // indirect
 	github.com/docker/buildx v0.31.1 // indirect
-	github.com/docker/compose/v5 v5.1.1 // indirect
 	github.com/docker/docker v28.5.2+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.9.5 // indirect
 	github.com/docker/go-connections v0.6.0 // indirect
diff --git a/backend/go.sum b/backend/go.sum
index d407051b..339e819c 100644
--- a/backend/go.sum
+++ b/backend/go.sum
@@ -21,7 +21,6 @@ github.com/DefangLabs/secret-detector v0.0.0-20250403165618-22662109213e h1:rd4b
 github.com/DefangLabs/secret-detector v0.0.0-20250403165618-22662109213e/go.mod h1:blbwPQh4DTlCZEfk1BLU4oMIhLda2U+A840Uag9DsZw=
 github.com/Microsoft/go-winio v0.6.2 h1:F2VQgta7ecxGYO8k3ZZz3RS8fVIXVxONVUPlNERoyfY=
 github.com/Microsoft/go-winio v0.6.2/go.mod h1:yd8OoFMLzJbo9gZq8j5qaps8bJ9aShtEA8Ipt1oGCvU=
-github.com/Microsoft/hcsshim v0.13.0 h1:/BcXOiS6Qi7N9XqUcv27vkIuVOkBEcWstd2pMlWSeaA=
 github.com/Microsoft/hcsshim v0.14.0-rc.1 h1:qAPXKwGOkVn8LlqgBN8GS0bxZ83hOJpcjxzmlQKxKsQ=
 github.com/Microsoft/hcsshim v0.14.0-rc.1/go.mod h1:hTKFGbnDtQb1wHiOWv4v0eN+7boSWAHyK/tNAaYZL0c=
 github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d h1:licZJFw2RwpHMqeKTCYkitsPqHNxTmd4SNR5r94FGM8=
@@ -30,7 +29,6 @@ github.com/alicebob/gopher-json v0.0.0-20230218143504-906a9b012302 h1:uvdUDbHQHO
 github.com/alicebob/gopher-json v0.0.0-20230218143504-906a9b012302/go.mod h1:SGnFV6hVsYE877CKEZ6tDNTjaSXYUk6QqoIK6PrAtcc=
 github.com/alicebob/miniredis v2.5.0+incompatible h1:yBHoLpsyjupjz3NL3MhKMVkR41j82Yjf3KFv7ApYzUI=
 github.com/alicebob/miniredis v2.5.0+incompatible/go.mod h1:8HZjEj4yU0dwhYHky+DxYx+6BMjkBbe5ONFIF1MXffk=
-github.com/anchore/go-struct-converter v0.0.0-20221118182256-c68fdcfa2092 h1:aM1rlcoLz8y5B2r4tTLMiVTrMtpfY0O8EScKJxaSaEc=
 github.com/anchore/go-struct-converter v0.1.0 h1:2rDRssAl6mgKBSLNiVCMADgZRhoqtw9dedlWa0OhD30=
 github.com/anchore/go-struct-converter v0.1.0/go.mod h1:rYqSE9HbjzpHTI74vwPvae4ZVYZd1lue2ta6xHPdblA=
 github.com/antithesishq/antithesis-sdk-go v0.6.0-default-no-op h1:kpBdlEPbRvff0mDD1gk7o9BhI16b9p5yYAXRlidpqJE=
@@ -107,17 +105,12 @@ github.com/codahale/rfc6979 v0.0.0-20141003034818-6a90f24967eb h1:EDmT6Q9Zs+SbUo
 github.com/codahale/rfc6979 v0.0.0-20141003034818-6a90f24967eb/go.mod h1:ZjrT6AXHbDs86ZSdt/osfBi5qfexBrKUdONk989Wnk4=
 github.com/compose-spec/compose-go/v2 v2.10.1 h1:mFbXobojGRFIVi1UknrvaDAZ+PkJfyjqkA1yseh+vAU=
 github.com/compose-spec/compose-go/v2 v2.10.1/go.mod h1:Ohac1SzhO/4fXXrzWIztIVB6ckmKBv1Nt5Z5mGVESUg=
-github.com/containerd/cgroups/v3 v3.0.5 h1:44na7Ud+VwyE7LIoJ8JTNQOa549a8543BmzaJHo6Bzo=
 github.com/containerd/cgroups/v3 v3.1.2 h1:OSosXMtkhI6Qove637tg1XgK4q+DhR0mX8Wi8EhrHa4=
 github.com/containerd/cgroups/v3 v3.1.2/go.mod h1:PKZ2AcWmSBsY/tJUVhtS/rluX0b1uq1GmPO1ElCmbOw=
 github.com/containerd/console v1.0.5 h1:R0ymNeydRqH2DmakFNdmjR2k0t7UPuiOV/N/27/qqsc=
 github.com/containerd/console v1.0.5/go.mod h1:YynlIjWYF8myEu6sdkwKIvGQq+cOckRm6So2avqoYAk=
 github.com/containerd/containerd/api v1.10.0 h1:5n0oHYVBwN4VhoX9fFykCV9dF1/BvAXeg2F8W6UYq1o=
 github.com/containerd/containerd/api v1.10.0/go.mod h1:NBm1OAk8ZL+LG8R0ceObGxT5hbUYj7CzTmR3xh0DlMM=
-github.com/containerd/containerd/v2 v2.1.6 h1:sc9Yoeb6UYgT96nEcnEljDzF5ZVkXcpoWyR3b+Xt8UY=
-github.com/containerd/containerd/v2 v2.1.6/go.mod h1:obFwwiJzy/EyTsXYtJgTAv/V/n7Z0JhZn3aXeHONVRI=
-github.com/containerd/containerd/v2 v2.2.1 h1:TpyxcY4AL5A+07dxETevunVS5zxqzuq7ZqJXknM11yk=
-github.com/containerd/containerd/v2 v2.2.1/go.mod h1:NR70yW1iDxe84F2iFWbR9xfAN0N2F0NcjTi1OVth4nU=
 github.com/containerd/containerd/v2 v2.2.2 h1:mjVQdtfryzT7lOqs5EYUFZm8ioPVjOpkSoG1GJPxEMY=
 github.com/containerd/containerd/v2 v2.2.2/go.mod h1:5Jhevmv6/2J+Iu/A2xXAdUIdI5Ah/hfyO7okJ4AFIdY=
 github.com/containerd/continuity v0.4.5 h1:ZRoN1sXq9u7V6QoHMcVWGhOwDFqZ4B9i5H6un1Wh0x4=
@@ -130,16 +123,13 @@ github.com/containerd/fifo v1.1.0 h1:4I2mbh5stb1u6ycIABlBw9zgtlK8viPI9QkQNRQEEmY
 github.com/containerd/fifo v1.1.0/go.mod h1:bmC4NWMbXlt2EZ0Hc7Fx7QzTFxgPID13eH0Qu+MAb2o=
 github.com/containerd/log v0.1.0 h1:TCJt7ioM2cr/tfR8GPbGf9/VRAX8D2B4PjzCpfX540I=
 github.com/containerd/log v0.1.0/go.mod h1:VRRf09a7mHDIRezVKTRCrOq78v577GXq3bSa3EhrzVo=
-github.com/containerd/nydus-snapshotter v0.15.2 h1:qsHI4M+Wwrf6Jr4eBqhNx8qh+YU0dSiJ+WPmcLFWNcg=
 github.com/containerd/nydus-snapshotter v0.15.10 h1:hphjuKOqSHLGznNJiAvmsOWkdu4qFXjf4DzGrWSuIsM=
 github.com/containerd/nydus-snapshotter v0.15.10/go.mod h1:EWRd/QJ0b6UKHAqYgiV5gHlqLC2qq5cQiSlXEdVovrA=
 github.com/containerd/platforms v1.0.0-rc.2 h1:0SPgaNZPVWGEi4grZdV8VRYQn78y+nm6acgLGv/QzE4=
 github.com/containerd/platforms v1.0.0-rc.2/go.mod h1:J71L7B+aiM5SdIEqmd9wp6THLVRzJGXfNuWCZCllLA4=
 github.com/containerd/plugin v1.0.0 h1:c8Kf1TNl6+e2TtMHZt+39yAPDbouRH9WAToRjex483Y=
 github.com/containerd/plugin v1.0.0/go.mod h1:hQfJe5nmWfImiqT1q8Si3jLv3ynMUIBB47bQ+KexvO8=
-github.com/containerd/stargz-snapshotter v0.16.3 h1:zbQMm8dRuPHEOD4OqAYGajJJUwCeUzt4j7w9Iaw58u4=
 github.com/containerd/stargz-snapshotter v0.17.0 h1:djNS4KU8ztFhLdEDZ1bsfzOiYuVHT6TgSU5qwRk+cNc=
-github.com/containerd/stargz-snapshotter/estargz v0.16.3 h1:7evrXtoh1mSbGj/pfRccTampEyKpjpOnS3CyiV1Ebr8=
 github.com/containerd/stargz-snapshotter/estargz v0.17.0 h1:+TyQIsR/zSFI1Rm31EQBwpAA1ovYgIKHy7kctL3sLcE=
 github.com/containerd/stargz-snapshotter/estargz v0.17.0/go.mod h1:s06tWAiJcXQo9/8AReBCIo/QxcXFZ2n4qfsRnpl71SM=
 github.com/containerd/ttrpc v1.2.7 h1:qIrroQvuOL9HQ1X6KHe2ohc7p+HP/0VE6XPU7elJRqQ=
@@ -153,7 +143,6 @@ github.com/creack/pty v1.1.24 h1:bJrF4RRfyJnbTJqzRLHzcGaZK1NeM5kTC9jGgovnR1s=
 github.com/creack/pty v1.1.24/go.mod h1:08sCNb52WyoAwi2QDyzUCTgcvVFhUzewun7wtTfvcwE=
 github.com/cyberphone/json-canonicalization v0.0.0-20241213102144-19d51d7fe467 h1:uX1JmpONuD549D73r6cgnxyUu18Zb7yHAy5AYU0Pm4Q=
 github.com/cyberphone/json-canonicalization v0.0.0-20241213102144-19d51d7fe467/go.mod h1:uzvlm1mxhHkdfqitSA92i7Se+S9ksOn3a3qmv/kyOCw=
-github.com/cyphar/filepath-securejoin v0.5.1 h1:eYgfMq5yryL4fbWfkLpFFy2ukSELzaJOTaUTuh+oF48=
 github.com/cyphar/filepath-securejoin v0.6.0 h1:BtGB77njd6SVO6VztOHfPxKitJvd/VPT+OFBFMOi1Is=
 github.com/cyphar/filepath-securejoin v0.6.0/go.mod h1:A8hd4EnAeyujCJRrICiOWqjS1AX0a9kM5XL+NwKoYSc=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
@@ -170,14 +159,10 @@ github.com/distribution/reference v0.6.0 h1:0IXCQ5g4/QMHHkarYzh5l+u8T3t73zM5Qvfr
 github.com/distribution/reference v0.6.0/go.mod h1:BbU0aIcezP1/5jX/8MP0YiH4SdvB5Y4f/wlDRiLyi3E=
 github.com/dlclark/regexp2 v1.11.0 h1:G/nrcoOa7ZXlpoa/91N3X7mM3r8eIlMBBJZvsz/mxKI=
 github.com/dlclark/regexp2 v1.11.0/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
-github.com/docker/buildx v0.29.1 h1:58hxM5Z4mnNje3G5NKfULT9xCr8ooM8XFtlfUK9bKaA=
-github.com/docker/buildx v0.29.1/go.mod h1:J4EFv6oxlPiV1MjO0VyJx2u5tLM7ImDEl9zyB8d4wPI=
 github.com/docker/buildx v0.31.1 h1:zbvbrb9nxBNVV8nnI33f2F+4aAZBA1gY+AmeBFflMqY=
 github.com/docker/buildx v0.31.1/go.mod h1:SD+jYLnt3S4SXqohVtV+8z+dihnOgwMJ8t+bLQvsaCk=
 github.com/docker/cli v29.2.1+incompatible h1:n3Jt0QVCN65eiVBoUTZQM9mcQICCJt3akW4pKAbKdJg=
 github.com/docker/cli v29.2.1+incompatible/go.mod h1:JLrzqnKDaYBop7H2jaqPtU4hHvMKP+vjCwu2uszcLI8=
-github.com/docker/compose/v5 v5.1.0 h1:HofsoOEJZSAjXzzeGxOub3WK/uEAL0v7g7J1kgmYETY=
-github.com/docker/compose/v5 v5.1.0/go.mod h1:vpHFFnEFlKTz9HjSkRqQz8x77ZuoMoq6CN1VicxFeQM=
 github.com/docker/compose/v5 v5.1.1 h1:Ngkcd7AtHra5TcEEkDWol9qcZ44pMHJiM7C3dzWcQPQ=
 github.com/docker/compose/v5 v5.1.1/go.mod h1:9MiC31ueo0VamH4PvTAPfUrd3lQHIWQYgww2yDB8Dx8=
 github.com/docker/distribution v2.8.3+incompatible h1:AtKxIZ36LoNK51+Z6RpzLpddBirtxJnzDrHLEKxTAYk=
@@ -281,7 +266,6 @@ github.com/golang-jwt/jwt/v4 v4.5.2 h1:YtQM7lnr8iZ+j5q71MGKkNw9Mn7AjHM68uc9g5fXe
 github.com/golang-jwt/jwt/v4 v4.5.2/go.mod h1:m21LjoU+eqJr34lmDMbreY2eSTRJ1cv77w39/MY0Ch0=
 github.com/golang-jwt/jwt/v5 v5.3.0 h1:pv4AsKCKKZuqlgs5sUmn4x8UlGa0kEVt/puTpKx9vvo=
 github.com/golang-jwt/jwt/v5 v5.3.0/go.mod h1:fxCRLWMO43lRc8nhHWY6LGqRcf+1gQWArsqaEUEa5bE=
-github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da h1:oI5xCqsCo564l8iNU+DwB5epxmsaqB+rhGL0m5jtYqE=
 github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 h1:f+oWsMOmNPc8JmEHVZIycC7hBoQxHH9pNKQORJNozsQ=
 github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8/go.mod h1:wcDNUvekVysuuOpQKo3191zZyTpiI6se1N1ULghS0sw=
 github.com/golang/protobuf v1.5.4 h1:i7eJL8qZTpSEXOPTxNKhASYpMn+8e5Q6AdndVa1dWek=
@@ -305,8 +289,6 @@ github.com/google/uuid v1.6.0 h1:NIvaJDMOsjHA8n1jAhLSgzrAzy1Hgr+hNrb57e+94F0=
 github.com/google/uuid v1.6.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/gorilla/mux v1.8.1 h1:TuBL49tXwgrFYWhqrNgrUNEY92u81SPhu7sTdzQEiWY=
 github.com/gorilla/mux v1.8.1/go.mod h1:AKf9I4AEqPTmMytcMc0KkNouC66V3BtZ4qD5fmWSiMQ=
-github.com/gorilla/websocket v1.5.3 h1:saDtZ6Pbx/0u+bgYQ3q96pZgCzfhKXGPqt7kZ72aNNg=
-github.com/gorilla/websocket v1.5.3/go.mod h1:YR8l580nyteQvAITg2hZ9XVh4b55+EU/adAjf1fMHhE=
 github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674 h1:JeSE6pjso5THxAzdVpqr6/geYxZytqFMBCOtn/ujyeo=
 github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674/go.mod h1:r4w70xmWCQKmi1ONH4KIaBptdivuRPyosB9RmPlGEwA=
 github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.3 h1:NmZ1PKzSTQbuGHw9DGPFomqkkLWMC+vZCkfs+FHv1Vg=
@@ -366,8 +348,6 @@ github.com/minio/sha256-simd v1.0.1 h1:6kaan5IFmwTNynnKKpDHe6FWHohJOHhCPchzK49dz
 github.com/minio/sha256-simd v1.0.1/go.mod h1:Pz6AKMiUdngCLpeTL/RJY1M9rUuPMYujV5xJjtbRSN8=
 github.com/mitchellh/hashstructure/v2 v2.0.2 h1:vGKWl0YJqUNxE8d+h8f6NJLcCJrgbhC4NcD46KavDd4=
 github.com/mitchellh/hashstructure/v2 v2.0.2/go.mod h1:MG3aRVU/N29oo/V/IhBX8GR/zz4kQkprJgF2EVszyDE=
-github.com/moby/buildkit v0.25.2 h1:mReLKDPv05cqk6o/u3ixq2/iTsWGHoUO5Zg3lojrQTk=
-github.com/moby/buildkit v0.25.2/go.mod h1:phM8sdqnvgK2y1dPDnbwI6veUCXHOZ6KFSl6E164tkc=
 github.com/moby/buildkit v0.27.1 h1:qlIWpnZzqCkrYiGkctM1gBD/YZPOJTjtUdRBlI0oBOU=
 github.com/moby/buildkit v0.27.1/go.mod h1:99qLrCrIAFgEOiFnCi9Y0Wwp6/qA7QvZ3uq/6wF0IsI=
 github.com/moby/docker-image-spec v1.3.1 h1:jMKff3w6PgbfSa69GfNg+zN/XLhfXJGnEx3Nl2EsFP0=
@@ -376,12 +356,8 @@ github.com/moby/go-archive v0.2.0 h1:zg5QDUM2mi0JIM9fdQZWC7U8+2ZfixfTYoHL7rWUcP8
 github.com/moby/go-archive v0.2.0/go.mod h1:mNeivT14o8xU+5q1YnNrkQVpK+dnNe/K6fHqnTg4qPU=
 github.com/moby/locker v1.0.1 h1:fOXqR41zeveg4fFODix+1Ch4mj/gT0NE1XJbp/epuBg=
 github.com/moby/locker v1.0.1/go.mod h1:S7SDdo5zpBK84bzzVlKr2V0hz+7x9hWbYC/kq7oQppc=
-github.com/moby/moby/api v1.53.0 h1:PihqG1ncw4W+8mZs69jlwGXdaYBeb5brF6BL7mPIS/w=
-github.com/moby/moby/api v1.53.0/go.mod h1:8mb+ReTlisw4pS6BRzCMts5M49W5M7bKt1cJy/YbAqc=
 github.com/moby/moby/api v1.54.0 h1:7kbUgyiKcoBhm0UrWbdrMs7RX8dnwzURKVbZGy2GnL0=
 github.com/moby/moby/api v1.54.0/go.mod h1:8mb+ReTlisw4pS6BRzCMts5M49W5M7bKt1cJy/YbAqc=
-github.com/moby/moby/client v0.2.2 h1:Pt4hRMCAIlyjL3cr8M5TrXCwKzguebPAc2do2ur7dEM=
-github.com/moby/moby/client v0.2.2/go.mod h1:2EkIPVNCqR05CMIzL1mfA07t0HvVUUOl85pasRz/GmQ=
 github.com/moby/moby/client v0.3.0 h1:UUGL5okry+Aomj3WhGt9Aigl3ZOxZGqR7XPo+RLPlKs=
 github.com/moby/moby/client v0.3.0/go.mod h1:HJgFbJRvogDQjbM8fqc1MCEm4mIAGMLjXbgwoZp6jCQ=
 github.com/moby/patternmatcher v0.6.0 h1:GmP9lR19aU5GqSSFko+5pRqHi+Ohk1O69aFiKkVGiPk=
@@ -451,16 +427,12 @@ github.com/planetscale/vtprotobuf v0.6.1-0.20240319094008-0393e58bdf10/go.mod h1
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 h1:Jamvg5psRIccs7FGNTlIRMkT8wgtp5eCXdBlqhYGL6U=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
-github.com/prometheus/client_golang v1.22.0 h1:rb93p9lokFEsctTys46VnV1kLCDpVZ0a/Y92Vm0Zc6Q=
 github.com/prometheus/client_golang v1.23.2 h1:Je96obch5RDVy3FDMndoUsjAhG5Edi49h0RJWRi/o0o=
 github.com/prometheus/client_golang v1.23.2/go.mod h1:Tb1a6LWHB3/SPIzCoaDXI4I8UHKeFTEQ1YCr+0Gyqmg=
-github.com/prometheus/client_model v0.6.1 h1:ZKSh/rekM+n3CeS952MLRAdFwIKqeY8b62p8ais2e9E=
 github.com/prometheus/client_model v0.6.2 h1:oBsgwpGs7iVziMvrGhE53c/GrLUsZdHnqNwqPLxwZyk=
 github.com/prometheus/client_model v0.6.2/go.mod h1:y3m2F6Gdpfy6Ut/GBsUqTWZqCUvMVzSfMLjcu6wAwpE=
-github.com/prometheus/common v0.62.0 h1:xasJaQlnWAeyHdUBeGjXmutelfJHWMRr+Fg4QszZ2Io=
 github.com/prometheus/common v0.66.1 h1:h5E0h5/Y8niHc5DlaLlWLArTQI7tMrsfQjHV+d9ZoGs=
 github.com/prometheus/common v0.66.1/go.mod h1:gcaUsgf3KfRSwHY4dIMXLPV0K/Wg1oZ8+SbZk/HH/dA=
-github.com/prometheus/procfs v0.15.1 h1:YagwOFzUgYfKKHX6Dr+sHT7km/hxC76UB0learggepc=
 github.com/prometheus/procfs v0.17.0 h1:FuLQ+05u4ZI+SS/w9+BWEM2TXiHKsUQ9TADiRH7DuK0=
 github.com/prometheus/procfs v0.17.0/go.mod h1:oPQLaDAMRbA+u8H5Pbfq+dl3VDAvHxMUOVhe0wYB2zw=
 github.com/quic-go/qpack v0.6.0 h1:g7W+BMYynC1LbYLSqRt8PBg5Tgwxn214ZZR34VIOjz8=
@@ -502,7 +474,6 @@ github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 h1:JIAuq3EE
 github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966/go.mod h1:sUM3LWHvSMaG192sy56D9F7CNvL7jUJVXoqM1QKLnog=
 github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8 h1:+jumHNA0Wrelhe64i8F6HNlS8pkoyMv5sreGx2Ry5Rw=
 github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8/go.mod h1:3n1Cwaq1E1/1lhQhtRK2ts/ZwZEhjcQeJQ1RuC6Q/8U=
-github.com/spdx/tools-golang v0.5.5 h1:61c0KLfAcNqAjlg6UNMdkwpMernhw3zVRwDZ2x9XOmk=
 github.com/spdx/tools-golang v0.5.7 h1:+sWcKGnhwp3vLdMqPcLdA6QK679vd86cK9hQWH3AwCg=
 github.com/spdx/tools-golang v0.5.7/go.mod h1:jg7w0LOpoNAw6OxKEzCoqPC2GCTj45LyTlVmXubDsYw=
 github.com/spf13/afero v1.15.0 h1:b/YBCLWAJdFWJTN9cLhiXXcD7mzKn9Dm86dNnfyQw1I=
@@ -561,7 +532,6 @@ github.com/ulikunitz/xz v0.5.14 h1:uv/0Bq533iFdnMHZdRBTOlaNMdb1+ZxXIlHDZHIHcvg=
 github.com/ulikunitz/xz v0.5.14/go.mod h1:nbz6k7qbPmH4IRqmfOplQw/tblSgqTqBwxkY0oWt/14=
 github.com/urfave/cli v1.22.17 h1:SYzXoiPfQjHBbkYxbew5prZHS1TOLT3ierW8SYLqtVQ=
 github.com/urfave/cli v1.22.17/go.mod h1:b0ht0aqgH/6pBYzzxURyrM4xXNgsoT/n2ZzwQiEhNVo=
-github.com/vbatts/tar-split v0.12.1 h1:CqKoORW7BUWBe7UL/iqTVvkTBOF8UvOMKOIZykxnnbo=
 github.com/vbatts/tar-split v0.12.2 h1:w/Y6tjxpeiFMR47yzZPlPj/FcPLpXbTUi/9H7d3CPa4=
 github.com/vbatts/tar-split v0.12.2/go.mod h1:eF6B6i6ftWQcDqEn3/iGFRFRo8cBIMSJVOpnNdfTMFA=
 github.com/vmihailenco/msgpack/v5 v5.4.1 h1:cQriyiUvjTwOHg8QZaPihLWeRAAVoCpE00IUPn0Bjt8=
```

backend/go.mod
```diff
@@ -47,6 +51,10 @@ require (
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 // indirect
+	github.com/Azure/go-ansiterm v0.0.0-20250102033503-faa5f7b0171c // indirect
+	github.com/DefangLabs/secret-detector v0.0.0-20250403165618-22662109213e // indirect
+	github.com/Microsoft/go-winio v0.6.2 // indirect
+	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d // indirect
```
merlin-northern, line 58: I do not understand why added deps? has this been rebase on top of master?

merlin-northern (dismissed):
so I left some comments. I do not wish to block it, as we can adjust later. but those comments where something that I really had to say. feel free to ignore. and also: I allowed myself to be just plain and do not beat around the bush, please excuse the tone :) I did not mean anything bad! this is very lovely.

backend/tests/runner/tests/backend_integration_opensource_test.go
frodeha: In principle I agree with you. However, in practice I found this quite hard to accomplish in a way that I found to be superior to the existing way of using a relative path.

Essentially, it's not so easy to automatically figure out what ROOTDIR actually is (at least with a pure go solution). I can't hard-code it as that's not much better than using this relative path. Recursively ascending up the directory hierarchy to determine what directory is ROOTDIR (the one with .git/ in it I suppose?) or even pulling in git to do `git rev-parse --show-toplevel` seems like overkill. I also want to avoid/improve the situation we have in the python tests where "what compose files are actually used" is a pretty hard question to answer. I'm also somewhat struggling to see the scenarios where the location of the compose files will change _and_ we also won't just update this accordingly - but maybe I'm missing something. Something else we can potentially do is add support setting the `COMPOSEFILE` environment variable to make it configurable. 

All in all, unless you're insistent on this I kindly ask that we revisit this if/when it proves an actual problem.

backend/tests/runner/tests/backend_integration_opensource_test.go
frodeha: I was envisioning having different files with opensource and enterprise implementation of the same interfaces/functions and have both stick around so we didn't have to rewrite stuff as part of the sync. However, I have undone this and reverted to your suggestion for the most part.

backend/tests/runner/tests/main_test.go
frodeha: This is only required if you "bring your own compose stack" in which case the tests won't run `create-org` and you will have to specify at least one user yourself.

backend/go.mod
frodeha: Done

backend/go.mod
frodeha: Because we're adding a dependency on [github.com/docker/compose/v5@v5.1.1](https://github.com/docker/compose/tree/v5.1.1) which depends on [github.com/moby/buildx@v0.31.1](https://github.com/docker/buildx/blob/v0.31.1/go.mod#L11) which depends on [github.com/containerd/containerd/v2@v2.2.1 ](https://github.com/docker/buildx/blob/v0.31.1/go.mod#L11)which depends on [k8s.io/client-go@v0.34.1](https://github.com/containerd/containerd/blob/v2.2.1/go.mod#L87) which requires [github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674](https://github.com/kubernetes/client-go/blob/v0.34.1/go.mod#L15).

It's really a "a dependency of my dependency is my dependency" type situation...

backend/go.mod
frodeha: It's all dependencies of dependencies basically 🤷

merlin-northern (approved):
before we all go insane, let's have it and then see what the future brings.

PR #1541, opened 2026-03-23
ci: fix undesired `---` remove from the changelog

oldgiova:
The Changelog header includes `---` and requires it, but the changelog generation is removing them by mistake. This commit fixes the original behavior


kjaskiewiczz (approved):

oldgiova:
Rebased

PR #1542, opened 2026-03-24
chore(reporting): remove the service from the monorepo

alfrunes:
The service is unused and not functional in the current state. This commit cleans it up to remove this overhead until we will reintroduce reporting and dashboards.


frodeha:
@alfrunes - there's Reporting stuff (client, config, code branches) in (at least) deployments and deviceauth too 😇

frodeha (commented):
Not really blocking stuff, but I suppose we could remove it from the [commit lint config](https://github.com/mendersoftware/mender-server/blob/aa59ded7d5d22e396bd656f1ba2260d8cb647878/commitlint.config.js#L10) and [dependabot](https://github.com/mendersoftware/mender-server/blob/aa59ded7d5d22e396bd656f1ba2260d8cb647878/.github/dependabot.yml#L117) as well?

frodeha (approved):

PR #1583, opened 2026-03-27
chore(pkg): add openapi-generator config file and aligned package name

alfrunes:
The package name conflicts with the import path.
The configuration file also disables markdown documentation.


backend/pkg/api/.openapi-generator.yaml
```diff
@@ -0,0 +1,15 @@
+#
+# .openapi-generator-config.yml
+# - https://openapi-generator.tech
+#
+additionalProperties:
+  generateSourceCodeOnly: true
+  hideGenerationTimestamp: true
+  packageName: "client"
```
merlin-northern, line 8: for me it can even be `geronimo!`; client is ok too :> just to leave some comment: it wont be to generic/general? we import as openapi. it is a client for backend over openapi3. I do not know. ignore.

merlin-northern (approved):

backend/pkg/api/.openapi-generator.yaml
alfrunes: I just found it confusing when the package name is not aligned with the import path, and I didn't want to change the directory name, although `geronimo` is a nice package name ;)

PR #1584, opened 2026-03-27
QA-1510: Generate mocks for openapi generated client

alfrunes:
This PR adds mockery (v3.7.0 - current latest) to generate mocks for all the generated service interfaces from openapi-generator-cli. It also refactors the docker commands into a compose file to make it easier for the reader.


backend/Makefile
```diff
@@ -47,16 +47,7 @@ generate-openapi-spec: $(COMBINED_SPEC)
...
-	@rm -Rf "$(OAPI_CLIENT)"
-	@docker run -u $$(id -u):$$(id -g) --rm -v ${MAKEDIR}:/backend \
-		openapitools/openapi-generator-cli:${OPENAPI_GENERATOR_VERSION} \
-		generate \
-		-i /backend/"$(COMBINED_SPEC)" \
-		-g go \
-		-o /"$(OAPI_CLIENT)" \
-		-c /backend/pkg/api/.openapi-generator.yaml
-	@echo "Generated $(OAPI_CLIENT)"
+	@$(MAKE) -C pkg/api
```
merlin-northern: this will make copy-paste of the commands impossible. also this brings `OPENAPI_GENERATOR_VERSION` variable decoupled into two places -- which I also do not like. docker compose as a source of truth is something that I have certain opinions on -- which I will hold.

backend/pkg/api/docker-compose.yml
```diff
@@ -10,3 +10,8 @@ services:
       - -o=client
       - -g=go
       - -c=.openapi-generator.yaml
+  generate-mocks:
+    image: vektra/mockery:v3.7.0
```
merlin-northern: hard coding version, can we make it a var with default value if unset?

backend/Makefile
alfrunes: I would argue that the version of a code generator should be predictable and not variable. And putting it in a compose file should make it more straight forward to maintain using dependabot/renovate. But I can also put this all in the makefile and remove the dependency on the compose plugin if you like.

merlin-northern (approved):
thanks 😘 !

PR #1598, opened 2026-03-30
MEN-9470: Fix inventory statistics response

frodeha:
**Description**

Adjust the return type of the inventory status statistics to use integers instead of numbers. Using "number" results in the corresponding Go type being float32 instead of int32. 

I don't think there should be any difference (and therefore no breakage...) for the GUI as "everything is a number in the javascript world" so I'm trusting that this fix-up should be okay despite it _technically_ being a breaking change.

Ticket: MEN-9470


mzedel (dismissed):

backend/docs/api/dist/openapi.yaml
```diff
@@ -9528,11 +9528,11 @@ components:
         - system
       properties:
         standard:
-          type: number
+          type: integer
```
merlin-northern, line 9531: if you want to be strict and escape the int32() casts you can specify type: int64 (just a comment based on lots of casting I had to do lately)

merlin-northern (dismissed):
why dont we use this lovely opportunity to make a small tst about this endpoint in the small place known as backend integration tests-ng? why the tests that already call this endpoint did not react to those changes?

backend/docs/api/dist/openapi.yaml
frodeha: ~`int` also works! That's even better - thanks!~ well, it works for the code-gen but not for the schema validator... :sigh:

frodeha:
> why dont we use this lovely opportunity to make a small tst about this endpoint in the small place known as backend integration tests-ng?

Done! 

> why the tests that already call this endpoint did not react to those changes?

Because, ahem, well python 🤷

merlin-northern (approved):
lovely

PR #1599, opened 2026-03-31
ci: os/ent prefix to pr based deployment to avoid name collision

oldgiova:
Avoid PR Based deployment name collision when the PR number is the same in both OS and ENT repos

Ticket: QA-1524

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -224,20 +224,18 @@ review:deploy:
       aud: https://gitlab.com
   environment:
     name: review/${CI_COMMIT_REF_NAME}
-    url: https://${CI_COMMIT_REF_SLUG}.${REVIEW_APPS_DOMAIN}
+    url: https://${REVIEW_APPS_PROJECT_PREFIX}-${CI_COMMIT_REF_SLUG}.${REVIEW_APPS_DOMAIN}
     on_stop: review:stop
     auto_stop_in: 7 days
   variables:
-    NAMESPACE: mender-${CI_COMMIT_REF_SLUG}
+    NAMESPACE: mender-${REVIEW_APPS_PROJECT_PREFIX}-${CI_COMMIT_REF_SLUG}
   before_script:
     - echo "Deploying review app for branch ${CI_COMMIT_REF_NAME}"
-
-    # Generate short identifier to avoid Kubernetes 52 character name limit
-    # Use first 12 characters of branch slug hash for uniqueness
-    - export RELEASE_NAME="mender-$(echo -n ${CI_COMMIT_REF_SLUG} | md5sum | cut -c1-12)"
+    - test -n "${REVIEW_APPS_PROJECT_PREFIX}" || (echo "REVIEW_APPS_PROJECT_PREFIX is not set"; exit 1)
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -224,20 +224,22 @@ review:deploy:
       aud: https://gitlab.com
   environment:
     name: review/${CI_COMMIT_REF_NAME}
-    url: https://${CI_COMMIT_REF_SLUG}.${REVIEW_APPS_DOMAIN}
+    url: https://${CI_PROJECT_NAME}-${CI_COMMIT_REF_SLUG}.${REVIEW_APPS_DOMAIN}
```
mzedel: would this not also have to be:
```suggestion
    url: https://${REVIEW_APPS_PROJECT_PREFIX}-${CI_COMMIT_REF_SLUG}.${REVIEW_APPS_DOMAIN}
```
?

.gitlab-ci.yml
```diff
@@ -224,20 +224,22 @@ review:deploy:
...
-    # Generate short identifier to avoid Kubernetes 52 character name limit
-    # Use first 12 characters of branch slug hash for uniqueness
-    - export RELEASE_NAME="mender-$(echo -n ${CI_COMMIT_REF_SLUG} | md5sum | cut -c1-12)"
+    - case "${CI_PROJECT_NAME}" in
+        mender-server-enterprise) export REVIEW_APPS_PROJECT_PREFIX="ent" ;;
+        mender-server)            export REVIEW_APPS_PROJECT_PREFIX="os" ;;
+        *) echo "Unknown project '${CI_PROJECT_NAME}', cannot determine prefix"; exit 1 ;;
+      esac
+    - export NAMESPACE="mender-${REVIEW_APPS_PROJECT_PREFIX}-${CI_COMMIT_REF_SLUG}"
+    - export RELEASE_NAME="mender-$(echo -n "${REVIEW_APPS_PROJECT_PREFIX}-${CI_COMMIT_REF_SLUG}" | md5sum | cut -c1-12)"
```
mzedel: Can you put this in a template that's referenced here? Mainly to ensure that this and the cleanup won't diverge.

.gitlab-ci.yml
oldgiova: That variable is not already available at that stage, but then the only way to show the correct `environment.url` is to use a dedicated `REVIEW_APPS_PROJECT_PREFIX` in Gitlab CICD variables, instead of computing it. Let's go this way. I avoided it in the first place because I don't like CICD variable proliferation, but it seems unavoidable

mzedel: If it is only needed to point to the correct location within gitlab then I'm fine with having the ultimate truth in the job log 👌 - I thought this might be used to track down the deployment. Then please ignore my comment.

oldgiova: Well, that's fine, as it would easily solve also the templating for the other comment. It's just two more vars

.gitlab-ci.yml
oldgiova: Moved to Gitlab CICD vars, this is not necessary anymore

mzedel (approved):

PR #1601, opened 2026-03-31
ci: remove ineffectual pipeline variable assignments

alfrunes:
`MENDER_IMAGE_GUI` is already expanded so redefining the tag and repository variables does not change it's value.


alfrunes:
Even if it's not ineffectual assignment (and modifying the global variable) wouldn't this behavior be undesirable regardless? :thinking: 
That is, pipeline consuming artifacts built and published by other pipelines.

PR #1603, opened 2026-03-31
ci: let e2e tests rely on minor versions of ent images on release-tags + -branches

mzedel:
- since these will be tagged on saas releases already, we can rely on their presence
- also before/ during the tag or release pipeline

### DIFF
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -222,7 +222,7 @@ test:frontend:acceptance:enterprise:
     - unset MENDER_IMAGE_REGISTRY MENDER_IMAGE_REPOSITORY
     - |
       if echo "$CI_COMMIT_REF_NAME" | grep -qE '^v?[0-9]+\.[0-9]+\.[0-9x]+$'; then
-        export MENDER_IMAGE_TAG="v$(echo "$CI_COMMIT_REF_NAME" | sed 's/\.x$//')"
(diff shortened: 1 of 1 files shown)


frontend/pipeline.yml
```diff
@@ -222,7 +222,7 @@ test:frontend:acceptance:enterprise:
     - unset MENDER_IMAGE_REGISTRY MENDER_IMAGE_REPOSITORY
     - |
       if echo "$CI_COMMIT_REF_NAME" | grep -qE '^v?[0-9]+\.[0-9]+\.[0-9x]+$'; then
-        export MENDER_IMAGE_TAG="v$(echo "$CI_COMMIT_REF_NAME" | sed 's/\.x$//')"
+        export MENDER_IMAGE_TAG="v$(echo "$CI_COMMIT_REF_NAME" | sed 's/^v//; s/\.[^.]*$//')"
```
merlin-northern, line 225: this is what we want, just double checking
```bash
#echo v4.1.2 | sed 's/^v//; s/\.[^.]*$//'
4.1
```

merlin-northern (approved):
approving not to block, with a small question.

frontend/pipeline.yml
mzedel: yes... in the OS repo I don't see another way of testing the changes with an enterprise backend while allowing us to make changes to the codebase + the related tests in one go... + for the time of a minor releases we should be stable enough to not break things in such a way that we can't rely on such an image 🤞

kjaskiewiczz (approved):

PR #1607, opened 2026-04-01
test: accounted for slower time in test runner, where 60s might be just 45s

mzedel:
for https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/13737079770 where despite the 60s + 5s wait for the log in page, the countdown on the dialog only came to 14s left... 😕


alfrunes (approved):
Ahh! I forgot to account for the Playwright theory of relativity 💡

PR #1649, opened 2026-04-07
ci: add qa-common retry and runner defaults

p-targowicz:
Include qa-common/runner.yml and qa-common/retry.yml from mendertesting to apply a consistent .qa-common-default-runner tag and .qa-common-default-retry policy to all jobs pipeline-wide.

Add network probes (apt/go/git-clone) to jobs that perform network operations, so transient failures trigger a retry instead of a hard fail.

Ticket: QA-1490

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,6 +93,10 @@ variables:
     value: "302.Release-information/03.Open-source-licenses/01.Mender-Server/docs.md"
 
 include:
+  - project: Northern.tech/Mender/mendertesting
+    file:
+      - qa-common/runner.yml
+      - qa-common/retry.yml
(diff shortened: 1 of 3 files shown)


oldgiova (dismissed):

mzedel (commented):
looks good by itself, but https://github.com/mendersoftware/mender-server/commit/438025e08e81f4c92d688f37c500376c805386bf (& predates the commit picked in here) set some jobs explicitly to the bigger k8s runner (at least for the frontend unit tests it's highly unlikely to work reliably with the smaller one 😬)

p-targowicz:
@mzedel ok good point and thank you for pointing out issue - now it seems to be fixed !

mzedel (approved):

PR #1654, opened 2026-04-08
fix(gui): refresh the configuration device details on finished deployment

merlin-northern:
fix(gui): refresh the configuration device details on finished deployment

Introduces:

* a maximum number of iterations for which deployments are queried
* a modified logic for the interval in the `DeviceConfiguration` component

Changelog: Title
Ticket: MEN-6415

### DIFF
--- a/frontend/src/js/components/devices/device-details/Configuration.tsx
+++ b/frontend/src/js/components/devices/device-details/Configuration.tsx
@@ -57,6 +57,8 @@ const textStyle = { textTransform: 'capitalize', textAlign: 'left' };
 
 const defaultReportTimeStamp = '0001-01-01T00:00:00Z';
 
+const maxDeploymentsChecks = 128;
+
 const configHelpTipsMap = {
   'mender-demo-raspberrypi-led': {
     position: 'right',
@@ -160,8 +162,10 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
   const [showLog, setShowLog] = useState(false);
   const [updateFailed, setUpdateFailed] = useState();
   const [updateLog, setUpdateLog] = useState();
+  const [isRefreshing, setIsRefreshing] = useState(false);
   const dispatch = useDispatch();
-  const deploymentTimer = useRef();
+  const deploymentTimer = useRef(null);
+  const deploymentCheckCount = useRef(0);
 
   useEffect(() => {
     if (!isEmpty(config) && !isEmpty(changedConfig) && !isEditingConfig) {
@@ -174,26 +178,39 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
     if (deployment.devices && deployment.devices[device.id]?.log) {
       setUpdateLog(deployment.devices[device.id].log);
     }
-
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [JSON.stringify(deployment.devices), device.id]);
 
   useEffect(() => {
-    clearInterval(deploymentTimer.current);
-    if (isRelevantDeployment && deployment.status !== DEPLOYMENT_STATES.finished) {
-      deploymentTimer.current = setInterval(() => dispatch(getSingleDeployment(deployment_id)), TIMEOUTS.refreshDefault);
-    } else if (deployment_id && !isRelevantDeployment) {
-      dispatch(getSingleDeployment(deployment_id));
-    }
-    return () => {
-      clearInterval(deploymentTimer.current);
-    };
-  }, [deployment.status, deployment_id, dispatch, isRelevantDeployment]);
-
-  useEffect(() => {
-    if (!isRelevantDeployment) {
+    if (!deployment_id || isRefreshing) {
       return;
     }
+    deploymentCheckCount.current = 0;
+    deploymentTimer.current = setInterval(() => {
+      if (deploymentCheckCount.current > maxDeploymentsChecks) {
+        clearInterval(deploymentTimer.current);
+        setIsRefreshing(false);
+        return;
+      }
+      setIsRefreshing(true);
+      deploymentCheckCount.current++;
+      dispatch(getSingleDeployment(deployment_id));
+    }, TIMEOUTS.refreshDefault);
+    dispatch(getSingleDeployment(deployment_id));
+    return () => {
+      clearInterval(deploymentTimer.current);
+      setIsRefreshing(false);
+    };
+  }, [deployment_id, isRefreshing]);
+
+  useEffect(() => {
+    if (deployment.status === DEPLOYMENT_STATES.finished) {
+      clearInterval(deploymentTimer.current);
+      setIsRefreshing(false);
+    }
+  }, [deployment.status]);
+
+  useEffect(() => {
     if (deployment.status === DEPLOYMENT_STATES.finished) {
       // we have to rely on the device stats here as the state change might not have propagated to the deployment status
       // leaving all stats at 0 and giving a false impression of deployment success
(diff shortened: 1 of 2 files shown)


merlin-northern:
@mzedel I am sure this is not very proper way of fixing it. but this is what the problem is: currently in main the `getDeviceConfigDeployment` will get stuck on returning the deployment which is `status:"inprogress"` and never recovers, because the `isRelevantDeployment` is false pretty much all the time. please give me some pointers and I will repay you with e2e test for this 😘

mzedel (commented):
IIRC this was even added to not be aggressive towards the backend in updating the deployment status 😅 - but all fine with me... let me know if I shall give you a hand with the failing unit test 👌

merlin-northern:
> IIRC this was even added to not be aggressive towards the backend in updating the deployment status 😅 - but all fine with me... let me know if I shall give you a hand with the failing unit test 👌

thank you for the offer. so it has become personal between me and the units. what kind of units are those, there is half of the browser running, including the storage (with mocks). and believe me for a person who never ever wrote a single line of units in js it is an experience on the side of metaphysical. this is the best I could do before going loose ;-)

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -179,7 +179,11 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
 
   useEffect(() => {
     clearInterval(deploymentTimer.current);
-    if (isRelevantDeployment && deployment.status !== DEPLOYMENT_STATES.finished) {
+    if (deployment_id && !deployment.status) {
+      dispatch(getSingleDeployment(deployment_id));
```
mzedel, line 183: FYI: this will only get the deployment once, so any deployment that doesn't have a status yet will only have the response of this call to provide one...

The list at the end of the `useEffect` are the dependencies of the call, what's in here will then only run if `[deployment.status, deployment_id, dispatch, isRelevantDeployment]` changes. `dispatch` and `deployment_id` are stable once they are set, and unless `deployment.status` is different in the response to the `getSingleDeployment` here, the `isRelevantDeployment` will likely also remain stable 😕.

merlin-northern: right. okey, so take two. this is the best I could do, and I wanted to keep changes to the minimum. I am sorry I am wasting your time, and thank you for being patient. a couple of years more with this PR and I will learn.

merlin-northern:
unreported pipeline here: https://gitlab.com/Northern.tech/Mender/mender-server/-/pipelines/2480335843

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -173,26 +175,32 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
...
+    }, TIMEOUTS.refreshDefault); // TIMEOUTS.refreshDefault/20 makes the unit pass
+  }, [deployment_id]);
 
   useEffect(() => {
-    if (!isRelevantDeployment) {
-      return;
+    if (deployment.status === DEPLOYMENT_STATES.finished) {
+      clearInterval(deploymentTimer.current);
+      deploymentTimer.current = null;
```
mzedel:
```suggestion
      clearInterval(deploymentTimer.current);
```
as that's what `clearInterval` is for

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -173,26 +175,32 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
...
-  }, [deployment.status, deployment_id, dispatch, isRelevantDeployment]);
+    deploymentCheckCount.current = 0;
+    deploymentTimer.current = setInterval(() => {
+      if (deploymentCheckCount.current > maxDeploymentsChecks) {
+        clearInterval(deploymentTimer.current);
+        return;
+      }
+      deploymentCheckCount.current++;
+      dispatch(getSingleDeployment(deployment_id));
+    }, TIMEOUTS.refreshDefault); // TIMEOUTS.refreshDefault/20 makes the unit pass
```
mzedel:
```suggestion
    }, TIMEOUTS.refreshDefault);
    dispatch(getSingleDeployment(deployment_id));
```
the change otherwise means the first deployment check will happen 30s after the component is shown - independent of the deployment state or age (+ the comment would rather be a comment in the PR than in the code...)

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -160,7 +160,9 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
   const [updateFailed, setUpdateFailed] = useState();
   const [updateLog, setUpdateLog] = useState();
   const dispatch = useDispatch();
-  const deploymentTimer = useRef();
+  const deploymentTimer = useRef(null);
+  const deploymentCheckCount = useRef(0);
+  const maxDeploymentsChecks = 128;
```
mzedel: this isn't local to the component IOW should be defined outside of the component

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -173,26 +175,32 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
...
+    deploymentCheckCount.current = 0;
+    deploymentTimer.current = setInterval(() => {
+      if (deploymentCheckCount.current > maxDeploymentsChecks) {
+        clearInterval(deploymentTimer.current);
+        return;
+      }
+      deploymentCheckCount.current++;
+      dispatch(getSingleDeployment(deployment_id));
+    }, TIMEOUTS.refreshDefault); // TIMEOUTS.refreshDefault/20 makes the unit pass
+  }, [deployment_id]);
```
mzedel:
```suggestion
    return () => {
      clearInterval(deploymentTimer.current);
    };
  }, [deployment_id]);
```
otherwise the `interval` will continue polling the backend even if the component is no longer visible

merlin-northern: fixed.

frontend/src/js/components/devices/device-details/Configuration.tsx
merlin-northern: moved out.

frontend/src/js/components/devices/device-details/Configuration.tsx
merlin-northern: fixed.

merlin-northern: sorry for leftover comment

frontend/src/js/components/devices/device-details/Configuration.tsx
merlin-northern:
```
sugar pretty please
could we keep this?
```

merlin-northern: okey so it was useEffect clean up being called very unexpectedly, I had to introduce another state.

frontend/src/js/components/devices/device-details/Configuration.tsx
```diff
@@ -174,26 +178,39 @@ export const DeviceConfiguration = ({ defaultConfig = {}, device: { id: deviceId
...
+      deploymentCheckCount.current++;
+      dispatch(getSingleDeployment(deployment_id));
+    }, TIMEOUTS.refreshDefault);
+    dispatch(getSingleDeployment(deployment_id));
     return () => {
       clearInterval(deploymentTimer.current);
+      setIsRefreshing(false);
     };
-  }, [deployment.status, deployment_id, dispatch, isRelevantDeployment]);
+  }, [deployment_id]);
```
mzedel:
```suggestion
  }, [deployment_id, isRefreshing]);
```
as otherwise you risk basing this on a stale value of `isRefreshing` (https://macwright.com/2024/09/19/the-extra-rules-of-hooks might be a good explanation) - not sure if that would cause issues though...

merlin-northern: it is ok! fixed, and works.

mzedel (approved):
Thanks for bearing with me here 😄

merlin-northern:
> Thanks for bearing with me here 😄

thanks for the valuable comments and suggestions

PR #1656, opened 2026-04-08
4.1.x - ci: ensured commit links in changelog can point to the correct repo

mzedel:
to prevent https://github.com/mendersoftware/mender-docs-changelog/pull/193/changes#diff-2bdfb898251bd081d88ebde7ab4aea4358e28ccd1ce0ad3f0d7c130508130104R17 on branches w/o the changelog generator script...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -788,7 +788,7 @@ release:mender-docs-changelog:
     - wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
   script:
     # Generate the change for this release only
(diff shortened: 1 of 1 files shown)


oldgiova (commented):
Not sure, what about cherry-picking the fixes we made along the way? Like 4973e04c6785f9333acf979e8c2756a507143355

I can follow up with a new PR if you want, as I should have done this at that time

mzedel:
Indeed, having the changelog generator script in 4.1.x would be nice - I wasn't sure what other parts would be connected to this and afraid that would then require a longer string of picks 🫣
I'm not attached to this at all, just noticed the links when reviewing the 4.1.x changelog.

PR #1664, opened 2026-04-09
🍒 4.1.x - ci: fixed qemu client reliability retries configuration

mzedel:
that should be the last missing piece for the retries 🫣


p-targowicz (approved):
LGTM thank you !

PR #1665, opened 2026-04-09
Fix changelog for 4.1.x

oldgiova:


.gitlab-ci.yml
```diff
@@ -785,22 +785,13 @@ release:mender-docs-changelog:
       else
       export CHANGELOG_SUFFIX="";
       fi;
-    - wget --output-document cliff.toml https://raw.githubusercontent.com/mendersoftware/mendertesting/master/utils/cliff.toml
   script:
-    # Generate the change for this release only
-    - git cliff --use-branch-tags --current --output this_mender_docs_changelog.md
```
mzedel, line 791: with this removed isn't the per release changelog no longer generated at all?

oldgiova: That's right. The approach which is currently on `main` as well, is updating the changelog in the repository (CHANGELOG.md, CHANGELOG-enterprise.md, CHANGELOG-saas.md), and it's using it as a base for the mender-docs-changelog

oldgiova:
Wait, it's missing the tag `hetzner-amd-beefy` removal. WIP

EDIT: too many conflicts, I'll do on a dedicated PR

mzedel (approved):

.gitlab-ci.yml
mzedel: Ah of course, I thought that part wasn't part of the 4.1.x as the changelog generation in the removed line was present - thanks for the explanation!

PR #1669, opened 2026-04-09
chore: backport enterprise ng test structure

frodeha:
**Description**
This is a backport of the ng integration test structure established over in enterprise. I'm backporting to ease os->ent sync efforts in the future.

There's some differences in common/ to that are necessary because of differences between os and enterprise. Otherwise these tests should be identical in enterprise (and the goal is to keep it that way).


merlin-northern (approved):
thanks for making the effort!

PR #1676, opened 2026-04-10
MEN-9285: Left Navigation bar design updates

mineralsfree:


frontend/src/js/components/LeftNav.tsx
```diff
@@ -50,19 +50,38 @@ const useStyles = makeStyles()(theme => ({
...
+      right: 0,
+      bottom: 0,
+      width: 1,
+      backgroundColor: theme.palette.divider,
+      zIndex: 1
+    }
   },
   navLink: {
     padding: theme.spacing(3.5),
+    paddingLeft: '43px',
```
mzedel: maybe you could put this in a constant to keep the 2 padding settings aligned, more general I think that should be in line with the general spacing (so `theme.spacing(4.5)`) - but I also see the inconsistency in Figma so 🤷

mzedel (dismissed):

mzedel (approved):

PR #1677, opened 2026-04-13
chore(deviceauth): Remove unused parameters, simplify code, use switch statements

LudvigAnderson:


backend/services/deviceauth/devauth/devauth.go
```diff
@@ -174,7 +174,7 @@ func (d *DevAuth) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
-func (d *DevAuth) setDeviceIdentity(ctx context.Context, dev *model.Device, tenantId string) error {
+func (d *DevAuth) setDeviceIdentity(ctx context.Context, dev *model.Device) error {
```
alfrunes: Hmm.. I believe the `tenantId` should be set in the `UpdateDeviceInventoryReq` as it is in enterprise although it does have any effect in a single tenant environment.

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -1538,8 +1533,7 @@ func (d *DevAuth) updateCheckInTime(
 	// and if it's different trigger reindexing (if enabled)
 	// and save check-in time in the database
 	if previous == nil ||
-		(previous != nil &&
-			!previous.Truncate(24*time.Hour).Equal(checkInTime.Truncate(24*time.Hour))) {
+		!previous.Truncate(24*time.Hour).Equal(checkInTime.Truncate(24*time.Hour)) {
```
alfrunes: Maybe also align this condition with enterprise?
```suggestion
		checkInTime.Sub(*previous) > day || previous.IsZero() {
```

Where
```go
const day = 24 * time.Hour
```
The enterprise version is preferable as the condition here will reset at midnight in whatever locale the expression is evaluated whereas the enterprise version just evaluates the 24h relative duration.

alfrunes (commented):
Appreciate the cleanups. Just found two cases where it might be better to align with enterprise.
Otherwise looks good to me.

alfrunes (approved):

PR #1678, opened 2026-04-13
ci: use round-robin when sharding integration tests

alfrunes:
Makes less reuse of class-scoped fixtures, but should achieve more balanced execution times between nodes.


oldgiova (dismissed):
LGTM; however, probably this would break the OS > ENT sync; the last time I had to implement this fix to the Merge PR: https://github.com/mendersoftware/mender-server-enterprise/pull/1135/changes/f378dafb6040557600a829b29f647d6dbcf9ebed

Perhaps do you know a better way to avoid the sync issue?

alfrunes:
Let's just remove the pytest filters. They are no longer needed since 3d01f34f6c94ce0b7369293c76a62423ea4e4ff0.

oldgiova (approved):

PR #1679, opened 2026-04-13
refactor(deviceauth): Change unlimited limit representation from 0 to -1

LudvigAnderson:
Changelog: Title
Ticket: MEN-9535

### DIFF
--- a/backend/services/deviceauth/devauth/devauth.go
+++ b/backend/services/deviceauth/devauth/devauth.go
@@ -1309,7 +1309,7 @@ func (d *DevAuth) GetLimit(ctx context.Context, name string) (*model.Limit, erro
 		limit, err = d.db.GetLimit(ctx, name)
 		if err != nil {
 			if errors.Is(err, store.ErrLimitNotFound) {
-				limit = &model.Limit{Name: name, Value: 0}
+				limit = &model.Limit{Name: name, Value: model.LimitUnlimited}
 				err = nil
 			} else {
 				return nil, err
@@ -1411,7 +1411,7 @@ func (d *DevAuth) canAcceptDevice(ctx context.Context) (bool, error) {
 		return false, errors.Wrap(err, "can't get current device limit")
 	}
 
-	if limit.Value == 0 {
+	if limit.IsUnlimited() {
 		return true, nil
 	}
 
@@ -1420,7 +1420,7 @@ func (d *DevAuth) canAcceptDevice(ctx context.Context) (bool, error) {
 		return false, errors.Wrap(err, "can't get current device count")
 	}
 
-	if uint64(accepted+1) <= limit.Value {
+	if int64(accepted+1) <= limit.Value {
 		return true, nil
 	}
 
--- a/backend/services/deviceauth/model/limit.go
+++ b/backend/services/deviceauth/model/limit.go
@@ -13,10 +13,15 @@
 //	limitations under the License.
 package model
 
-import "github.com/mendersoftware/mender-server/pkg/mongo/v2/oid"
+import (
+	"github.com/pkg/errors"
+
+	"github.com/mendersoftware/mender-server/pkg/mongo/v2/oid"
+)
 
 const (
-	LimitMaxDevicesCount = "max_devices"
+	LimitMaxDevicesCount       = "max_devices"
+	LimitUnlimited       int64 = -1
 )
 
 var (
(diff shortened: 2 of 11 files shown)


backend/services/deviceauth/model/limit.go
```diff
@@ -26,11 +31,31 @@ var (
...
+		return errors.Errorf("unsupported limit %v", r.Name)
+	}
+
+	isNormalOrUnlimited := r.Value >= 0 || r.IsUnlimited()
+
+	if !isNormalOrUnlimited {
+		return errors.Errorf("unsupported limit value %d", r.Value)
+	}
+
+	return nil
```
frodeha, line 48: This is a matter of taste I suppose, but I would suggest removing the extra line breaks here. They don't really add anything in terms of grouping the code into logical blocks (I would argue the opposite is true), but I'll leave it up to you.

```suggestion

	isNormalOrUnlimited := r.Value >= 0 || r.IsUnlimited()
	if !isNormalOrUnlimited {
		return errors.Errorf("unsupported limit value %d", r.Value)
	}

	return nil
```

backend/services/deviceauth/model/limit.go
```diff
@@ -26,11 +31,31 @@ var (
...
+}
+
+// IsUnlimited checks if the limit has no limit (aka it's unlimited)
+func (l Limit) IsUnlimited() bool {
+	// The special limit value of `-1` means that the limit is unlimited.
```
frodeha: More "it's a matter of taste". I don't see that we need these types of Claude code looking comments. The code is plenty self explanatory as it is.

backend/services/deviceauth/store/mongo/datastore_mongo.go
```diff
@@ -570,6 +571,10 @@ func (db *DataStoreMongo) MigrateTenant(ctx context.Context, database, version s
 			ds:  db,
 			ctx: ctx,
 		},
+		&migration_2_0_2{
```
frodeha, line 574: We have to configure that the [database should be on version](https://github.com/mendersoftware/mender-server/pull/1679/changes#diff-c97666c84e2ee73acad740d3562fadcce9c110ddb02fa98936b5008173495642R45) 2.0.2 for this migration to be run.

backend/services/deviceauth/store/mongo/migration_2_0_2_test.go
```diff
@@ -0,0 +1,64 @@
...
+		ds:  ds,
+		ctx: context.Background(),
+	}
+
+	err = migration.Up(migrate.MakeVersion(2, 0, 1))
+	require.NoError(t, err, "unexpected error when running migration")
+
+	lim, err := ds.GetLimit(ctx1, model.LimitMaxDevicesCount)
+	require.NoError(t, err, "unexpected error when reading limit")
+	assert.Equal(t, int64(-1), lim.Value)
```
frodeha, line 59: You were kind enough to add a constant for unlimited, we should use it also in tests :)

backend/services/deviceauth/devauth/devauth.go
frodeha, line 1301: For sake of completeness I would argue that we should include the cache buster we used in enterprise as well: https://github.com/mendersoftware/mender-server-enterprise/commit/bf02ebed8fe0ffaf0521961b97f525545573b190. 

I don't necessarily think the lack of this is going to result in a lot of trouble for users, but since we're already doing this work we might as well be thorough about it.

backend/services/deviceauth/model/limit.go
LudvigAnderson: Yeah that's reasonable, I'm with you on that.

backend/services/deviceauth/model/limit.go
LudvigAnderson: I copied this straight from enterprise, so it wasn't me :joy: 
But either way, I agree. The comment about "-1" is also obsolete now because I changed it to be a global constant.

backend/services/deviceauth/store/mongo/datastore_mongo.go
LudvigAnderson: Ah, I missed the DbVersion constant. That's what you were talking about, right?

backend/services/deviceauth/store/mongo/migration_2_0_2_test.go
LudvigAnderson: Are you sure, Frode? I feel like this should stay as `-1` because the constant for unlimited is to ensure that production code and things that test production code is using the same constant for unlimited.

However, this test is specifically for migrating from 0 to -1, so in theory, if we for some reason wanted to once again change the representation of unlimited from -1 to something else, this database migration should still be using `-1`.

Please correct me if I'm wrong.

backend/services/deviceauth/store/mongo/datastore_mongo.go
frodeha: Hmm, I see my link didn't really take you to where I wanted it to. But yes, that is indeed what I ment.

backend/services/deviceauth/store/mongo/migration_2_0_2_test.go
frodeha: Fair enough 👍

backend/services/deviceauth/devauth/devauth.go
LudvigAnderson: That's a good point. Is editing the `keyLimit` method the only thing I must do, then?

frodeha: Yep, that changes the Redis key used for looking up cached limit values. Since no key will have a value you effectively force a cache miss, which will result in the limit being fetched from the database and cached with the new key. Old key/values will simply be left to expire on their own when their TTL expires.

frodeha (approved):
LGTM 👍

PR #1683, opened 2026-04-14
MEN-8541: search improvements

mineralsfree:


frontend/src/js/components/header/Header.tsx
```diff
@@ -147,7 +147,25 @@ const useStyles = makeStyles()(theme => ({
...
   },
-  search: { alignSelf: 'center' }
+  headerSide: {
+    flex: 1,
+    '&:last-child': {
+      justifyContent: 'flex-end'
+    }
+  },
+  search: {
+    alignSelf: 'center',
```
mzedel:
```suggestion
    alignSelf: 'center',
    transition: 'all 0.1s ease',
```
to make it a slight bit smoother maybe?

frontend/src/js/components/header/Header.tsx
```diff
@@ -147,7 +147,25 @@ const useStyles = makeStyles()(theme => ({
...
+      flexGrow: 1
+    }
+  },
+  searchTextField: {
+    width: 220,
+    maxWidth: 640,
+    '&:focus-within': {
+      width: '100%'
+    }
+  }
```
mzedel: not sure if you've looked into it, but instead of the 3 separate classes I could imagine a single one defining a grid with columns might describe this too and by that get around the sudden jumps in width/ the off center input? 

Image

I tried it with:
```
  search: {
    justifyContent: 'center',
    display: 'flex',
    transition: 'all 0.1s ease',
    '&:focus-within': {
      width: '100%'
    }
  },
  searchTextField: {
    width: 220,
    maxWidth: 640
  }
```
and in the `header` class:
```
    gridTemplateColumns: 'max-content 1fr max-content',
    '&.service-provider': {
      gridTemplateColumns: '1fr max-content'
    },
```
to adjust this for the 2 variants we have...
and there you could get around the need for the `searchClassName` prop as the search `form` can be selected via `.MuiFormControl-root`

frontend/src/js/common-ui/Search.tsx
```diff
@@ -103,14 +104,20 @@ export const ControlledSearch = ({
       onSearch(debouncedSearchTerm).then(focusAndLock);
     }
   }, [debouncedSearchTerm, onSearch, shouldTriggerSearch]);
-
-  const resetSearchAdornment = searchValue ? (
-    <InputAdornment position="end" className={clearButtonOnHover ? classes.adornment : ''}>
-      <IconButton size="small" onClick={() => resetField(name)}>
+  const resetSearchAdornment = (
+    <InputAdornment position="end" style={!searchValue ? { visibility: 'hidden' } : {}}>
```
mzedel: by moving the visibility check here the content of the search will be obfuscated when not focused:

Image

frontend/src/js/common-ui/Search.tsx
```diff
@@ -30,14 +30,14 @@ const startAdornment = (
...
 const MINIMUM_SEARCH_LENGTH = 2;
 const useStyles = makeStyles()(() => ({
-  adornment: {
-    visibility: 'hidden',
-    opacity: 0,
-    transition: 'opacity 0.3s ease-in-out'
-  },
-  textFieldRoot: {
+  clearOnHover: {
+    '& .MuiInputAdornment-positionEnd': {
```
mzedel:
```suggestion
    [`& .${inputAdornmentClasses.positionEnd}`]: {
```

frontend/src/js/common-ui/Search.tsx
mzedel: also there's the last comment here: https://www.figma.com/design/vewvWAMm9vPomw4YZOHiGt?node-id=6864-7279#1619899903 - which could simplify things

mzedel (commented):
looks good, some suggestions for some details...
regarding the `ClickAwayListener`, there was an explicit request to add this in the past, which caused: https://github.com/mendersoftware/gui/pull/3791 - not sure if that has changed now...

mzedel (dismissed):

mzedel (approved):

PR #1688, opened 2026-04-15
MEN-9302 - device details content section alignment

mzedel:


frontend/src/js/common-ui/ContentSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const useStyles = makeStyles()(theme => ({
+  title: { display: 'grid', gridAutoFlow: 'column', gridAutoColumns: 'max-content', gap: theme.spacing(), alignItems: 'center' }
+}));
+
+export const ContentSection = ({ children, className = '', isAddOn = false, postTitle, title }: ContentSectionProps) => {
+  const { classes } = useStyles();
+
+  return (
+    <div className={`margin-bottom margin-top-medium ${className}`}>
```
mineralsfree, line 35: It might be used everywhere now, but I'd drop the margins for the outer element of the shared component. I think it should be the consumer's responsibility to set up spacings

frontend/src/js/common-ui/ContentSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className="flexbox space-between">
+        <div className={`margin-bottom-small ${classes.title}`}>
+          <Typography variant="subtitle1">{title}</Typography>
+          {postTitle}
+        </div>
+        {isAddOn && <Chip className="" label="Add-on" />}
```
mineralsfree: is `className=""` a leftover?

frontend/src/js/components/devices/device-details/DeviceTwin.tsx
```diff
@@ -207,17 +189,23 @@ export const DeviceTwin = ({ device, integration }) => {
   const widthStyle = { maxWidth: isSync ? maxWidth : 'initial' };
 
   return (
-    <DeviceDataCollapse
+    <ContentSection
       header={
```
mineralsfree: header prop is not there anymore, need to migrate

mineralsfree (changes requested):
The wrong prop is passed in DeviceTwin component that should be fixed, other things are minor 😅

frontend/src/js/common-ui/ContentSection.tsx
mzedel: yeah, the designs use a common margin across the pages and I thought this would be the easiest to gain a consistent appearance across pages... how about setting them as the default value for `className` or maybe rather something like: https://mui.com/material-ui/api/container/#container-prop-disableGutters or https://mui.com/material-ui/api/list/#list-prop-disablePadding ?

frontend/src/js/common-ui/ContentSection.tsx
mzedel: indeed, from https://github.com/mendersoftware/mender-server/blob/main/frontend/src/js/components/devices/device-details/DeviceDataCollapse.tsx#L28 - removing 👌

frontend/src/js/common-ui/ContentSection.tsx
mineralsfree: Maybe it's possible to move margins here: https://github.com/mendersoftware/mender-server/blob/0b2178191922916373137b5fc560156de0719462/frontend/src/js/components/devices/ExpandedDevice.tsx#L303
If we want to share this component, I think it's best to render it without margins by default.
The default className approach might cause confusion if you add a class and your layout suddenly shifts. Mui approach is better, but there it's used to disable padding, which is legitimate internal spacing. 

I’ll leave it up to you, it’s not a big deal either way

mineralsfree (dismissed):

frontend/src/js/common-ui/ContentSection.tsx
mzedel: Indeed, in the linked instance it's used for padding, but there are others where the margin is set by default: e.g. https://github.com/mui/material-ui/blob/master/packages/mui-material/src/FormControl/FormControl.js#L49-L50 (or e.g. AccordionSummary, ListItemText, some FormControls - for these we used to have overrides in the old theme). Alternatively: a related section from the v9 migration: https://mui.com/x/migration/migration-pickers-v8/#disablemargin-prop-removal (which I'd is say not far off from overriding the defaults with a passed class via `className`).
The main point being, the library we rely on considers margins legitimate spacing too - with varying levels of ease in overriding this.

And yes the task is to make use of the component across the UI when working on the redesign tasks - with the goal of getting a consistent appearance of all the content sections in the drawers. I added the change to the release details to clarify...

mineralsfree (approved):

PR #1692, opened 2026-04-17
MEN-9371: Redis cluster `max_retries` handles node IP changes

alfrunes:
The redis URL parameter `max_retries` no longer use different default
than ordinary client options which has default of 3.
The `max_retries` also uses a hook that better handles nodes being
assigned new IP addresses, for example if hosted as Kubernetes
statefulset.

### DIFF
--- a/backend/pkg/redis/redis.go
+++ b/backend/pkg/redis/redis.go
@@ -39,8 +39,7 @@ var (
 	ErrCacheInvalid = errors.New("cache invalidated")
 )
 
-// nolint:lll
-// NewClient creates a new redis client (Cmdable) from the parameters in the
+// ClientFromConnectionString creates a new redis client (Cmdable) from the parameters in the
 // connectionString URL format:
 // Standalone mode:
 // (redis|rediss|unix)://[<user>:<password>@](<host>|<socket path>)[:<port>[/<db_number>]][?option=value]
@@ -64,6 +63,8 @@ var (
 // read_timeout        duration
 // tls                 bool
 // write_timeout       duration
+//
+//nolint:lll
 func ClientFromConnectionString(
 	ctx context.Context,
 	connectionString string,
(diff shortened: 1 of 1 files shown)


backend/pkg/redis/redis.go
```diff
@@ -80,18 +81,21 @@ func ClientFromConnectionString(
 	// in case connection string was provided in form of host:port
 	// add scheme and parse again
 	if redisurl.Host == "" {
-		redisurl, err = url.Parse("redis://" + connectionString)
-		if err != nil {
-			return nil, err
-		}
+		redisurl.Host = net.JoinHostPort(redisurl.Scheme, redisurl.Opaque)
+		redisurl.Scheme = "redis"
+		redisurl.Opaque = ""
```
merlin-northern, line 125: that looks strange first we use Opaque and then we set it to empty, why is that?

backend/pkg/redis/redis.go
```diff
@@ -104,52 +108,57 @@ func ClientFromConnectionString(
 			addrs = append(addrs, fmt.Sprintf("%s:%d", host, srv[i].Port))
 		}
 		redisurl.Host = strings.Join(addrs, ",")
-		// cleanup the scheme with one known to Redis
-		// to avoid: invalid URL scheme: tcp-redis+srv
```
merlin-northern, line 108: okey I am not an expert on redis urls right now, but this is handled, right?

backend/pkg/redis/redis.go
```diff
@@ -104,52 +108,57 @@ func ClientFromConnectionString(
...
-		redisurl.Scheme = "redis"
+	}
+	// Allow host to be a comma-separated list of hosts.
+	addrs := strings.Split(redisurl.Host, ",")
+	if len(addrs) > 1 {
+		redisurl.Host = addrs[0]
+		q := redisurl.Query()
+		for _, addr := range addrs {
+			q.Add("addr", addr)
+		}
```
merlin-northern, line 159: what I would envision is all that dance around the url given in a helper function, like parseRedisUrl (any name you want). that would make it easier to read (and even many test)

merlin-northern (dismissed):
dropped some minor comments, this is not a breaking change, and all the redis urls out there in the wild will work, right?

backend/pkg/redis/redis.go
alfrunes: It's mostly to do with how golang [parses URLs](https://pkg.go.dev/net/url@go1.26.2#URL) without a scheme (like `host:port`). It will interpret whatever comes before the colon as scheme and whatever comes after is put into the Opaque. When stringified, the Opaque takes presedence if set which is why it needs to be cleared once "normalized".

backend/pkg/redis/redis.go
alfrunes: I remembered this was a pretty unusable workaround that I came up with that I just want to get rid of.

backend/pkg/redis/redis.go
alfrunes: Lemme give it another try.

alfrunes: Is this better? 4fab88e60fd0107ca8eecd6ed68dd8e8229526c2

merlin-northern: magical.

merlin-northern (approved):

PR #1693, opened 2026-04-17
ci: use branch-matching image tag for frontend acceptance tests

pasinskim:
The frontend acceptance jobs hardcoded MENDER_IMAGE_TAG=main, causing release branches (e.g. 4.1.x) to pull all backend and GUI images from main instead of the matching release tag.

- pipeline.yml: use CI_COMMIT_BRANCH as the image tag in the template before_script and in the enterprise acceptance jobs (falls back to main for tag pipelines or when unset)
- docker-compose.e2e-tests.yml: GUI image fallback now honours MENDER_IMAGE_TAG instead of hardcoding :main

### DIFF
--- a/frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
+++ b/frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
@@ -67,8 +67,8 @@ services:
   ## use playwright locally using `npm install` in the e2e_tests directory and run `npm run test`
   ##
   gui:
-    # Provide override variable MENDER_IMAGE_GUI, but fall back on default from dev/docker-compose.yml
-    image: ${MENDER_IMAGE_GUI:-docker.io/mendersoftware/gui:main}
+    # Provide override variable MENDER_IMAGE_GUI, but fall back on default using MENDER_IMAGE_TAG
+    image: ${MENDER_IMAGE_GUI:-docker.io/mendersoftware/gui:${MENDER_IMAGE_TAG:-main}}
     environment:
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e_tests/docker-compose.e2e-tests.yml
```diff
@@ -67,8 +67,8 @@ services:
   ## use playwright locally using `npm install` in the e2e_tests directory and run `npm run test`
   ##
   gui:
-    # Provide override variable MENDER_IMAGE_GUI, but fall back on default from dev/docker-compose.yml
-    image: ${MENDER_IMAGE_GUI:-docker.io/mendersoftware/gui:main}
+    # Provide override variable MENDER_IMAGE_GUI, but fall back on default using MENDER_IMAGE_TAG
+    image: ${MENDER_IMAGE_GUI:-docker.io/mendersoftware/gui:${MENDER_IMAGE_TAG:-main}}
```
mzedel, line 71: for the gui `MENDER_IMAGE_GUI` is always defined in CI: `MENDER_IMAGE_GUI: ${MENDER_IMAGE_REGISTRY}/${MENDER_IMAGE_REPOSITORY}/gui:${MENDER_IMAGE_TAG}` - at the top of the frontend pipeline.yml

mzedel (commented):
I think instead we could pick e0add17f213434a02e10aab8df4489220caf2723 to 4.1.x

pasinskim:
I think you are right @mzedel. Let's do that instead.

PR #1694, opened 2026-04-17
Improved docker compose handling in ng tests

frodeha:
**Description**

When specifying the `PROJECT_NAME` environment variable we check if that environment is already running and use it if it is.


merlin-northern (approved):
we could check if the running ProjectName is healthy.

PR #1695, opened 2026-04-17
🍒 4.1.x - ci: let e2e tests rely on pipeline built images in os suite

mzedel:
- this is now possible as we build all service images in all pipelines


pasinskim (approved):
Nice. Thanks for looking into it!

PR #1697, opened 2026-04-17
refactor(deviceauth): don't aggregate usage if device limit is zero

LudvigAnderson:
Ticket: MEN-9471
Changelog: Title


frodeha (dismissed):
Nice, LGTM! 👍 

Although approved, it would be nice if you have the time and opportunity to add a test case that verifies this to ensure we don't regress on it in the future :)

LudvigAnderson:
@frodeha, I added a test. Does it look OK, or is there somewhere else it should be tested?

alfrunes (approved):

frodeha (approved):

frodeha:
> @frodeha, I added a test. Does it look OK, or is there somewhere else it should be tested?

Thanks!

PR #1703, opened 2026-04-20
chore: Update mender-artifact arguments to match latest version

elkoniu:
In the latest version of mender-artifact parameter `--device-type` has been replaced with `--compatible-types`. This requires update all mender-artifact calls which still use old version of argument name.

Ticket: MEN-9344


elkoniu:
@alfrunes @merlin-northern I have checked the repo in search for the `device-type` but rest of the occurences looks to be an internal backend implementation. Please let me know if this need to be adjusted too (I would prefer some backend dev to look into it).

alfrunes (approved):

PR #1707, opened 2026-04-21
MEN-9289: Code component adjustments 

mineralsfree:
needs https://github.com/NorthernTechHQ/nt-gui/pull/673


frontend/src/js/common-ui/CopyCode.tsx
```diff
@@ -11,47 +11,95 @@
...
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+import type { CSSProperties, ReactNode } from 'react';
 import { useState } from 'react';
 import CopyToClipboard from 'react-copy-to-clipboard';
 
-import { FileCopy as CopyPasteIcon } from '@mui/icons-material';
-import { Button, IconButton } from '@mui/material';
+import { ContentCopy as CopyPasteIcon } from '@mui/icons-material';
+import { Button, IconButton, Typography, type TypographyProps } from '@mui/material';
```
mzedel, line 19: what do you think about adding `@typescript-eslint/consistent-type-imports` in here as we have it in `nt-gui`?

mzedel (dismissed):

frontend/src/js/common-ui/CopyCode.tsx
mzedel: oh nice, I didn't expect you to add + fix them right away - but I appreciate it! 😄

mzedel (approved):

PR #1709, opened 2026-04-21
docs(deployments): added description of tag filtering in releases endpoint

frodeha:
**Description**
Added a description of how tag based filtering works to the OAS list releases endpoint. The context of this is that I'm working through the Server API Docs epic and this is the one endpoint I could find where we don't document this behavior (which is an acceptance criteria in that Epic).

Ticket: None

### DIFF
--- a/backend/docs/api/deployments/management_v2.yaml
+++ b/backend/docs/api/deployments/management_v2.yaml
@@ -175,7 +175,9 @@ paths:
         name: name
         schema:
           type: string
-      - description: Tag filter.
+      - description: |-
+          Filter the releases based on their associated tags and only return
+          releases that have at least one matching tag (i.e. OR matching).
         explode: true
         in: query
         name: tag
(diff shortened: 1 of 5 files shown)


backend/docs/api/deployments/management_v2.yaml
```diff
@@ -175,7 +175,9 @@ paths:
         name: name
         schema:
           type: string
-      - description: Tag filter.
+      - description: |-
+          Filter the releases based on their associated tags and only return
+          releases that have at least one matching tag (e.g OR matching).
```
alfrunes: Id est?
```suggestion
          releases that have at least one matching tag (i.e. OR matching).
```

frodeha: Guess I should brush up on my latin! 💅 - thanks, fixed: e6d3b181243462166e7b646c507a1f1d38ce0e83

alfrunes (approved):

PR #1724, opened 2026-04-22
Upgrade/mui

mineralsfree:


frontend/vitest.config.ts
```diff
@@ -1,18 +1,51 @@
...
+const muiIconsDeepImports = (): Plugin => {
+  const barrel = /import\s*\{([^}]+)\}\s*from\s*['"]@mui\/icons-material['"]\s*;?/g;
+  return {
+    name: 'mui-icons-deep-imports',
+    enforce: 'pre',
+    transform(code) {
+      if (!code.includes('@mui/icons-material')) return null;
+      barrel.lastIndex = 0;
+      if (!barrel.test(code)) return null;
+      barrel.lastIndex = 0;
```
mzedel: is this one needed?

frontend/src/js/common-ui/forms/Autocomplete.tsx
```diff
@@ -16,16 +16,24 @@ import { Controller, useFormContext } from 'react-hook-form';
...
       name={name}
-      render={({ field: { onChange: formOnChange, ...props } }) => {
-        const onChangeHandler = (e, data) => formOnChange(data);
-        return <Autocomplete {...(freeSolo ? { freeSolo, onInputChange: onChangeHandler } : { onChange: onChangeHandler })} {...props} {...remainder} />;
+      render={({ field: { onChange: formOnChange, ref, value, ...field } }) => {
+        const onChangeHandler = (_e, data) => formOnChange(data);
+        const wrappedRenderInput = params => renderInput({ ...params, inputRef: ref });
+        return (
+          <Autocomplete
+            {...remainder}
```
mzedel: could you put the `remainder` at the end again to ensure outside overrides go through (that is somewhat aligned with the mui codebase, although there it's pretty hidden in the useSlots hook nowadays 🕵️)

frontend/src/js/components/deployments/Report.tsx
```diff
@@ -314,7 +314,12 @@ export const DeploymentReport = ({ abort, onClose, past, retry, type, open }) =>
         </ColumnWidthProvider>
         {Boolean(deviceId.length) && <LogDialog canAi={canAi} deviceId={deviceId} deployment={deployment} onClose={() => setDeviceId('')} />}
       </div>
-      <Divider className={classes.divider} light />
+      <Divider
+        className={classes.divider}
+        sx={{
+          opacity: '0.6'
+        }}
```
mzedel: Could we remove the divider opacity instead? I _think_ the dividers are all the same now 🤔

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
```diff
@@ -78,12 +78,23 @@ const FeedbackSection = ({ deploymentId, deviceId }) => {
   return (
     <div className="flexbox align-items-center">
       {feedbackSubmitted ? (
-        <Typography variant="body2" color="text.secondary">
+        <Typography
+          variant="body2"
+          sx={{
+            color: 'text.secondary'
+          }}
```
mzedel: So the subvariant color definitions have to go through `sx` while regular color can go into the prop?

frontend/src/js/common-ui/forms/TimeframePicker.tsx
```diff
@@ -111,13 +111,12 @@ export const TimeframePicker = ({
             <DatePicker
               disableFuture
               slotProps={{
-                textField: props => ({
+                textField: {
                   size: isNextTheme ? 'small' : 'medium',
```
mzedel, line 115: 📝 - I'll add a cleanup ticket to find this & other remnants of the theme migration...

mzedel (dismissed):
Thank you very much @mineralsfree! I do particularly like the slot prop unification & the often easier prop passing... Just out of curiosity: were you able to see a benefit from the vitest plugin?

frontend/vitest.config.ts
mineralsfree: yes, regexps are stateful, so they store lastIndex after each `.test` execution

frontend/src/js/components/deployments/deployment-report/AiLogAnalysis.tsx
mineralsfree: That's what codemode suggests. ;) But I found it's still possible to use `color="textSecondary"`, updating 🌔

mineralsfree:
@mzedel without the icon import plugin, one test file which renders a component with an icon takes 90 seconds to import on my machine 😓

frontend/vitest.config.ts
mzedel: you could avoid the state handling:

```ts
const muiIconsDeepImports = (): Plugin => ({
  name: 'mui-icons-deep-imports',
  enforce: 'pre',
  transform(code) {
    if (!code.includes('@mui/icons-material')) return null;
    const barrel = /import\s*\{([^}]+)\}\s*from\s*['"]@mui\/icons-material['"]\s*;?/g;
    const transformed = code.replace(barrel, (_, names: string) =>
      names
        .split(',')
        .map((n: string) => n.trim())
        .filter(Boolean)
        .map((n: string) => {
          const [orig, alias] = n.split(/\s+as\s+/).map(s => s.trim());
          return `import ${alias || orig} from '@mui/icons-material/${orig}';`;
        })
        .join('\n')
    );
    return transformed === code ? null : { code: transformed, map: null };
  }
});
```

while also reducing the 2nd/3rd check - the local definition should not impact performance (at least for me locally it even was slightly quicker)

mzedel (dismissed):
The barrel reset just looks so impure 🙈, but it shouldn't stop this from getting in + I didn't expect the imports to have such an impact 😱 - thanks again for looking into it!

mzedel (approved):

PR #1734, opened 2026-04-24
MEN-9639 - fix(deployments): fixed an issue that prevented filtering deployments by group in some cases

mzedel:


frontend/tests/e2e_tests/integration/03-advanced/04-deployments.spec.ts
```diff
@@ -133,6 +133,16 @@ test.describe('Deployments', () => {
...
     await page.getByRole('listitem').first().waitFor({ timeout: timeouts.sixtySeconds });
   });
+
+  test('allows deployment filtering by name', async ({ demoDeviceName, page }) => {
+    await navbar.getByRole('link', { name: /deployments/i }).click();
+    await page.getByRole('tab', { name: /Finished/i }).click();
+    await page.getByRole('combobox', { name: /Target devices/i }).click();
+    await page.getByRole('combobox', { name: /Target devices/i }).fill(demoDeviceName);
+    await page.getByRole('listitem').first().waitFor({ timeout: timeouts.fiveSeconds });
+    expect(await page.getByRole('listitem').all()).toBeTruthy();
```
mineralsfree: This assertion is always true. `.all()` returns an array, and arrays are truthy even when empty

mineralsfree (dismissed):
Maybe @alfrunes should take a look at the backend part. The frontend part looks good to me

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -1724,22 +1724,22 @@ func ParseDeploymentLookupQueryV2(vals url.Values) (model.Query, error) {
 			"id_scope requires id_attribute",
 		)
 	}
-	if idScope != "" && idScope != "identity" && idScope != "inventory" {
+	if idScope != "" && idScope != "identity" && idScope != "tags" {
```
alfrunes, line 1727: The device `group` is in the `system` scope. Should we simply allow any valid scopes here? Not sure what we allow for the user settings.

mzedel: The user settings are entirely free form so hard to restrict and the UI should try to work with everything. But in the UI configuration it will only offer `identity` attributes + the device name (so only `name` attribute with `tags` scope - this I falsely remembered as being an `inventory` attribute 🤦).
My idea was to have the friction of only allowing the "properly" supported device identifier attributes to prevent poor requests from causing extra load on the deployments search endpoint. But if it's not an issue I'd happily remove the limitations entirely 🤷

alfrunes: Ahh, ok. I was just a little confused from the commit message / PR title. Then I think it's fine to just change the scope for now.

mineralsfree (approved):

PR #1736, opened 2026-04-27
ci: let e2e test jobs rely on docker install from testing image

mzedel:
relying on https://github.com/mendersoftware/mender-test-containers/pull/696


oldgiova (approved):
👍

PR #1743, opened 2026-04-27
MEN-9138: Number input implementation

mineralsfree:


frontend/src/js/common-ui/forms/NumberField.tsx
```diff
@@ -0,0 +1,120 @@
...
+export const NumberField = (props: NumberFieldProps) => {
+  const { id: idProp, label, error, size = 'medium', showSteps = false, helperText, onBlur, inputStyle, endAdornment = null, ...other } = props;
+  let id = useId();
+  if (idProp) {
+    id = idProp;
+  }
+  return (
+    <BaseNumberField.Root
+      {...other}
+      allowOutOfRange
```
mzedel: similar to https://github.com/mendersoftware/mender-server/pull/1724#discussion_r3137827509 and in particular in combination with the `allowOutOfRange` - I'm not sure what our latest stance on the free input values is, but I'm afraid it won't be always consistently applied 😬

frontend/src/js/common-ui/forms/NumberField.tsx
```diff
@@ -0,0 +1,120 @@
...
+import InputAdornment from '@mui/material/InputAdornment';
+import InputLabel from '@mui/material/InputLabel';
+import OutlinedInput from '@mui/material/OutlinedInput';
+
+import { NumberField as BaseNumberField } from '@base-ui/react/number-field';
+
+type NumberFieldProps = BaseNumberField.Root.Props & {
+  endAdornment?: ReactNode;
+  error?: boolean;
+  helperText?: string;
```
mzedel:
```suggestion
  helperText?: ReactNode | string;
```

and then allow it in the `NumberInput`/ remove the `helperText as string`?

frontend/src/js/common-ui/forms/NumberField.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { useId } from 'react';
+
+import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import KeyboardArrowUpIcon from '@mui/icons-material/KeyboardArrowUp';
+import FormControl from '@mui/material/FormControl';
+import FormHelperText from '@mui/material/FormHelperText';
+import IconButton from '@mui/material/IconButton';
+import InputAdornment from '@mui/material/InputAdornment';
+import InputLabel from '@mui/material/InputLabel';
+import OutlinedInput from '@mui/material/OutlinedInput';
```
mzedel: these seem to be the leftover from the (IMO nicely handled) import from the MUI examples, OTOH maybe this files is particularly quick to be processed in the unit tests 😉

frontend/src/js/components/deployments/deployment-wizard/RolloutOptions.tsx
```diff
@@ -112,25 +115,18 @@ export const Retries = ({
...
-  };
-
-  const onInputChange = (e, value, reason) => {
-    if (reason === 'clear') {
-      return setDeploymentSettings({ retries: 0 });
-    } else if ((reason === 'reset' && !e) || reason === 'blur') {
-      return;
+  const onAttemptsChange = (value: number | null) => {
+    if (value == null) {
+      return setCurrentAttempts(null);
```
mzedel: This will set the deployment retries to -1 when cleared, the previous path would set this directly on the deployment settings/ prevent it from being entirely free of a value via `formatValue` combined with the `autoSelect` 🙃
Having this centrally handled in the `useEffect` is better IMO, so maybe we could be strictly stay non-null.

frontend/src/js/components/deployments/deployment-wizard/RolloutOptions.tsx
```diff
@@ -146,29 +142,14 @@ export const Retries = ({
...
-                className={classes.retryInput}
-                slotProps={{
-                  ...params.slotProps,
-                  htmlInput: { ...params.slotProps.htmlInput, value: formatValue(params.slotProps.htmlInput.value) }
-                }}
-                type="number"
-              />
-            )}
-            value={currentAttempts}
+            value={currentAttempts ?? null}
```
mzedel: this won't be `undefined`

frontend/src/js/components/deployments/deployment-wizard/RolloutOptions.tsx
```diff
@@ -146,29 +142,14 @@ export const Retries = ({
...
+            onValueChange={onAttemptsChange}
+            min={0}
```
mzedel: "Internally speaking" that's right, but the UI talks about attempts and there at least 1 should be taken

frontend/src/js/components/settings/Global.tsx
```diff
@@ -239,24 +234,18 @@ export const GlobalSettings = () => {
...
-            {!!intervalErrorText && <FormHelperText className="warning">{intervalErrorText}</FormHelperText>}
-            <FormHelperText>Choose how long a device can go without reporting to the server before it is considered “offline”.</FormHelperText>
-          </FormControl>
+          <Typography className="margin-bottom-small" variant="body2">
+            Choose how long (days) a device can go without reporting to the server before it is considered &quot;offline&quot;
+          </Typography>
+          <NumberField
+            min={1}
+            max={maxOfflineIntervalDays}
+            inputStyle={{ width: 120 }}
```
mzedel: considering that this is the only place to have the explicit `inputStyle`, I'd say an override via a `className` targeting the `OutlinedInput` would be more flexible

frontend/src/js/components/settings/Global.tsx
```diff
@@ -167,9 +162,9 @@ export const GlobalSettings = () => {
     [JSON.stringify(channelSettings), dispatch]
   );
 
-  const onChangeOfflineInterval = ({ target: { validity, value } }) => {
-    if (validity.valid) {
-      setCurrentInterval(value || 1);
+  const onChangeOfflineInterval = (value: number | null) => {
+    if (value != null && value > 0 && value <= maxOfflineIntervalDays) {
```
mzedel: this correctly sets the value, but once you have a valid number set (e.g. 1000) and enter an additional number to disturb the peace, the value will reset, but the error message won't:
Image

frontend/src/js/components/subscription/SubscriptionPage.tsx
```diff
@@ -362,17 +362,16 @@ const SubscriptionForm = ({
...
+    const value = numValue ?? 0;
     const { div: step } = deviceTypes[tier].limitConstrains[selectedPlan.id];
     const snappedValue = Math.ceil(value / step) * step;
     const effectiveLimit = Math.min(
       Math.max(snappedValue, selectedPlan.tierLimitsConstrains[tier].min, currentDeviceLimits[tier]),
       selectedPlan.tierLimitsConstrains[tier].max
     );
     if (value !== effectiveLimit) {
-      setValue(tier, effectiveLimit);
+      setValue(tier, effectiveLimit, { shouldValidate: true });
```
mzedel: would this not require the `rules` to be set for the input like in the `TenantCreateForm`?

mzedel (changes requested):
Very nice, among some nitpicks there's an issue with the deployment retries as clearing the input will now result in an error on deployment creation + the possibility for the unexpected error state in the offline threshold...

mzedel (approved):

PR #1744, opened 2026-04-27
QA-1588: Add `x-mender-plan` and `x-mender-addons` tags to api spec

bahaa-ghazal:


backend/docs/api/deviceconfig/devices_v1.yaml
```diff
@@ -34,6 +34,8 @@ paths:
           $ref: '../common/responses.yaml#/components/responses/NotFound'
         "500":
           $ref: '../common/responses.yaml#/components/responses/InternalServerError'
+      x-mender-addons:
```
frodeha: I'm sorry for this, but I see that the issue description ~is~ was inconsistent with the [api-docs implementation](https://github.com/mendersoftware/mender-api-docs/blob/master/scripts/generate-organized-docs.mjs#L95). This property is called `x-mender-addon` (singular instead of plural), mostly to remain consistent with the `x-mender-plan` property.

frodeha (dismissed):
Changes looks good to me, we just need to rename `x-mender-addons` to `x-mender-addon` everywhere 🙃

frodeha (approved):
LGTM

PR #1746, opened 2026-04-27
fix(deployments) Stop returning more pages than the query returns

nickanderson:
When a client requests a page beyond the available pages, the endpoint would
incorrectly fall back to a different query and return incorrect data.

This fix adds pagination validation to return a 404 (Not Found) response when
the requested page exceeds the available pages, with error message:
"No more pages available".

Ticket: ME-651
Changelog: Return 404 when requested page exceeds available pages


mzedel:
@nickanderson I think this can be closed with https://github.com/mendersoftware/mender-server/pull/1747 + https://github.com/mendersoftware/mender-server/pull/1756 in

nickanderson:
tyhanks

PR #1749, opened 2026-04-28
fix(tests): pytest -k filter parsing in integration test runner

LudvigAnderson:
Changelog: None
Ticket: MEN-9645

### DIFF
--- a/backend/tests/run-integration.bash
+++ b/backend/tests/run-integration.bash
@@ -161,7 +161,11 @@ prepare_pytest_args() {
         if [ "$val" == "-k" ]; then
             filter="next"
         elif [ "$filter" == "next" ]; then
-            PYTEST_FILTER="$PYTEST_FILTER and $val"
+            if [ -z "$PYTEST_FILTER" ]; then
+                PYTEST_FILTER="$val"
+            else
+                PYTEST_FILTER="$PYTEST_FILTER and $val"
+            fi
             filter="done"
         else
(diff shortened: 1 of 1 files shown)


LudvigAnderson:
As mentioned in the Jira ticket, this is just a backporting of https://github.com/mendersoftware/mender-server-enterprise/commit/5bf86cc0

alfrunes:
Thanks for backporting this commit.
Can you please update the commit title to use `ci`, `chore` or `tests` so that this internal thing does not end up in the user facing changelog?
Otherwise looks good to me.

LudvigAnderson:
> Thanks for backporting this commit. Can you please update the commit title to use `ci`, `chore` or `tests` so that this internal thing does not end up in the user facing changelog? Otherwise looks good to me.

Yes, I thought of that as well, but I thought that it would not end up in the changelog because I specified "Changelog: None", or does that not do what I think it does?

alfrunes (approved):

alfrunes:
> Yes, I thought of that as well, but I thought that it would not end up in the changelog because I specified "Changelog: None", or does that not do what I think it does?

You're right, we do have a custom filter in the git-cliff config that supress these commits in the changelog. It does still impact the way the next version is evaluated, but since it's only a patch version bump it doesn't really matter. So in the end I guess it all comes down to personal preference :shrug:
Thanks for updating the commit message regardless :)

PR #1762, opened 2026-05-01
test(gui): select text in number input using keystroke in e2e

mineralsfree:


mzedel (dismissed):
There should also be a `.clear()` or similar...

mzedel (approved):

PR #1787, opened 2026-05-04
test(gui): gave tenant modifying tests more time to fully settle api requests

mzedel:
to maybe help with the failures like [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14193322473)


mineralsfree (approved):
I was looking into this one as well, and as a side note, our `poolOptions` and `disableConsoleIntercept` should be in `test` in `vitest.config.ts`.  That could also cover this issue

PR #1790, opened 2026-05-04
gui unit test config trim

mzedel:
Since the switch to the k8s runners the pool options shouldn't really be relevant any more...


mineralsfree (dismissed):

frontend/vitest.config.ts
mineralsfree: All of this should go as well.  same as cpus import

mineralsfree (approved):

PR #1797, opened 2026-05-05
chore: select right label in signup e2e tests

frodeha:
**Description**

(Hopefully?) fix the e2e failures on main.


mzedel (approved):
Argh, sorry about that - thought I adjusted this too with the recent change 😞

PR #1798, opened 2026-05-05
migrate deployment creation dialog to RHF

mzedel:


frontend/src/js/components/deployments/CreateDeployment.tsx
```diff
@@ -86,6 +88,17 @@ const getAnchor = (element, heightAdjustment = 3) => ({
   left: element.offsetLeft + element.offsetWidth
 });
 
+const defaultValues = {
```
mineralsfree: can it be typed `DeploymentSettings`?

frontend/src/js/components/deployments/deployment-wizard/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+import type { DeploymentFormValues } from './types';
+
+export const deploymentFormSections: Record<keyof DeploymentFormValues, string> = {
+  delta: 'group',
```
mineralsfree: shouldn't it be 'delta'?

frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
```diff
@@ -155,17 +159,9 @@ export const Devices = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [device.id, device.attributes?.mender_is_gateway, dispatch]);
 
-  const deploymentSettingsUpdate = (e, value, reason) => {
-    let update = { group: value };
-    if (reason === 'clear') {
```
mineralsfree, line 160: It seems like resetting the autocomplete is not covered in the new approach.

mineralsfree (changes requested):
I think there is a copy-paste mistake in the form mapping, and autocomplete clearing doesn't reset the state; otherwise, it's a very nice improvement 👍

frontend/src/js/components/deployments/CreateDeployment.tsx
mzedel: I now used `DeploymentFormValues` - but now that I write this, maybe the `DeploymentSettings` are rather useless with the latest changes 🤨

frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
mzedel: there was stale device counts interfering here, so I extracted the related handling to a hook that's then used instead of passing the data - WDYT about that as an alternative?

frontend/src/js/components/deployments/deployment-wizard/utils.ts
```diff
@@ -23,3 +32,64 @@ export const deploymentFormSections: Record<keyof DeploymentFormValues, string>
...
+
+  // Compute device count from group selection
+  useEffect(() => {
+    if (group === ALL_DEVICES) {
+      setDeploymentDeviceCount(acceptedDeviceCount);
+    } else if (groups[group]) {
+      dispatch(getGroupDevices({ group, perPage: 1 }))
+        .unwrap()
+        .then(
+          ({
```
mineralsfree, line 64: Looking at `getGroupDevices` thunk I think we should improve our thunks API a bit 😄

mineralsfree (approved):

frontend/src/js/components/deployments/deployment-wizard/SoftwareDevices.tsx
mineralsfree: Looks reasonable to me

PR #1800, opened 2026-05-05
chore(gui): minimized situations nested paragraphs would trigger a warning by avoiding some nested paragraphs

mzedel:


mineralsfree (dismissed):
Looks like we have some more issue with nested Buttons: 
https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14219527235#L53
if you want to have a look at that as well 🙈

mineralsfree (approved):

PR #1802, opened 2026-05-06
MEN-9680: links and icons fixes

mineralsfree:


mzedel (dismissed):

mzedel:
ah wait: I guess the second `chore(gui): unified links appearance in the DeviceList` is a unification of the deployment icons instead + the linter reported leftover `LaunchIcon` 👌 - but then 👍

frontend/src/js/components/login/Signup.tsx
```diff
@@ -77,7 +77,7 @@ const useStyles = makeStyles()(theme => ({
     alignItems: 'start',
     display: 'grid',
     placeContent: 'center',
-    gridTemplateColumns: inputWidth,
+    gridTemplateColumns: signupWidth,
```
mzedel: this will prevent the input from aligning - I'm aware of the difference to the design, but thought this was just another inconsistency (especially with the default input width being 400 across the codebase)...

mzedel (dismissed):
not sure about the sign up adjustment, as with this it at least for me looks like:
Image
vs
Image

frontend/src/js/components/login/Signup.tsx
mineralsfree: Reverted the 400 width, as for the links, I couldn't reproduce them. 🤔, looks like the screen before I changed it
image

mzedel: yeah I just used staging to check my hunch 😅

mzedel (approved):

PR #1805, opened 2026-05-06
fix(gui): stabilized search callback to prevent infinite render loop

nickanderson:
- captured `onSearch` in a ref via `useLayoutEffect` so the debounced
  search effect and key/focus handlers in `ControlledSearch` no longer
  re-run when the parent passes a new callback identity (e.g. after a
  `trigger` toggle), which under React 19 surfaced as
  "Maximum update depth exceeded"
- removed the `triggerDebounceRef` workaround that previously tried to
  swallow these spurious re-runs
- reverted the parent-side `triggerRef` shim in `Search` now that the
    child owns its own callback stability

Changelog: Fixed an infinite render loop triggered by the search input under React 19
Ticket: ME-656

Co-authored-by: Claude <noreply@anthropic.com>
Co-authored-by: Gemini <noreply@google.com>
Signed-off-by: Nick Anderson <nick@cmdln.org>

### DIFF
--- a/frontend/src/js/common-ui/Search.tsx
+++ b/frontend/src/js/common-ui/Search.tsx
@@ -11,7 +11,7 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
-import { useCallback, useEffect, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


mzedel:
@nickanderson I think this can be closed with https://github.com/mendersoftware/mender-server/pull/1807 deployed & apparently working

nickanderson:
ty

PR #1810, opened 2026-05-07
docs(deployments): Add `GET /software` endpoint for listing softwares

bahaa-ghazal:
Ticket: MEN-9607


backend/docs/api/deployments/management_v1alpha1.yaml
```diff
@@ -0,0 +1,134 @@
...
+  - name: Management API
+paths:
+  /api/management/v1alpha1/deployments/software:
+    get:
+      description: |-
+        Returns software information according to the specified parameters.
+        "Software" in this context refers to base information shared across both Releases and Manifests.
+        Note that the operation between the `name` and `kind` parameters is an AND operation.
+      operationId: Get Deployment Software
+      parameters:
```
mzedel, line 18: Based on the release filtering dialog that would use this endpoint in [the design](https://www.figma.com/design/PzBkamIHb6aIDKiEe0s48u/Filters-for-selecting-a-release-for-deployment?node-id=6198-11666&t=WfOU9FQ9eJQ8hU9b-4) this would also need a filter for an "Artifact type" (what's `update_type` in the `GET /api/management/v2/deployments/deployments/releases` endpoint)

mzedel (changes requested):
The "artifact type" query param is the only thing I see blocking this (and admittedly it's not immediately apparent from the general filter dialog).

Sidenote:
A nice to have would be to get the artifact count as well for releases (then we could rely on this not only for the filter, but also listing releases and manifests in their respective tabs on the software page of the UI). But this would of course increase the complexity in the endpoint where the UI could continue to rely on the separate endpoints instead... + in this case we could expand the response data without breaking API compatibility - so we can add it later if we want to rely more on this endpoint.

backend/docs/api/deployments/management_v1alpha1.yaml
bahaa-ghazal: FYI, I added the `update_type` as a parameter. I also added `kind` as part of the response, where is it useful to know what kind of software you are working with (release or manifest).

frodeha: Thank you Bahaa!

mzedel (approved):
Thanks for the adjustments @bahaa-ghazal, now that should work just fine!

PR #1813, opened 2026-05-08
Test/e2e things

mzedel:
passing staging job is [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14308440304)

### DIFF
--- a/frontend/tests/e2e_tests/run.js
+++ b/frontend/tests/e2e_tests/run.js
@@ -12,7 +12,7 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 import chalk from 'chalk';
-import { mkdirSync, writeFileSync } from 'fs';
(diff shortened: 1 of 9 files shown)


frontend/tests/e2e_tests/run.js
```diff
@@ -85,7 +85,8 @@ try {
   } else {
     console.log(chalk.red('💥 Test runner failed!'));
   }
-  if (process.env.ALLOWED_TO_FAIL) {
+  if (existsSync('storage/allowed-to-fail')) {
```
mineralsfree, line 88: maybe reuse `allowedToFailPath` variable here?

mineralsfree (approved):
Looks good!

frontend/tests/e2e_tests/run.js
mzedel: that's in a `.ts` file, but the run script is still `.js` to ease running it 😬

PR #1809, opened 2026-05-07
MEN-9434: Added spec for get software tags alpha endpoint

frodeha:
**Description** 

Added (suggested) OAS for endpoint for listing software tags. Please refer to the ticket ([MEN-9434](https://northerntech.atlassian.net/browse/MEN-9434)) for further details.

[MEN-9434]: https://northerntech.atlassian.net/browse/MEN-9434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1817, opened 2026-05-11
MEN-9434: Implement endpoint for getting software tags

frodeha:
**Description**

Added implementation for the alpha versioned get software tags endpoint documented in https://github.com/mendersoftware/mender-server/pull/1809. I also opted for some internal refactoring in 7f323206b002110fdd79ff476870fc35c630534b where I re-used this new implementation where we previously would have used the specific "get release tags" business logic.

This is added in OS and is outside the feature flag of manifests because the GUI will use this endpoint also outside manifest context (e.g in the filter for releases, the new software picket for deployments etc).

Lastly, the codecov isn't very happy, which is because it ignores coverage generated by integration tests (which is the testing I've added here). This is being looked into separately.

Ticket: MEN-9434


bahaa-ghazal (approved):
LGTM!
Only test case that is missing from the integration tests is trying to pass multiple kinds at once.

frodeha:
> LGTM! Only test case that is missing from the integration tests is trying to pass multiple kinds at once.

Yeah, one drawback of statically types generated clients tied to the schema is that it doesn't allow you to send incorrect information like that (e.g, only a string kind is supported, not a slice). I can work around it though, but prefer to do so in a separate PR.

PR #1819, opened 2026-05-11
QA-1612: unit tests speed-up

mineralsfree:


frontend/tests/setupTests.ts
```diff
@@ -11,11 +11,52 @@
...
+vi.mock('@northern.tech/store/thunks', { spy: true });
+
+// Avoid inserting links in DOM
+Tracking.cookieconsent = () => Promise.resolve({ trackingConsentGiven: false });
+
+// disable ripple to reduce act warnings
+ButtonBase.defaultProps = {
+  ...ButtonBase.defaultProps,
+  disableRipple: true
+};
```
mzedel: looking at the job output there seem to be quite a few more messages about the act usage now (e.g. https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14303956324) - are you sure this is doing what the comment says?

frontend/tests/setupTests.ts
```diff
@@ -11,11 +11,52 @@
...
+Object.defineProperty(globalThis, 'IS_REACT_ACT_ENVIRONMENT', {
+  configurable: true,
+  get: () => true,
+  set: () => {}
+});
+
+// Unified id's assigned by react to stabilize snapshots
+expect.addSnapshotSerializer({
+  test: val => typeof val === 'string' && /_r_[a-z0-9]+_/.test(val),
+  serialize: (val, config, indent, depth, refs, printer) => printer(val.replace(/_r_[a-z0-9]+_/g, '_r_X_'), config, indent, depth, refs)
```
mzedel, line 43: hopefully they stick to the current format for a while now, as we've gone through iterations here already 😅

mzedel (commented):
quite a nice speedup while retaining the request validation 👍 - can you however look again into the `act` related messages as these now lead to a log too big for the simple Gitlab job log view...

frontend/tests/setupTests.ts
mineralsfree: That's due to 
```
The current testing environment is not configured to support act(...)
```
is now
```
An update to ForwardRef(TouchRipple) inside a test was not wrapped in act(...).
When testing, code that causes React state updates should be wrapped into act(...):
act(() => {
  /* fire events that update state */
});
/* assert on the output */
```
due to https://github.com/mendersoftware/mender-server/pull/1819/changes/BASE..104552e574dadcaad55d7ba91953c72b3456a405#diff-5cf3130dccbecbf507a9237b452e0c30bee3ee800988a87d09b62e0f1b03d65bR49


I've created a follow-up ticket to address those: 
https://northerntech.atlassian.net/browse/QA-1630

frontend/tests/setupTests.ts
mineralsfree: Maybe at least as soon as it's MUI v9 it will be stable 🤞

frontend/tests/setupTests.ts
mzedel: Looking at https://github.com/testing-library/react-testing-library/issues/1302 that points to the location the environment is set to false and also seems to not get acted on I could imagine this to be rather tied to the overall setup 😕

Considering that this PR introduces the extra log output I'd rather not have this go in in the first place, as the job logs get less usable - instead I'd have the follow up address the `The current testing environment is not configured to support act(...)` messages.
To me the connection to the speed changes is not clear while the result decreases usability of the test output - but I may miss the connection...

mineralsfree: Fair, I'll remove this part from PR

mzedel (dismissed):

mzedel (dismissed):

mzedel (approved):

PR #1822, opened 2026-05-11
Test/troubleshoot downloads

mzedel:
I've tried tried running into the firefox download issue a few times with this (e.g. in here: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14312648641), but was luckily unsuccessful


frodeha (approved):
Sure, let's give it a whirl! 🤷

PR #1825, opened 2026-05-12
fix(gui): fixed an issue that prevented requiring confirmation on deployment creation

mzedel:


mineralsfree (approved):
It seems like it was broken for a long time without anyone noticing 😄

mzedel:
> It seems like it was broken for a long time without anyone noticing 😄

yep... I'd be interested in the number of user of this 😬

PR #1827, opened 2026-05-12
fix(deployments): Omit empty tags when listing software tags

bahaa-ghazal:


backend/services/deployments/store/mongo/datastore_mongo_software.go
```diff
@@ -34,7 +34,9 @@ func (db *DataStoreMongo) ListSoftwareTags(
 	}
 	ret := make([]model.Tag, 0, len(tagKeys))
 	for _, elem := range tagKeys {
-		ret = append(ret, model.Tag(elem))
+		if elem != "" {
+			ret = append(ret, model.Tag(elem))
+		}
```
frodeha, line 39: How are we getting empty tags...? That shouldn't happen surely

frodeha: Ah, I see, if there's a release with no tags the mongodb distinct returns empty string (I guess?). Fun! Could we please add a comment describing what's going on here?

bahaa-ghazal: image


It is because of the release tag index we have ([the docs](https://www.mongodb.com/docs/manual/core/index-unique/#missing-document-field-in-a-unique-single-field-index), not the same type of index but I'm sure it behaves the same)
The `release_tags` index tries to fill missing tags with null, which the query catches.

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
```diff
@@ -138,3 +109,41 @@ func (u *DeploymentsManagementV1Alpha1Suite) TestListSoftwareTags() {
 		assert.Equal(u.T(), http.StatusBadRequest, res.StatusCode)
 	})
 }
+
+func (u *DeploymentsManagementV1Alpha1Suite) GenerateTestArtifact(
+	ctx context.Context, t *testing.T, name string) {
```
frodeha: I'm not opposed to this refactor, but I'm partial to returning errors from support functions and delegate whether the test should terminate or not on an error to the main test function (e.g, don't do require.NoError in this function, but rather in the calling function).

Further more, I would suggest not taking an explicit dependency on the *testing.T and instead take an interface that defines what you need (something like [this](https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/tests/runner/tests/enterprise/deployments_management_v1alpha1_test.go#L733-L737)). This improves re-usability. 

Lastly, I would suggest making the function package private.

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
```diff
@@ -60,39 +60,10 @@ func (u *DeploymentsManagementV1Alpha1Suite) TestListSoftwareTags() {
...
-				if err != nil {
-					require.Equal(http.StatusNotFound, res.StatusCode, "unexpected status code from get release")
-					time.Sleep(200 * time.Millisecond)
-					continue
-				}
-				created = true
-				break
-			}
-			require.True(created, "artifact not created in time")
+			u.GenerateTestArtifact(ctx, t, name)
```
frodeha: Seeing as this test fails now if there is a release without a tag present, wouldn't it be a good idea to add a release without a tag to the test itself to make sure we don't regress on this in the future?

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
bahaa-ghazal: alright lets do some refactoring

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
bahaa-ghazal: I did that but didn't add it to the commit 🙃

frodeha (approved):

PR #1830, opened 2026-05-13
MEN-9607: Implement endpoint for listing software

bahaa-ghazal:


bahaa-ghazal:
I wasn't able to test the tags because it would have affected the get software tags tests...

frodeha:
@bahaa-ghazal - grab [this line](https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/Makefile#L136) from the enterprise Makefile (specifically the -tags=nopkcs11 build tag) to fix the ng-test build issues.

backend/services/deployments/store/mongo/datastore_mongo_software.go
```diff
@@ -41,3 +43,83 @@ func (db *DataStoreMongo) ListSoftwareTags(
...
+		}
+
+		if len(filter.Names) > 0 {
+			fltr[StorageKeyReleaseName] = bson.M{"$in": filter.Names}
+		} else if filter.NamePrefix != "" {
+			fltr[StorageKeyReleaseName] = bson.M{
+				"$regex": bson.Regex{
+					Pattern: "^" + regexp.QuoteMeta(filter.NamePrefix) + ".*",
+					Options: "i",
+				},
```
frodeha, line 69: To quote Alf-Rune's comment on the list manifests PR [here](https://github.com/mendersoftware/mender-server-enterprise/pull/1160#discussion_r3109351126)
> The appended .* does not have any effect, and the case insensitive option eliminates the use of index.

Let's go with the same filter as in list manifests: https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/services/deployments/store/mongo/datastore_mongo_manifest_releases.go#L111-L113

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
```diff
@@ -117,6 +125,239 @@ func (u *DeploymentsManagementV1Alpha1Suite) TestListSoftwareTags() {
...
+
+	u.Run("Success/NoFilter", func() {
+		ctx := u.T().Context()
+		softwares, res, err := u.APIClient.DeploymentsV1alpha1ManagementAPIAPI.
+			GetDeploymentSoftware(common.JWTAuthContext(ctx, u.JWT)).
+			Execute()
+		require.NoError(u.T(), err)
+
+		// the count of softwares is effacted by other tests in same suite.
+		// there shoud be minimum 3 softwares.
```
frodeha, line 184: Another option is to update the other tests to clean up after themselves (e.g delete releases when they're done)

frodeha (commented):
Nice, thank you! Looks pretty much good to me, only some minor comments.

backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go
bahaa-ghazal: I think we should have more bulletproof solution in the future  to decouple tests from each other.

frodeha (approved):

PR #1832, opened 2026-05-14
fix(pkg): masking error when checking for invalidated cache

alfrunes:
When getting a value from the cache using the pkg/redis.GetCache, any other error than redis.Nil is masked by ErrCacheInvalid making it appear that the cache was invalidated on any other type of error.


kjaskiewiczz (approved):
nice!

PR #1836, opened 2026-05-16
ci: add healthcheck verification job for the gui container

p-targowicz:
Adds test:frontend:docker:healthcheck to the frontend pipeline , job runs after build:frontend:docker on every pipeline, starts the built gui image, and waits for the Docker HEALTHCHECK to report healthy fails the pipeline if the container becomes unhealthy or does not reach healthy state within 150s

Ticket: QA-608

### DIFF
--- a/frontend/pipeline.yml
+++ b/frontend/pipeline.yml
@@ -162,6 +162,43 @@ build:frontend:docker:
     paths:
       - frontend/licenses.json
 
+test:frontend:docker:healthcheck:
+  stage: test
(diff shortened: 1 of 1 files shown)


mzedel (commented):
IMO this is a waste of resources and I'd rather go for: https://northerntech.atlassian.net/browse/QA-724 IOW running e2e tests on every PR - as this will bring much more meaningful insights.

p-targowicz:
no longer needed closing ticket too

PR #1838, opened 2026-05-19
fix(deviceconnect): mender commands regression after stream migration

alfrunes:
Due to a duplicate implementation of the mender commands API for the internal APIs, only the public APIs were migrated to use the new connection oriented streams.

### DIFF
--- a/backend/services/deviceconnect/api/http/internal.go
+++ b/backend/services/deviceconnect/api/http/internal.go
@@ -15,17 +15,23 @@
 package http
 
 import (
+	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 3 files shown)


merlin-northern:
@alfrunes there are at least two components for which you have to run the full integration pipeline before merging anything: deviceconnect and mender-gateway. a small reminder ;-)

alfrunes:
To prove that it fixes the issue in the integration test that triggers the internal check update API, I started an integration pipeline targeting the build artifacts from this PR: https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/2537586869
The open source version using the build artifacts from this PR passes, where as the same test for enterprise fails with the upstream image.

kjaskiewiczz (approved):

PR #1840, opened 2026-05-19
CI ECR pull-through-cache 

oldgiova:
Pull containers from Amazon ECR which is a pass-through cache for docker.io

Ticket: QA-1643

Co-authored-with: Claude

Requires:
* [x] https://github.com/mendersoftware/saas/pull/1850
* [x] https://github.com/NorthernTechHQ/nt-iac/pull/527
* [x] https://github.com/mendersoftware/mender-test-containers/pull/710

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,6 +45,16 @@ variables:
   MENDER_IMAGE_TAG_TEST: "test-${CI_COMMIT_SHA}"
   MENDER_IMAGE_TAG_BUILDER: "builder-${CI_COMMIT_SHA}"
 
+  AWS_REGION:
+    value: "eu-central-1"
+    description: "AWS region of the ECR pull-through cache registry"
+  ECR_REGISTRY:
+    value: "017659451055.dkr.ecr.eu-central-1.amazonaws.com"
+    description: "ECR registry endpoint that fronts the Docker Hub pull-through cache"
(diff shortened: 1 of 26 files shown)


frodeha (commented):
What do you think about doing the same for the images in the [docker-compose files](https://github.com/mendersoftware/mender-server/blob/main/docker-compose.yml#L402)? I suppose it's not that important as most of the pulls will be done through the ECR cache, but it's probably better to have everything the same if we can, right?

oldgiova:
> What do you think about doing the same for the images in the [docker-compose files](https://github.com/mendersoftware/mender-server/blob/main/docker-compose.yml#L402)? I suppose it's not that important as most of the pulls will be done through the ECR cache, but it's probably better to have everything the same if we can, right?

Yes, I missed those

oldgiova:
@frodeha I added the mirror to the docker compose files, as noted, and also renamed the variable to `MIRROR_REGISTRY` to make it more agnostic for the future

mzedel (approved):
TIL about the `library` organization 📚 👍

frodeha:
> @frodeha I added the mirror to the docker compose files, as noted, and also renamed the variable to MIRROR_REGISTRY to make it more agnostic for the future

Thank you! I was just playing with it locally and it seems to work fine there as well :)

PR #1843, opened 2026-05-20
MEN-9609: manifest removal

mineralsfree:


frontend/src/js/components/releases/manifests/dialogs/RemoveManifest.tsx
```diff
@@ -0,0 +1,54 @@
...
+            Are you sure you want to remove <i>{manifest?.name}</i>?
+          </>
+        )}
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={onCancel}>Cancel</Button>
+        <div style={{ flexGrow: 1 }} />
+        <Button variant="contained" color="secondary" onClick={onRemove}>
+          Remove {pluralize('manifest', count)}
+        </Button>
```
mzedel: I can't seem to find the design for the removal flow, but is there some direction on the button positioning ("both in the corner" or "space between")? and related: could this maybe use the `ConfirmModal` instead?

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
```diff
@@ -119,7 +124,15 @@ export const ManifestQuickActions = () => {
 
   const onTagManifest = useCallback(() => dispatch(setSnackbar('Tagging Manifests is not yet supported')), [dispatch]);
 
-  const onDeleteManifest = useCallback(() => dispatch(setSnackbar('Deleting Manifests is not yet supported')), [dispatch]);
+  const onDeleteManifest = useCallback(() => setConfirmManifestDeletion(true), []);
```
mzedel: should this + the `onCancelDeletion` not be rather stable as they are both not dependent on any state/ props/ ... + the state setter is a regular react managed one..? IOW I think the `useCallback` isn't needed here

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
```diff
@@ -138,7 +151,22 @@ export const ManifestQuickActions = () => {
       onClick: () => action({ ...actionCallbacks, selection: selectedRows })
     }));
 
-  return <BaseQuickActions actions={actions} ariaLabel="manifest-actions" label={`${selectedRows.length} ${pluralized} selected`} />;
+  return (
+    <>
+      <BaseQuickActions
+        actions={actions}
+        ariaLabel="manifest-actions"
+        label={selectedRows.length ? `${selectedRows.length} ${pluralized} selected` : 'Manifest actions'}
```
mzedel: if you select a single manifest and open the manifest it will still show `1 Manifest selected`, whereas `Manifest actions` would be in line with release + device details... (just as I stumbled across this too 😬 https://github.com/mendersoftware/mender-server/pull/1845/changes/8aea4c6f26ad090e3608b00c40fc286fdb1df55d

mzedel (changes requested):

frontend/src/js/components/releases/manifests/dialogs/RemoveManifest.tsx
mineralsfree: There is no design indeed, so I just followed the way Release removal handles it in `RemoveArtifactDialog` compoinent, but It's better to keep the number of components as small as possible, good catch

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
mineralsfree: ok, I'll just drop my commit

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
mineralsfree: fair point, left over from the previous code

mzedel (approved):
Excellent thank you! I'll adjust the other PRs accordingly 👌

PR #1844, opened 2026-05-21
QA-1511: Replace iot-manager internal clients with OAS generated

alfrunes:


backend/services/iot-manager/app/app_iot_core.go
```diff
@@ -77,7 +78,7 @@ func (a *app) setDeviceStatusIoTCore(ctx context.Context, deviceID string, statu
 
 func (a *app) deployConfiguration(ctx context.Context, deviceID string, dev *iotcore.Device) error {
 	if dev.Certificate != "" && dev.PrivateKey != "" && *dev.Endpoint != "" {
-		err := a.wf.ProvisionExternalDevice(ctx, deviceID, map[string]string{
+		err := a.provisionExternalDevice(ctx, deviceID, "AWS", map[string]string{
```
frodeha, line 81: Just want to check that this is right. In the previous version, the `"provider"` is hard-coded to `"Azure"` and here we're changing it to `"AWS"`. Is that intentional?

frodeha (approved):
LGTM, just one sanity check/question is all I have. Nice 👍

backend/services/iot-manager/app/app_iot_core.go
alfrunes: Yeah, I noticed this was wrong (it said `Azure` even in the context of AWS IoT Core), so I "corrected" that.
However, I see now that the parameter isn't even used by the workflow itself and probably only exists for historical reasons.

frodeha: Cool cool, I assumed as much. Just better safe than sorry and all that you know 👍

PR #1845, opened 2026-05-21
MEN-9730 - editable long text as outlined input

mzedel:


frontend/src/js/common-ui/EditableLongText.tsx
```diff
@@ -14,38 +14,72 @@
...
+  original,
+  placeholder = '-',
+  showConfirmation = true
+}: EditableLongTextProps) => {
+  const [isEditingInternal, setIsEditingInternal] = useState(false);
   const [value, setValue] = useState(original);
   const { classes } = useStyles();
 
+  const isControlled = isEditingProp !== undefined;
+  const isEditing = isControlled ? isEditingProp : isEditingInternal;
```
mineralsfree, line 43: maybe 
```
const isEditing = isEditingProp ?? isEditingInternal;
```

frontend/src/js/common-ui/EditableLongText.tsx
```diff
@@ -14,38 +14,72 @@
...
+}
+
+export const EditableLongText = ({
+  fullWidth,
+  isEditing: isEditingProp,
+  onChange,
+  onEditToggle,
+  original,
+  placeholder = '-',
+  showConfirmation = true
```
mineralsfree: Should we maybe remove this prop altogether? since it's true for all occurrences, or am I missing something?

frontend/tests/e2e_tests/fixtures/terminalContent.png
mineralsfree, line 1: something changed here? 😄

mzedel: yep, I've adjusted the spacing in the `ContentSection` (to afaict get a bit closer to the design + reduce the layout shifts for the (in)visible edit buttons embedded in the title) and that changed the screenshot enough to just exceed the diff threshold 😄

frontend/src/js/common-ui/EditableLongText.tsx
mzedel: indeed! I already forgot that the entire part was no longer needed and not just a rebase required - cleaned it up now 👌

frontend/src/js/common-ui/EditableLongText.tsx
mzedel: I rather want to keep the check for the prop consistent across the component 👌

mineralsfree (approved):
Looks good to me! 👍

PR #1847, opened 2026-05-21
QA-1499: Fix specs and regression after migrating to OAS specs

alfrunes:


frodeha (dismissed):

backend/docs/api/deployments/internal_v1.yaml
```diff
@@ -667,24 +667,21 @@ components:
     ConfigurationDeploymentRequest:
       example:
         name: config-1.1
-        configuration: '{"foo":"bar"}'
+        configuration:
+          foo: bar
```
alfrunes, line 671: Strictly speaking, this endpoint is:
```yaml
configuration:
  oneOf:
  - type: string
  - type: object
      additionalProperties:
        type: string
   - type: anythingElse
     description: Causes the endpoint to fail due to serialization issues.
```
But since the deviceconfig only accept string maps, I think we should use that here as well.

frodeha (approved):

PR #1850, opened 2026-05-22
refactor(inventory): remove devicemonitor client

frodeha:
Devicemonitor is not a supported opensource feature. As part of an ongoing migration effort to use autogenerated internal clients we are replacing handwritten clients wherever possible. Since devicemonitor is not a supported opensource feature, the autogenerated client in opensource does not include the devicemonitor API. Therefore we're simply removing the handwritten one here.

We're leaving the inventory HTTP APIs that used the client previously as to not introduce API breakage, but they will always find 0 alerts, effectively making them into no-ops here in opensource.

Ticket: QA-1502


oldgiova:
@mender-test-bot sync

alfrunes (approved):

PR #1855, opened 2026-05-26
test(gui): added wait for request to settle in upgrade flow

mineralsfree:
This to avoid: 
https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14518115779

### DIFF
--- a/frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
+++ b/frontend/src/js/components/subscription/SubscriptionDrawer.test.tsx
@@ -212,7 +212,12 @@ describe('Subscription Drawer component', () => {
     expect(screen.getByText(/Upgrade your subscription/i)).toBeVisible();
     expect(screen.getByText(/ok@ok.ok/i)).toBeVisible();
     await waitFor(() => expect(screen.queryByRole('button', { name: /confirm subscription/i })).toBeVisible(), { timeout: TIMEOUTS.fiveSeconds });
-    await user.click(screen.getByRole('button', { name: /confirm subscription/i }));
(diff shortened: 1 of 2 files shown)


mzedel (dismissed):
This should improve things 👍
Looking at the [`requestPlanUpgrade`](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/store/src/organizationSlice/thunks.ts#L480) (which still  resolves too early due to the missing `unwrap`) and considering the 3s wait for the device limits retrieval + the [per test settling](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/testing/src/setupTests.tsx#L130-L134) we could give this test a
```
await act(async () => {
  vi.runOnlyPendingTimers();
  vi.runAllTicks();
 });
```
in addition... WDYT?

mineralsfree:
> This should improve things 👍 Looking at the [`requestPlanUpgrade`](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/store/src/organizationSlice/thunks.ts#L480) (which still resolves too early due to the missing `unwrap`) and considering the 3s wait for the device limits retrieval + the [per test settling](https://github.com/NorthernTechHQ/nt-gui/blob/main/packages/testing/src/setupTests.tsx#L130-L134) we could give this test a
> 
> ```
> await act(async () => {
>   vi.runOnlyPendingTimers();
>   vi.runAllTicks();
>  });
> ```
> 
> in addition... WDYT?

Good catch, I think it's better to actually unwrap it; otherwise, the error handling makes no sense. Also added a fix on the thunk side: https://github.com/NorthernTechHQ/nt-gui/pull/743

mzedel (approved):

PR #1856, opened 2026-05-26
ci: retry internal gitlab runner job stages

frodeha:
**Description**

Add retries to internal gitlab runner job stages. The default for each of these are `1` (e.g they're not retries). My hope is that setting these will work around issues we're having where jobs fail before our own stuff even starts. For example, my hope is that setting `GET_SOURCES_ATTEMPTS` will mitigate issues like these: https://gitlab.com/Northern.tech/Mender/mender-server-enterprise/-/jobs/14523408479.

The others I have less of direct proof is happening, but I rather just add retries for them while I'm at it as I don't really see the harm.

Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -114,6 +114,14 @@ variables:
       - "true"
       - "false"
 
+  # Retries for internal Gitlab Runner job stages:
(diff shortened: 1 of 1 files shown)


mzedel (approved):

frodeha:
@mzedel - wow, that's quick reviewing there! Thanks :joy:

PR #1857, opened 2026-05-26
QA-1500: Migrate workflows client to OAS generated and then remove it

alfrunes:
This PR migrates the workflows client to OAS generated client and then removes it (from open source).
Note that the last commit will be reverted before merging to enterprise, so the expected end state is that of the second commit.


frodeha (approved):
LGTM - Nice 👍

PR #1872, opened 2026-06-01
chore: separate test dependencies from application dependencies

alfrunes:
Makes the sbom clearer and allows us to generate mocks (until the next go release bump).


frodeha (approved):
Nice, thanks! 👍

PR #1878, opened 2026-06-01
Men 9558 2

mineralsfree:


frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+          width={72}
+          max={200}
+          showSteps
+          id={type}
+          size="small"
+          defaultValue={update_strategy?.order ?? null}
+          rules={{
+            required: 'Order is required',
+            min: { value: 1, message: 'Values must be from 1 to 200' },
+            max: { value: 200, message: 'Values must be from 1 to 200' }
```
mzedel: Maybe that's changed by now, but https://docs.mender.io/orchestrate-updates/manifest#update-strategy says 100

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+import type { ManifestComponent, Software } from '@northern.tech/types/MenderTypes';
+import { customSort, toggle } from '@northern.tech/utils/helpers';
+
+import { SoftwareArtifactFilter } from '../../deployments/deployment-wizard/ReleaseArtifactFilter';
+
+const { setActiveTab } = storeActions;
+
+const useStyles = makeStyles()(() => ({
+  orderInput: {
+    '& .MuiFormHelperText-root': {
```
mzedel:
```suggestion
    `& .{formHelperTextClasses.root}: {
```

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+      alignSelf: 'flex-end',
+      width: 'max-content',
+      whiteSpace: 'nowrap'
+    }
+  }
+}));
+
+interface ComponentTypesTableProps {
+  componentTypes: Record<string, ManifestComponent>;
+  editable?: boolean;
```
mzedel: here and in the other places:
```suggestion
  isEditable?: boolean;
```
to highlight the boolean nature of the variable

frontend/src/js/components/deployments/deployment-wizard/ReleaseArtifactFilter.tsx
```diff
@@ -96,6 +96,8 @@ export const SoftwareArtifactFilter = (props: SoftwareArtifactFilterProps) => {
 
   const onCloseModal = () => {
     setShowAdvancedFilters(false);
+    setFilterCount(0);
```
mzedel: seeing this I wonder if the `filterCount` couldn't just be a calculated value, without relying on the component state - it will be based on the debounced filtering anyway...

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+      return artifact_path || '-';
+    }
+  },
+  {
+    key: 'order',
+    title: 'Order',
+    sortable: true,
+    sortProp: 'update_strategy.order',
+    cellProps: { align: 'right' },
+    render: ({ type, update_strategy }, { editable, onOrderChange, orderInputClassName }) =>
```
mzedel: not necessarily for this change, but considering that we have the per column `extras` approach in other places - what do you think about:
```suggestion
    render: ({ key, type, update_strategy }, { editable, onOrderChange, classes }) => {
    const { root: orWhateverElse, input: orderInputClassName } = classes[key]
```
to somewhat align with mui and the `slots` & `classes` props - OTOH more universally applicable classes would still not fit in in such a setup 🤔

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <Pagination
+          className="margin-top-none"
+          count={total}
+          rowsPerPage={perPage}
+          onChangePage={onChangePagination}
+          onChangeRowsPerPage={newPerPage => onChangePagination(1, newPerPage)}
+          page={page}
+        />
+      </div>
+      {editable && (
```
mzedel: considering https://github.com/mendersoftware/mender-server/pull/1878/changes#diff-13edcf799d383a79f955b04ae1e2e84af50a9e712b4a71686609d700443f8741R112 - this might no longer be needed

frontend/src/js/components/releases/manifests/ManifestDrawer.tsx
```diff
@@ -158,81 +172,150 @@ export const AddManifestDrawer = ({ onClose, open }: AddManifestDrawerProps) =>
...
+                <>
+                  <Button
+                    size="large"
+                    color="neutral"
+                    variant="outlined"
+                    className={classes.releaseSelect}
+                    endIcon={copyReleaseOpen ? <ExpandLessIcon /> : <ExpandMoreIcon />}
+                    onClick={() => setCopyReleaseOpen(!copyReleaseOpen)}
+                  >
+                    <span className={`'${classes.releaseSelectText}' text-overflow`}>{parsedManifest?.name ?? 'Select a Manifest'}</span>
```
mzedel:
```suggestion
                    <span className={`${classes.releaseSelectText} text-overflow`}>{parsedManifest?.name ?? 'Select a Manifest'}</span>
```

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
```diff
@@ -119,7 +119,12 @@ export const ManifestQuickActions = () => {
     [navigate, selectedManifests]
   );
 
-  const onCopyManifest = useCallback(() => dispatch(setSnackbar('Creating a copy from a Manifest is not yet supported')), [dispatch]);
+  const onCopyManifest = useCallback(() => {
+    const name = !isEmpty(selectedManifest) ? selectedManifest.name : selectedManifests[0]?.name;
+    if (name) {
+      onCopy?.(name);
+    }
+  }, [onCopy, selectedManifest, selectedManifests]);
```
mzedel, line 127: I would imagine this needs a check for applicable actions (either here or in the `ManifestsList`) as selecting multiple manifests would be possible, but only the first would be acted on...

frontend/src/js/components/releases/manifests/ManifestDrawer.tsx
```diff
@@ -158,81 +172,150 @@ export const AddManifestDrawer = ({ onClose, open }: AddManifestDrawerProps) =>
...
+    [dispatch, setValue]
+  );
+  const onSelectCopy = (item: Software) => loadCopyManifest(item.name);
+
+  useEffect(() => {
+    if (open && copyFromManifest) {
+      setMode('copy');
+      loadCopyManifest(copyFromManifest);
+    }
+  }, [open, copyFromManifest, loadCopyManifest]);
```
mzedel: the effect + `loadCopyManifest` would conventionally live alongside the other `useEffect` at the top

frontend/src/js/components/releases/manifests/ManifestDrawer.tsx
```diff
@@ -95,16 +98,22 @@ const validateFile = async (file: File) => {
...
 
 type ManifestFormValues = {
   description: string;
+  name: string;
   tags: string[];
 };
 
-export const AddManifestDrawer = ({ onClose, open }: AddManifestDrawerProps) => {
+type Mode = 'upload' | 'copy';
+
```
mzedel, line 118:
```suggestion
type Mode = 'upload' | 'copy';
type ModeInfo = { key: Mode; title: string };

const modes: Record<Mode, ModeInfo> = {
  upload: { key: 'upload', title: 'Upload'},
  copy: { key: 'copy', title: 'Copy'}
};
```
to then also further down be able to:
```
Object.values(modes).map(({ key, title })=> <FormControlLabel value={key} control={<Radio />} label={title} key={key} />)
```
not sure if that's already enough to automatically get the proper type assigned in the `<RadioGroup onChange`, but likely in the `useState` & avoid the mode strings being spread around

mzedel (changes requested):
Looks very nice! I guess only the `releaseSelectText` & order limits _need_ attention - the rest is 💅

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
mineralsfree: I was confused by the backend limit, which is set to 200, but in the docs it's 100. The backend team is informed and will adjust; thank you!

frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx
mineralsfree: I think the check is already inplace https://github.com/mendersoftware/mender-server/blob/053a70e677fd9cfd61881a1730d135855f0478f5/frontend/src/js/components/releases/manifests/ManifestQuickActions.tsx#L75

Or did I get it wrong?

frontend/src/js/components/releases/manifests/ComponentTypesTable.tsx
mineralsfree: This (`{type, update_strategy }`) is per component, so not the place to pass `key`
I passed the classes instead

mzedel (approved):

PR #1908, opened 2026-06-03
test(e2e): fixed e2e selector after MUI bump

mineralsfree:


mzedel (approved):

mzedel (commented):
the commit could however be better in explaining what's going on - based on the test failure the name seems to now resolve to the label instead of the shown value

PR #1913, opened 2026-06-04
Add compatible_types to list software

frodeha:
**Description**

Adds `compatible_types` to software returned from the list software endpoint. `compatible_types` is the name adopted by mender-artifact to cover both `device_types_compatible` (normal Artifact) and `system_types_compatible` (Manifest) and since Software wraps both of those I kept the same name here (it's all still just `device_types_compatible` in the Artifact though).

### DIFF
--- a/backend/services/deployments/model/software.go
+++ b/backend/services/deployments/model/software.go
@@ -25,17 +25,33 @@ func (s SoftwareTagsFilter) Validate() error {
 
 type Software struct {
 	ReleaseBase `bson:"inline"`
+	Artifacts   []Image `json:"-" bson:"artifacts"`
 }
 
-// custom marshal to include the kind in the response
+// custom marshal to include the kind and compatible_types in the response
 func (s Software) MarshalJSON() ([]byte, error) {
+	var compatible_types []string
+	for _, a := range s.Artifacts {
+		compatible_types = append(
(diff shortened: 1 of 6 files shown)


frodeha:
cc: @mineralsfree for the API spec changes

backend/services/deployments/model/software.go
```diff
@@ -25,17 +25,33 @@ func (s SoftwareTagsFilter) Validate() error {
...
 	ReleaseBase `bson:"inline"`
+	Artifacts   []Image `json:"-" bson:"artifacts"`
 }
 
-// custom marshal to include the kind in the response
+// custom marshal to include the kind and compatible_types in the response
 func (s Software) MarshalJSON() ([]byte, error) {
+	var compatible_types []string
+	for _, a := range s.Artifacts {
+		compatible_types = append(
```
mineralsfree, line 35: Won't it contain duplicates if artifacts have the same DeviceTypesCompatible?

frodeha: They can't contain overlapping DeviceTypesCompatible - the database doesn't allow that (I learnt this just today myself 🤷 )

mineralsfree (approved):

alfrunes (approved):

PR #1918, opened 2026-06-08
chore(e2e): submit autocomplete value explicitly in device filters

mineralsfree:


mzedel (commented):
the commit message restates what you changed, can you please give some context as to why that's needed?

mineralsfree:
> the commit message restates what you changed, can you please give some context as to why that's needed?

We have E2E jobs failing in Firefox. The likely cause is that the open autocomplete prevents the select from opening. To make this more reliable, pressing Enter now closes the autocomplete, instead of relying on the filter click to both close the autocomplete and open the select

mzedel:
> We have E2E jobs failing in Firefox. The likely cause is that the open autocomplete prevents the select from opening. To make this more reliable, pressing Enter now closes the autocomplete, instead of relying on the filter click to both close the autocomplete and open the select

In the commit message (as that otherwise just restates what the code change shows) 😉

mzedel (approved):

PR #1920, opened 2026-06-09
Fix deviceconnect stream close error handling and improve rest errors

alfrunes:
* Closing a NATS stream should not terminate the connection
* Refactored rest.utils error rendering utilities
* Prevent writing duplicate error responses if one is already written
  * write a meaningful log message instead

### DIFF
--- a/backend/pkg/accesslog/middleware_gin.go
+++ b/backend/pkg/accesslog/middleware_gin.go
@@ -75,7 +75,12 @@ func (a AccessLogger) LogFunc(
 		lc.addFields(logCtx)
 	}
 	if r := recover(); r != nil {
-		trace := log.CollectTrace()
+		// Skip 4
+		// = accesslog.LogFunc
+		// + log.CollectTrace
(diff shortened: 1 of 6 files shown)


frodeha (commented):
Maybe I'm wrong, but isn't [this](https://github.com/frodeha/mender-server-enterprise/blob/main/backend/services/deviceconnect/api/http/management_filetransfer.go#L257-L260) the root cause of the double write? I don't necessarily disagree with the changes here I think, but shouldn't we also fix the root cause?

alfrunes:
> Maybe I'm wrong, but isn't [this](https://github.com/frodeha/mender-server-enterprise/blob/main/backend/services/deviceconnect/api/http/management_filetransfer.go#L257-L260) the root cause of the double write? I don't necessarily disagree with the changes here I think, but shouldn't we also fix the root cause?

Nice find! I did not see that. :+1:

frodeha (dismissed):

frodeha (approved):

PR #1921, opened 2026-06-10
Ci/review tests

mzedel:


.gitlab-ci.yml
```diff
@@ -427,6 +427,37 @@ review:stop:
 
     - echo "Review app cleanup completed successfully!"
 
+review:e2e-tests:
+  stage: review
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/mendersoftware/mender-test-containers:gui-e2e-testing
+  needs: []
+  rules:
+    - if: '$RUN_REVIEW_E2E == "true"'
```
oldgiova, line 450: I guess this will be triggered by the mender-test-bot, right?

.gitlab-ci.yml
```diff
@@ -427,6 +427,37 @@ review:stop:
...
+    - npx playwright install chromium
+    # base url will rely on the env variable configured above
+    - npm run script -- --environment=${REVIEW_APPS_TEST_ENVIRONMENT} --project=chromium
+  artifacts:
+    expire_in: 2w
+    paths:
+      - frontend/tests/e2e_tests/test-results
+    when: always
+  tags:
+    - hetzner-amd-beefy-privileged
```
oldgiova: Maybe `k8s` or `k8s-qemu`  runners?

compose/k8s/create-initial-user.enterprise.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
```
oldgiova: I don't fully get it, there's already the `create-initial-user.sh` script, won't that work?

oldgiova (commented):
I think it should work, but few clarifications for my understanding

.gitlab-ci.yml
mzedel: yes - [this PR](https://github.com/mendersoftware/integration-test-runner/pull/443) is supposed to do this 👌

.gitlab-ci.yml
mzedel: 🤦 of course, will adjust!

compose/k8s/create-initial-user.enterprise.sh
mzedel: not for the enterprise setup, but I guess they could get integrated 👌

oldgiova: The script have the same naming, but diverges on enterprise:

```
▶ diff compose/k8s/create-initial-user.sh ../../mender-server/main/compose/k8s/create-initial-user.sh 
3c3
< # Post-install script to create initial tenant and user for Mender Enterprise Review Apps
---
> # Post-install script to create initial  user for Mender Review Apps
49d48
....
```

I think that my initial intention was to have both `compose/k8s/create-initial-user.sh` and `compose/k8s/create-initial-user.enterprise.sh` but during review rounds we moved to a single file. Anyway, I have no strong opinion on that, if maintaining the two with a single diverged files or with two different scoped files.

mineralsfree (dismissed):
e2e-test side looks good to me

compose/k8s/create-initial-user.enterprise.sh
mzedel: right, I forgot - I thought this way would be easier to avoid conflicts, but the differences are really minimal -> adjusted to the single file 👌

.gitlab-ci.yml
mzedel: adjusted 👌

oldgiova (approved):

oldgiova:
LGTM (for the CI parts)

PR #1926, opened 2026-06-11
ci: separated frontend sync from the rest

mzedel:
@merlin-northern @pasinskim  - as an interim solution maybe?


merlin-northern:
I was waiting a bit for the DL; I did not react here, because I am a bit torn here. the separation makes sense from one point of view; perhaps let's discuss during MSTM.

mzedel:
> I was waiting a bit for the DL; I did not react here, because I am a bit torn here. the separation makes sense from one point of view; perhaps let's discuss during MSTM.

It was more than a month and is a repeating pattern + you're not alone in this - so collaboration seems to not be a priority.

PR #1929, opened 2026-06-11
MEN-9851 - fix(gui): ensured manifest filter inputs are only usable for users on the required plan

mzedel:
Screenshot 2026-06-11 at 15 29 09

note the selected text - an editable input would get focused when clicked on the placeholder text

### DIFF
--- a/frontend/src/js/components/releases/manifests/ManifestsFilters.tsx
+++ b/frontend/src/js/components/releases/manifests/ManifestsFilters.tsx
@@ -18,14 +18,16 @@ import { Typography } from '@mui/material';
 
 import ChipSelect from '@northern.tech/common-ui/ChipSelect';
 import { ControlledSearch } from '@northern.tech/common-ui/Search';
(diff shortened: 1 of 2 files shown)


frodeha (approved):

frontend/src/js/components/releases/manifests/ManifestsFilters.tsx
```diff
@@ -18,22 +18,24 @@ import { Typography } from '@mui/material';
...
+  const isEnterprise = useSelector(getIsEnterprise);
   const dispatch = useDispatch();
 
   const manifestSearchUpdated = useCallback(searchTerm => dispatch(setManifestsListState({ searchTerm })), [dispatch]);
 
   const onFiltersChange = useCallback(({ name, tags }) => dispatch(setManifestsListState({ selectedTags: tags, searchTerm: name })), [dispatch]);
 
   return (
-    <>
+    <ClickFilter disabled={!isEnterprise}>
```
mineralsfree, line 38: This won't disable the keyboard navigation, so it's possible to use filters still

Image

PR #1930, opened 2026-06-11
MEN-9835 - fix(gui): ensured tags & device types don't contain whitespace

mzedel:
to align w/ backend & client expectations

### DIFF
--- a/frontend/src/js/common-ui/ChipSelect.tsx
+++ b/frontend/src/js/common-ui/ChipSelect.tsx
@@ -45,7 +45,9 @@ export const ChipSelect = ({
       return setValue('');
     }
     const lastIndex = value.lastIndexOf(',');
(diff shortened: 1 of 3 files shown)


frodeha (approved):

frontend/src/js/common-ui/ChipSelect.tsx
```diff
@@ -55,7 +57,7 @@ export const ChipSelect = ({
 
   const onTextInputLeave = (value, setCurrentSelection) => {
     const selection = getValues(name);
-    const nextSelection = unionizeStrings(selection, [value]);
+    const nextSelection = unionizeStrings(selection, [].concat(value.trim().split(/\s+/)).filter(Boolean));
```
mineralsfree, line 60: [].concat does nothing here. 
is's enough to 
`value.trim().split(/\s+/).filter(Boolean)`

PR #1934, opened 2026-06-15
MEN-9832 - chore(gui): let password reset page always show - even when logged in

mzedel:
@mineralsfree mostly as a suggestion as I came across the `App.tsx` content - I think that should be enough...


mineralsfree (approved):
Looks good to me. Thank you! 👍

PR #1945, opened 2026-06-17
MEN-9315: Audit log page adjustments

mineralsfree:


frontend/src/js/common-ui/forms/Filters.tsx
```diff
@@ -42,7 +42,18 @@ const useStyles = makeStyles()(theme => ({
   }
 }));
 
-export const Filters = ({ className = '', defaultValues, filters = [], initialValues, onChange, fieldResetTrigger = '', dirtyField, clearDirty }) => {
+export const Filters = ({
+  className = '',
+  bottomSpacing = 'margin-bottom',
+  topSpacing = 'margin-top',
```
mzedel: This smells bad IMO.
OTOH the per direction doesn't seem like a good longtime idea. But this will also lead to inconsistencies between the pages & their filters - looking at the design of the past deployments + auditlogs + releases => that's the case in there too... 🤡
So using the inconsistencies in the designs as a guide I'd say to keep the margins as is for consistency (or make the bottom `margin-bottom-small`) and push the filters down via the page title.

mzedel (changes requested):
Looks good and thanks for the `ControlledSelect` 👍 - only the spacing props are worrying IMO - it would be nice if we could be able to rely on the consistent usage of the components 🫠

frontend/src/js/common-ui/forms/Filters.tsx
mineralsfree: I'll go back to original margins for now

mzedel (approved):

PR #1947, opened 2026-06-17
fix(deviceconnect): consolidate default config with config.yaml file

alfrunes:
The default nats configuration was silently overwritten by the default configuration file. The actual default does not make much sense in a containerized environments so aligned the default with the value inherited (by accident) from the config file.


alfrunes:
Somewhat related: https://github.com/mendersoftware/mender-server/pull/1949
If configuring a bad NATS URL, the service will look healthy, but won't serve any device connections.

frodeha (approved):

PR #1950, opened 2026-06-17
Remove config file overrides

alfrunes:
Removes all config file overrides and corrects inconsistencies for all the services except create-artifact-worker that depends on these overrides.


frodeha (approved):
LGTM

PR #1951, opened 2026-06-17
MEN-8470: trial expiration header adjustments

mineralsfree:


frontend/src/js/components/header/TrialNotification.tsx
```diff
@@ -11,50 +11,46 @@
...
 const TrialNotification = ({ sectionClassName, expiration }) => {
+  const navigate = useNavigate();
   const expirationDate = dayjs(expiration);
   const duration = dayjs.duration(expirationDate.diff(dayjs()));
-  const daysLeft = Math.floor(duration.asDays());
+  const days = Math.floor(duration.asDays());
+  const months = Math.floor(duration.asMonths());
+  const expiresSoon = months < 4;
+
+  const label = days > 30 ? `${months} ${pluralize('months', months)}` : `${days} ${pluralize('day', days)}`;
```
mzedel, line 41: I guess this _could_ be handled through dayjs too to have the days through a config instead:
```js
// import durationDayJs from 'dayjs/plugin/duration';
// import pluralize from 'pluralize';

// dayjs.extend(durationDayJs);
import relativeTimeDayJs from 'dayjs/plugin/relativeTime';
import updateLocaleDayJs from 'dayjs/plugin/updateLocale';

dayjs.extend(relativeTimeDayJs);
dayjs.extend(updateLocaleDayJs);
```
and then
```js
dayjs.extend(relativeTime, {
  thresholds:[ { l: 'd', r: 1 }, { l: 'dd', r: 30, d: 'day' }, { l: 'M', r: 1 }, { l: 'MM', r: 11, d: 'month' }],
  rounding: Math.floor
});

dayjs.updateLocale('en', { relativeTime: { d: '1 day', dd: '%d days', M: '1 month', MM: '%d months' } });
```
and then the actual relevant part becomes just:
```js
const expirationDate = dayjs(expiration);
const expiresSoon = expirationDate.diff(dayjs(), 'month') < 4;
const label = dayjs().to(expirationDate, true);
```
but OTOH the config is quite a bit more then the single line for the label 😆

mzedel (approved):
👍 seems very nice to me!

frontend/src/js/components/header/TrialNotification.tsx
mineralsfree: I'd keep it simple here, either by accepting the inaccuracy of default day.js or keeping it as it is now. The proposed change will alter the global day.js configuration, which can cause issues in other places of usage.

mzedel: yep, fully agree!

PR #1954, opened 2026-06-17
test(e2e): verify custom roles can list software during deployment creation

p-targowicz:
## What

Adds e2e RBAC coverage for the create-deployment software picker. The three release roles in `05-rbac.spec.ts` are now also granted deploy access, and each asserts that opening **Create a deployment → Select software** lists software without an RBAC `forbidden` error (the picker must render a release).

## Why

A custom (non-admin) role could not list software when creating a deployment — the `/api/management/v1alpha1/deployments/software` endpoints had no matching RBAC permission-set entry, so every non-admin role was rejected with `forbidden by role-based access control`. The existing RBAC tests only exercised the Releases page (old endpoint) with release-only roles, and the deployment-creation tests only ran as admin (which bypasses RBAC) — so the picker + custom-role combination was never covered.

## Note

This is intentionally a **failing regression guard**: it stays red until the corresponding backend RBAC permission-set entry is added (separate change). It exercises the exact gap that caused the incident. The flow has been validated end to end in the enterprise acceptance pipeline (role creation, user creation, and picker navigation all pass); the only failing step is the empty software picker, which is the regression itself.

### DIFF
--- a/frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
+++ b/frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
@@ -11,14 +11,31 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+import type { Page } from '@playwright/test';
+
 import test, { expect } from '../../fixtures/fixtures';
 import { isEnterpriseOrStaging, prepareNewPage } from '../../utils/commands';
 import { releaseTag, selectors, timeouts } from '../../utils/constants';
+import { locateReleaseByName } from '../../utils/utils.ts';
+
+// A custom role with release-read + deploy must be able to list software when creating a deployment.
+// Regression: the /deployments/software endpoints had no RBAC permission-set entry, so every non-admin
+// role was rejected with "forbidden by role-based access control" when opening the software picker.
+const expectSoftwareListingAllowed = async (page: Page) => {
+  await page.locator('.leftFixed.leftNav').getByRole('link', { name: /deployments/i }).click();
+  await page
+    .getByRole('button', { name: /create a deployment/i })
+    .first()
+    .click();
+  await page.getByRole('button', { name: 'Select software' }).click();
+  // the picker must list software rather than fail with a forbidden error
+  await expect(locateReleaseByName(page, 'mender-demo-artifact')).toBeVisible();
+};
 
 const releaseRoles = [
-  { name: 'test-releases-role', permissions: ['Read'], tag: undefined },
-  { name: `test-manage-${releaseTag}-role`, permissions: ['Manage'], tag: releaseTag },
-  { name: `test-ro-${releaseTag}-role`, permissions: ['Read'], tag: releaseTag }
+  { name: 'test-releases-deploy-role', permissions: ['Read'], tag: undefined },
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
```diff
@@ -11,9 +11,32 @@
...
+    { timeout: timeouts.tenSeconds }
+  );
+  await page.locator('.leftFixed.leftNav').getByRole('link', { name: /deployments/i }).click();
+  await page
+    .getByRole('button', { name: /create a deployment/i })
+    .first()
+    .click();
+  await page.getByRole('button', { name: 'Select software' }).click();
+  const response = await softwareResponse;
+  expect(response.status()).not.toBe(403);
```
mzedel:
```suggestion
  await page.locator('.leftFixed.leftNav').getByRole('link', { name: /deployments/i }).click();
  await page
    .getByRole('button', { name: /create a deployment/i })
    .first()
    .click();
  await page.getByRole('button', { name: 'Select software' }).click();
```
to keep the e2e focused on user like behaviour - if the request would get an error the expect right below (`await expect(locateReleaseByName(page, 'mender-demo-artifact')).toBeVisible();`) won't pass

frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
```diff
@@ -100,6 +123,12 @@ test.describe('RBAC functionality', () => {
           await page.getByRole('option', { name: permission }).click();
         }
         await page.press('body', 'Escape');
+        // also grant deploy on a device group so the role can create deployments and reach the software picker
+        await dialog.getByLabel(/Search groups/i).click({ force: true });
+        await page.getByRole('option', { name: 'testgroup' }).click();
+        await dialog.locator(`[id="mui-component-select-groups.0.uiPermissions"]`).click();
+        await page.getByRole('option', { name: 'Deploy', exact: true }).click();
+        await page.press('body', 'Escape');
```
mzedel: that makes the role names somewhat obsolete, but I guess that's tolerable in light of the narrow area the issue is visible ATM 👌

mzedel (commented):
Thanks for the quick submission! I've triggered a pipeline that should run these tests [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14917797322)

frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
p-targowicz: done

p-targowicz:
> Thanks for the quick submission! I've triggered a pipeline that should run these tests [here](https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/14917797322)

thank you !

frontend/tests/e2e_tests/integration/03-advanced/05-rbac.spec.ts
p-targowicz: changed to `test-releases-deploy-role` too

p-targowicz:
@mzedel Ran it multiple times  It's red, but that's as expected the test fails due to the bug is actually there the custom role gets an empty software picker (the RBAC /software gap) 
it'll go green once the backend fix lands

mzedel (approved):

PR #1955, opened 2026-06-18
MEN-9841 - fix(gui): let all manifest related file types be rejected if they are too large - not just `.mender` artifacts

mzedel:


mineralsfree (approved):
😄 👍

PR #1956, opened 2026-06-18
fix(gui): make device detail tab strip scrollable so all tabs stay reachable on mobile (ME-683)

nickanderson:
## Summary

On the device-details drawer, the trailing tabs (e.g. **Troubleshoot**, **Monitoring**) become unreachable on narrow / mobile viewports — there is no way to scroll the tab strip to them. Reported as ME-683 (reproduced on mobile Firefox).

### Root cause

`frontend/src/js/components/devices/ExpandedDevice.tsx` renders `<Tabs>` with MUI's default `variant="standard"`, which has **no horizontal scrolling** — overflowing tabs are clipped, with no scroll buttons and no touch-drag.

This stayed hidden until `72d1fa0c` _("refactor(gui): adopted basedrawer across applicable drawer instances", 2026-04-14)_ migrated the drawer to `<BaseDrawer size="lg">`. `BaseDrawer` pins the paper width (`100vw` on mobile, `67vw` at ≥sm), whereas the previous `<Drawer>` set only `minWidth` and grew with its content — so the user could pan/slide horizontally to reach the hidden tabs. With a fixed width the standard strip now clips and the last tabs are stranded.

### Fix

Make the tab strip scroll on its own, independent of drawer width — the standard MUI pattern:

```jsx
<Tabs ... variant="scrollable" scrollButtons="auto" allowScrollButtonsMobile>
```

`allowScrollButtonsMobile` keeps the scroll arrows available on touch (MUI hides them on mobile by default); native touch-drag scrolling comes with `scrollable`.

### Notes

- Test-first: the first commit is the failing regression test (asserts the scroller carries `tabsClasses.scrollableX`); the fix + snapshot update follow.
- Verified locally with `vitest` — test red before the fix, green after.

### DIFF
--- a/frontend/src/js/components/devices/ExpandedDevice.test.tsx
+++ b/frontend/src/js/components/devices/ExpandedDevice.test.tsx
@@ -11,6 +11,8 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+import { tabsClasses } from '@mui/material';
+
 import { defaultState, render } from '@/testUtils';
 import { EXTERNAL_PROVIDER } from '@northern.tech/store/constants';
 import { undefineds } from '@northern.tech/testing/mockData';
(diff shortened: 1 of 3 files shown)


nickanderson:
Oh that passed locally before humm.

mzedel (approved):

PR #1957, opened 2026-06-19
MEN-9584 + MEN-9670: test device flags support + filtering

mineralsfree:
needs:

- [ ] https://github.com/NorthernTechHQ/nt-gui/pull/793


mineralsfree:
> ## [Codecov](https://app.codecov.io/gh/mendersoftware/mender-server/pull/1957?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mendersoftware) Report
> ✅ All modified and coverable lines are covered by tests.
> 
> 📢 Thoughts on this report? [Let us know!](https://github.com/codecov/feedback/issues/255)

I wonder how reliable this thing is. Do you know if it is configured for frontend @mzedel ? 😄 🤭

mzedel:
> I wonder how reliable this thing is. Do you know if it is configured for frontend @mzedel ? 😄 🤭

@mineralsfree you can check https://app.codecov.io/gh/mendersoftware/mender-server/tree/main/?displayType=tree - that shows _something_ - but I don't trust any of these tools a lot 🫣 (no way we're at 99% coverage in the frontend)

frontend/src/js/common-ui/ConfirmModal.tsx
```diff
@@ -25,14 +25,26 @@ interface ConfirmModalProps {
...
 export const ConfirmModal = (props: ConfirmModalProps) => {
-  const { close, onConfirm, className = '', toType, header, description, open, maxWidth = 'xs', confirmButtonText = 'Confirm' } = props;
+  const { close, onConfirm, className = '', toType, header, description, open, maxWidth = 'xs', confirmButtonText = 'Confirm', isDanger = true } = props;
   const [inputValue, setInputValue] = useState<string>('');
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const onSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setIsSubmitting(true);
+    try {
+      await Promise.resolve(onConfirm());
```
mzedel, line 42: would this not always resolve via the `Promise.resolve()` and not wait for the potential `Promise<unknown>` of the `onConfirm`?

frontend/src/js/common-ui/ConfirmModal.tsx
```diff
@@ -43,16 +55,12 @@ export const ConfirmModal = (props: ConfirmModalProps) => {
...
-            onConfirm();
-            close();
-          }
+          onSubmit
         }
       }}
     >
       <DialogContent>
-        <DialogContentText>{description}</DialogContentText>
+        <div>{description}</div>
```
mzedel: can that be dependent on the `description`? 
```suggestion
        {typeof description === 'string' ? <DialogContentText>{description}</DialogContentText> : description}
```
or maybe 
```suggestion
        <Typography component="div">{description}</Typography>
```
to have it somewhat aligned with the theme support..?

frontend/src/js/components/devices/DeviceStatus.tsx
```diff
@@ -57,38 +57,39 @@ const useStyles = makeStyles()(theme => ({
...
   if (pendingAuthSetsCount) {
     icon = <NumberIcon className={classes.numberIcon} value={pendingAuthSetsCount} />;
     notification = statusTypes.authRequests.notification[deviceStatus] ?? statusTypes.authRequests.notification[DEVICE_STATES.accepted];
     label = `new ${pluralize('request', pendingAuthSetsCount)}`;
-  } else if (Object.values(monitor).some(i => i)) {
-    color = statusTypes.monitor.color;
-    icon = statusTypes.monitor.icon;
-    label = statusTypes.monitor.label;
+  } else if (Object.values(monitor).some(Boolean)) {
+    ({ color, icon, label } = statusTypes.monitor);
```
mzedel, line 76: Oh nice, I didn't know that the destructuring can also be used to reassign 📝

frontend/src/js/components/devices/widgets/FilterItem.tsx
```diff
@@ -30,6 +30,12 @@ const filterOptionsByPlan = {
   enterprise: DEVICE_FILTERING_OPTIONS
 };
 
+const TEST_DEVICE_ATTRIBUTE = 'test_device';
+const testDeviceFilterOptions = {
+  [DEVICE_FILTERING_OPTIONS.$eq.key]: DEVICE_FILTERING_OPTIONS.$eq,
+  [DEVICE_FILTERING_OPTIONS.$ne.key]: DEVICE_FILTERING_OPTIONS.$ne
```
mzedel, line 36: `os`/ `basic` customers don't have the full filtering available - but since you have the `updateFilterKey` already => that should be able to take care of the handling using the `$eq` only...

frontend/src/js/components/devices/widgets/FilterItem.tsx
```diff
@@ -116,7 +127,8 @@ export const FilterItem = ({ attributes, onChange, onSelect, plan, reset, onSave
     onSave({ key, operator, scope, value });
   };
 
-  const filterOptions = plan ? filterOptionsByPlan[plan] : DEVICE_FILTERING_OPTIONS;
+  const isTestDevice = key === TEST_DEVICE_ATTRIBUTE;
```
mzedel: maybe:
```suggestion
  const isTestDeviceFilter = key === TEST_DEVICE_ATTRIBUTE;
```
? as the focus is likely not a single device

frontend/src/js/components/devices/device-details/Identity.tsx
```diff
@@ -11,18 +11,38 @@
...
-  const { created_ts, tier, id, identity_data = {}, status = DEVICE_STATES.accepted } = device;
+const testDeviceTooltip = testDeviceUsed => (
+  <div style={{ maxWidth: 350 }}>
+    <TestDeviceLimit testDeviceUsed={testDeviceUsed} />
+    <Typography className="margin-top-small">
+      Enable up to 10 test devices to bypass rate limits and check in more frequently. Set or remove the ‘test device’ status from the <b>Device actions</b>{' '}
+      menu.
+    </Typography>
+  </div>
+);
```
mzedel: this looks too much like a component to not be a component 😅 
```suggestion
const TestDeviceTooltip = ({ testDeviceUsed }) => (
  <div style={{ maxWidth: 350 }}>
    <TestDeviceLimit testDeviceUsed={testDeviceUsed} />
    <Typography className="margin-top-small">
      Enable up to 10 test devices to bypass rate limits and check in more frequently. Set or remove the ‘test device’ status from the <b>Device actions</b>{' '}
      menu.
    </Typography>
  </div>
);
```

frontend/src/js/common-ui/ConfirmModal.tsx
mineralsfree: if promise or thenable is passed, Promise.resolve returns this promise: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

frontend/src/js/components/devices/widgets/FilterItem.tsx
```diff
@@ -90,6 +102,11 @@ export const FilterItem = ({ attributes, onChange, onSelect, plan, reset, onSave
   const updateFilterKey = ({ key, scope }) => {
     setKey(key);
     setScope(scope);
+    //Special handing for test device flag
+    if (key === TEST_DEVICE_ATTRIBUTE) {
+      setValue('true');
+      setOperator(operator => (operator in testDeviceFilterOptionsByPlan ? operator : DEVICE_FILTERING_OPTIONS.$eq.key));
```
mzedel: this is where I thought the handling for the `os` users would have to happen... say you have:
```js
  const { hasFullFiltering, plan } = useSelector(getTenantCapabilities);
```
in the `Filters` that contain the `FilterItem`, then you could pass that/ the `tenantCapabilities` or as we by now only ever have one `FilterItem` move the selector in and do:
```js
  if (hasFullFiltering) {
    setValue('true');
    setOperator(testDeviceFilterOptions[key]);
  } else {
    setValue(current => current === 'true' ? 'false' : 'true');
    setOperator(DEVICE_FILTERING_OPTIONS.$eq.key);
  }
```
where then I assume we will get false values for the query for all the devices that have no `TEST_DEVICE_ATTRIBUTE` defined 🎉... - but that would then be an issue for the backend/ a migration needed on ALL THE DEVICES!!!!1!1!

frontend/src/js/common-ui/ConfirmModal.tsx
mzedel: oh nice, thanks for the pointer!

mzedel (changes requested):
thanks for the adjustments, but there you will always end up with a `DEVICE_FILTERING_OPTIONS.$eq.key` as operator + `'true'` as value - IOW only ever test devices...

frontend/src/js/components/devices/widgets/FilterItem.tsx
mineralsfree: I thought it was an upsell point. 🤑 I will reach out to Michael, as proposed behaviour might be a bit too confusing for a user IMO 😅 Maybe we could/should allow !eq for basic plans instead?

mzedel (approved):

PR #1960, opened 2026-06-22
fix(deviceconnect): resolve send on closed channel panic in PipeWriter

frodeha:
**Description**
Fix a race condition leading to `send on closed channel` panic in the deviceconnect PipeWriter. When the `mu` channel is closed between the `return len(d), nil` (line 44) and the deferred `r.mu <- struct{}{}` (line 47) the result is the above mentioned panic. I don't think the close handling of this PipeWriter needs to be that extravagant and this simpler version does the trick just fine afaict.

This issue has been provoked a couple of times in our CI test runs ([here](https://gitlab.com/Northern.tech/Mender/mender-server-enterprise/-/jobs/14935180104) for example) and is made quite clear if one runs the tests with the race detector (`--race`).

Ticket: None

### DIFF
--- a/backend/services/deviceconnect/app/playback.go
+++ b/backend/services/deviceconnect/app/playback.go
@@ -24,33 +24,22 @@ const (
 )
 
 type PipeWriter struct {
-	mu        chan struct{}
 	dataChan  chan []byte
+	closed    chan struct{}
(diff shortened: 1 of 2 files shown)


kjaskiewiczz (approved):
nice!

PR #1967, opened 2026-06-24
chore(gui): updated base images after dependabot stopped taking care of it

mzedel:
- the registry workaround seems to have broken the automatic bump detection

FYI @oldgiova - notably https://github.com/mendersoftware/mender-server/pull/1886 landed, but `4.1.x` doesn't have the registry - whereas the images on `main` are getting older...


mineralsfree (approved):

oldgiova (approved):

oldgiova:
@mzedel this should fix it (moving from dependabot to renovate): https://github.com/mendersoftware/mender-server/pull/1968

PR #1970, opened 2026-06-25
MEN-7891 - fix(gui): excluded status from device identity details as it's shown below

mzedel:
the main issue from the ticket was solved in https://github.com/mendersoftware/mender-server/pull/379 - but what's considered to be the device identity information is still misaligned between this section + the device identity text at the top + the status has a separate section all for itself 
Screenshot 2026-06-25 at 13 13 42
vs.
Screenshot 2026-06-25 at 13 13 53


michaelntech (approved):
Makes sense!

mineralsfree (approved):

PR #1981, opened 2026-06-26
ci: don't automatically merge security PRs

oldgiova:
I didn't catch this on the previous PR :see_no_evil: 

Ticket: QA-1479


mzedel (approved):
I don't think we're at risk of these getting merged without review as the review should be required everywhere...

oldgiova:
> I don't think we're at risk of these getting merged without review as the review should be required everywhere...

I agree and I hope we'll land with that.. But let's go step-by-step ;)

PR #1988, opened 2026-06-29
ci: force chore type for all renovate updates to avoid unintended server version bumps

mzedel:
to prevent bumps like: https://github.com/mendersoftware/mender-server/pull/1985 - based on  https://docs.renovatebot.com/semantic-commits/#changing-the-semantic-commit-type - unfortunately there doesn't seem to be a "per dependency group"/ regular config setting (at least based on [the config schema](https://docs.renovatebot.com/renovate-schema.json))

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -6,15 +6,15 @@
   "extends": [
     "config:recommended",
     ":gitSignOff",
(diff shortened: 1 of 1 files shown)


p-targowicz (approved):
lgtm

PR #1993, opened 2026-06-29
ME-680 - fix(gui): clear stale group selection when opening devices from search

mzedel:
only consider a pre-existing group selection when navigating if the url is set accordingly - otherwise assume device selection is preferred

Ticket: ME-680


mzedel:

mineralsfree (approved):

mzedel:

PR #1996, opened 2026-06-29
fix(deviceconnect): restore read deadline to reap dead connections

LudvigAnderson:
MEN-9029 removed SetReadDeadline from the device WebSocket, so a device that disconnects ungracefully sends no Close frame and the half-open connection lingers until the kernel's TCP retransmit timeout (~15 min) fails a ping. Terminal sessions route to the dead socket for that whole window.

Reset the read deadline on every inbound frame (pong, ping, data) so a dead peer is reaped within ~pongWait (1 min) without killing a busy session (MEN-8843). SetReadDeadline errors are propagated, and the ping/pong handlers are unit-tested through a WSConn interface.

**NOTE:** I confirmed this issue myself, but this code is entirely Claude-generated and I only gave it a quick look-through.
**ALSO NOTE:** I edited this after adding more code.


frodeha:
@LudvigAnderson - Sorry for the slowness in providing a review for this. For your information, I'm looking into these types of issues in deviceconnect more generally as we speak and I'll get back to this specific PR when more solid conclusions have been reached.

LudvigAnderson:
@frodeha No problem, thanks for giving me an update and for spending the time to properly investigate deviceconnect. In the end, my wish is that I can get back to the customer and say that their issue will be solved soon, and so thank you for helping out with that :blush:

frodeha:
@LudvigAnderson On the topic of these ReadDeadlines, can I ask how you reproduced this issue? My reason for asking is that my experience conflicts somewhat with the claims in the PR description - specifically this:
> The connection stays half-open for hours and terminal sessions route to the dead socket.

Working on a similar problem in context of the "test in poor network environment" integration test I did not see it remain half-open for "hours", but rather 2-3 minutes at which point the TCP keep-alive's timed out the connection server side and the connection was closed. Of course, 2-3 minutes is longer than the 60 seconds of the ReadDeadline suggested here, but it's also a lot shorter than "hours". 

In addition, we've added changes to maintain only one connection (or websocket request if you will) per device at any given time, so half-open connections would be closed forcibly by the server when mender-connect reconnects (which it does in a loop forever). 

The re-introduction of the application layer ReadDeadline (this PR) is of course possible, but I'm unsure of the actual need for it and in general I'd advocate for piling on less "stuff" (we already have TCP keep-alives) unless there's a proven need. What's the exact issue the customer is facing? And, depending on that, I'd be very interested to see if it's still a problem in saas.14 (the next hosted version). 

As a general comment to the code I'd just remark that we shouldn't be ignoring errors, so I'd argue that should at least be changed before proceeding. 

As I'm going on vacation after today I'll hand this over to @alfrunes (he's the master of this anyway so you're trading up - congrats!). I'm sure you guys will figure it out :) Cheers!

LudvigAnderson:
@alfrunes as mentioned, this code was just Claude-generated to get something started, so it might need edits. But since Frode asked how to reproduce this, just see the video I attached here: https://northerntech.atlassian.net/browse/MEN-9934

It's been a while since this PR was opened, but I just went to Hosted Mender and tried the same approach again, and the behavior is the same. I've had a terminal open for like 10 minutes or something and it's still there even though the device is literally not even on (killed the docker container).

alfrunes:
@LudvigAnderson Thanks for the video. You are right, it seems like the connection is hanging way longer than the 2m30s default keep-alive idle timeout for the TCP connections in Go. I will come back to this PR later, but in the meantime, could you please fixup the error handling?

alfrunes:
I repeated your experiment on staging, it seems like it takes 15m (and not 2m30s) for the connection to time out. :monocle_face:

> `[...] responsetime=16m56.127s`

(it took me almost 2 minutes after opening the session to disconnecting the client)

LudvigAnderson:
@alfrunes the errors are handled now and I also generated a mock for websocket connection and added several tests. The PR is significantly bigger now, so do let me know if I was on the wrong track with anything.

Besides it seems you're right about the 15 minute thing. I thought I tested longer but it might be that some time was spent before actually disconnecting the device, I didn't take that into account. Still, 15 minutes is too long :smile:

backend/services/deviceconnect/api/http/mocks/WSConn.go
alfrunes, line 1: Mocking the connection won't really let you test the actual case we're supposed to fix here (unresponsive idle connections). It does make it easier to play the game of coverage though :)

backend/services/deviceconnect/api/http/device.go
```diff
@@ -226,9 +226,22 @@ func (m *streamMap) Delete(conn stream.Conn) {
...
+//go:generate ../../../../utils/mockgen.sh
+type WSConn interface {
+	ReadMessage() (messageType int, p []byte, err error)
+	WriteMessage(messageType int, data []byte) error
+	WriteControl(messageType int, data []byte, deadline time.Time) error
+	SetReadDeadline(t time.Time) error
+	SetPongHandler(h func(appData string) error)
+	SetPingHandler(h func(appData string) error)
+	Close() error
+}
```
alfrunes, line 240: Side-note, this does create some extra overhead if we ever want to migrate to https://github.com/coder/websocket.

alfrunes (approved):
The change looks good to me. For next time, I would split the `refactor`ing and `tests` into separate commits so that it's both clearer in the commit history and more friendly for the reviewer.

LudvigAnderson:
Yeah codecov was crying to me until I made sure everything was tested :sweat_smile: and yeah sorry about the messy commit, I'll split it better next time :pray:

PR #2001, opened 2026-06-30
MEN-9360 - autocomplete variations

mzedel:
depends on #1999


frontend/src/js/common-ui/forms/CheckboxAutocomplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const selectedValues = value ?? [];
+
+  const sortedOptions = useMemo(() => {
+    const selectedSet = new Set(selectedValues);
+    return [...options].sort((a, b) => {
+      const aSelected = selectedSet.has(a);
+      const bSelected = selectedSet.has(b);
+      if (aSelected && !bSelected) return -1;
+      if (!aSelected && bSelected) return 1;
+      return 0;
```
mineralsfree, line 55: maybe `Number(selectedSet.has(b)) - Number(selectedSet.has(a))`?

frontend/src/js/common-ui/forms/CheckboxAutocomplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+      multiple
+      {...field}
+      value={selectedValues}
+      onChange={(_e, data) => formOnChange(data)}
+      options={sortedOptions}
+      getOptionLabel={option => (typeof option === 'string' ? option : option[labelAttribute])}
+      isOptionEqualToValue={(option, val) => option === val || (option[labelAttribute] && option[labelAttribute] === val[labelAttribute])}
+      renderOption={({ key, ...optionProps }, option, { selected }) => (
+        <li key={key} {...optionProps}>
+          <Checkbox checked={selected} sx={{ mr: 1 }} />
```
mineralsfree: while I am ok with `sx`, we commonly use `margin-right-x-small` in our codebase 😄

frontend/src/js/common-ui/forms/ChipSelect.tsx
```diff
@@ -76,26 +80,28 @@ export const ChipSelect = ({
...
+              ? (values, getItemProps) =>
+                  values.map((option, index) => {
+                    const { key, onDelete, ...tagProps } = getItemProps({ index });
+                    return (
+                      <Chip
+                        label={option}
+                        key={key}
+                        onDelete={onDelete}
+                        size="small"
+                        deleteIcon={<CancelIcon onClick={onDelete} aria-label={`${name}-delete`} />}
```
mineralsfree: I think aria-label here should be dependent on option, not on field name

mineralsfree (commented):
Thank you! 👍

frontend/src/js/common-ui/forms/CheckboxAutocomplete.tsx
mzedel: 🤣 oh my 😆 adjusted 👌

frontend/src/js/common-ui/forms/CheckboxAutocomplete.tsx
mzedel: there I thought the more explicit variant would be ...more explicit? - but I also have no real preference here

mineralsfree: Not a strong preference as well; sorting function is confusing no matter how you write it 😄

mineralsfree (approved):

PR #2004, opened 2026-06-30
MEN-9705: Implement V2 /users/me for own user updates

bahaa-ghazal:


kjaskiewiczz (commented):
looks good, but I thought it will be implemented in the enterprise only

bahaa-ghazal:
> looks good, but I thought it will be implemented in the enterprise only

The one that will be enterprise only is the endpoint that starts the email change flow. This will be used in both

kjaskiewiczz (approved):
looks good, just remember to add it to Basic RBAC permissions when moving to enterprise

PR #2009, opened 2026-07-01
MEN-9927 - feat(gui): expand access to advanced device filter operators for everyone & all plans

mzedel:
~not sure if that's enough already...~ on closer inspection it seems to be enough already!


mineralsfree (approved):
lgtm

PR #2021, opened 2026-07-06
fix(deviceconnect): only allow one connection per device at a time

frodeha:
**Description**
This PR aims to fix a problem where multiple overlapping websocket requests issued from the same device ends up recording the incorrect connectivity status for the device in the deviceconnect database. This makes it impossible for consumers of the management API to initiate terminal sessions and/or file up/downloads with the device because the server consults the database to see if the device is connected or not before allowing the session to be established. This is the cause of the consistent failures in the "test_in_poor_network_environment" integration test(s).

A typical request sequence provoking the problem looks something like this:
* websocket request "1" is initiated from the device
* the server saves the status of the device as "connected" with "version 1" in the database
* -- some time passes --
* websocket request "2" is initiated from the device
* the server saves the status of the device as "connected" with "version 2" in the database
* websocket request "2" is closed by the device
* the server saves the status of the device as "disconnected" in the database, referencing "version 2" which is the most recent and thus the update is performed
* any consumer of the management API tries to establish a connection with the device
* the server sees that the device is "disconnected" in the database and as a result rejects the mgmt request with a 404 (this continues to do so until the device reconnects)
* the original websocket request (request "1") is still connected, but is ignored because of the (incorrect) state in the database

The strategy taken to fix this problem in this PR is to only allow a single websocket request per device at one time. If a "newer" websocket request is received from a device that already has one open, the "older" one is forcibly closed by the server in favor of the "newer" one. 

Here's an [integration pipeline](https://gitlab.com/Northern.tech/Mender/integration/-/jobs/15203380875) with the changes in this PR where the troublesome test is passing. The test fails consistently when used with [4.1.x](https://gitlab.com/Northern.tech/Mender/integration/-/jobs/15189061614#L476) (e.g without these changes). 

_I also intend to port these changes to main once everything looks good as this issue exists there as well._

Ticket: None


backend/services/deviceconnect/app/app.go
```diff
@@ -348,18 +353,26 @@ func (a *app) ShutdownDone() {
...
-	a.shutdownCancels[id] = cancel
-	return id
+	if exclusive {
+		for handleID, handle := range a.shutdownCancels {
+			if handle.id == id {
+				handle.cancel()
+				delete(a.shutdownCancels, handleID)
+			}
+		}
+	}
```
frodeha, line 370: This is the important bit. If we're asked to ensure an "exclusive handle" we close any existing handles for the same "id" before registering the new one.

kjaskiewiczz (approved):
looks good

PR #2027, opened 2026-07-07
Backport support for external environment in ng tests

frodeha:
**Description**
Backporting the structural changes in the ng tests that allow running them against external environments. Note that I changed the environment variables to make sense in os context. Specify `USER=<username>:<password>` instead of `SERVICE_PROVIDER_USER=<username>:<password>` to give the admin credentials. Also note that the variable `PROFESSIONAL_USER` is being ignored in os.

Ticket: None


merlin-northern (approved):
thank you!

PR #2029, opened 2026-07-07
fix(deviceconnect): only allow one connection per device at a time

frodeha:
**Description**
If a new websocket connection is initiated from a device that's already conncted to deviceconnect the older connection is forcibly closed by the server in favor of the newer connection. This is to ensure packets are delivered over the correct/most likely to be alive connection (which is assumed to always be the newest one).

Ticket: None

This is a port of https://github.com/mendersoftware/mender-server/pull/2021 from 4.1.x to main.


kjaskiewiczz (approved):
lgtm

PR #2030, opened 2026-07-07
test(gui): added polling to webhook e2e test

mineralsfree:


frodeha (approved):

frontend/tests/e2e_tests/integration/03-advanced/02-webhooks.spec.ts
```diff
@@ -84,8 +84,12 @@ test.describe('Webhooks Functionality', () => {
   test('shows webhook details for inventory events', async ({ baseUrl, environment, page }, { retry }) => {
     test.skip(environment !== 'enterprise');
     await page.goto(`${baseUrl}ui/settings/integrations`);
+    // the activity list renders as empty until the initial events fetch resolves, so take the baseline
+    // count from the response instead of the DOM
+    const eventsResponse = page.waitForResponse(/\/api\/management\/v1\/iot-manager\/events/);
     await page.getByText(/view details/i).click();
-    const inventoryChangeCount = (await page.getByText(/inventory changed/).all()).length;
+    const events = await (await eventsResponse).json();
+    const inventoryChangeCount = events.filter(({ type }) => type === 'device-inventory-changed').length;
```
mzedel, line 92: I'm aware the change is to work around the flakiness of the test, but this isn't in line with what the tests should do (replicating actual user behaviour), so it should rather be based on the user visible change instead of waiting for the network events

mineralsfree: I agree that the proper fix would be to have a loading state for the component and rely on visual UI, but as of now there is no other way to distinguish `no events` from the loading state, so the only reliable option is to watch for the network event. I assumed it's fine since we use waitForResponse in `triggerDeploymentCreation`

mzedel: IMO the proper fix would be to ensure the processing of the webhook events happens more reliable (as https://github.com/mendersoftware/mender-server/pull/2042 addresses related issues) - IOW nothing the tests could & should handle, as this will only obfuscate the fact the backend isn't handling this as expected.

`triggerDeploymentCreation` handles a case that's not possible to be caught by regular user observation (deployments that get marked as finished right when they are created & thus wouldn't be listed in the active deployments list) - the alternative flow to check finished deployments for the deployment would maybe be an option, but still less reliable...

PR #2032, opened 2026-07-08
Fix/global attempts

mineralsfree:


whbruce:
I believe this PR is result of Mender hub issue [deployments sometimes ignoring update attempts per device settings/8317](https://hub.mender.io/t/deployments-sometimes-ignoring-update-attempts-per-device-settings/8317). How does this PR address requested retry count sometimes being ignored and set to 1?

mineralsfree:
> I believe this PR is result of Mender hub issue [deployments sometimes ignoring update attempts per device settings/8317](https://hub.mender.io/t/deployments-sometimes-ignoring-update-attempts-per-device-settings/8317). How does this PR address requested retry count sometimes being ignored and set to 1?

The root cause was a mismatch between the API (retries) and the UI (attempts = retries + 1): the report was showing retries instead of attempts, and the attempt value in global settings was similarly misaligned

aleksandrychev (approved):
the code looks correct, but I can't judge the logic since I'm not fully familiar with it

PR #2033, opened 2026-07-08
fix(gui): ensure correct global retry count is displayed in deploymen…

mineralsfree:
…t options

### DIFF
--- a/frontend/src/js/components/deployments/CreateDeployment.tsx
+++ b/frontend/src/js/components/deployments/CreateDeployment.tsx
@@ -130,7 +130,13 @@ export const CreateDeployment = ({ deploymentObject = {}, onDismiss, onScheduleS
   const deploymentAnchor = useRef();
   const { classes } = useStyles();
   const methods = useForm<DeploymentFormValues>({ mode: 'onChange', defaultValues: defaultValues });
-  const { control, reset, watch } = methods;
+  const {
+    control,
+    formState: { dirtyFields },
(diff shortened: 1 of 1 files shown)


mzedel (dismissed):

frontend/src/js/components/deployments/CreateDeployment.tsx
```diff
@@ -161,15 +167,24 @@ export const CreateDeployment = ({ deploymentObject = {}, onDismiss, onScheduleS
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [open, reset]);
 
-  // Notify parent of form value changes for URL param sync
+  // the global settings can arrive after the form was initialized, so keep the retries default in sync until an
+  // explicit value was passed in or the user changed the field
+  useEffect(() => {
+    if (!open || deploymentObject.retries != null || dirtyFields.retries) {
```
mzedel: Could `deploymentObject.retries` ever be null? the `defaultValues` define it as a `number`, until it's initialized it's `undefined` AFAICT (& once user touched, it would be dirty) - but maybe I'm missing something?

mineralsfree: No, `!= null` is only a way to check if the value is not `null` or `undefined`. I understand `null` can be confusing in this context, I will switch to `!== undefined` to avoid confusion

mzedel (approved):

PR #2034, opened 2026-07-09
chore: shared helpers for the integration:ng test suites

p-targowicz:
First of a series porting the python backend integration suite to the Go
integration:ng runner, one test file per PR. The python tests are NOT
removed: both suites run in parallel until the port has proven itself,
and retiring the python side will be a separate follow-up decision.

This PR adds the shared foundations in tests/common — device onboarding
(auth requests signed with RSA, ECDSA P-224/256/384/521 and Ed25519
keys), artifact building, context-aware polling, a raw-request helper —
and refactors the existing suites onto them. Strictly gateway-only per
the cross-repo test taxonomy: the runner exercises public management and
device APIs with devices simulated via API; internal APIs and CLIs stay
with the per-service acceptance suites.

The follow-up PRs in the series (one suite file each) build on this one:
health, device auth, deployment lifecycle, artifacts, inventory,
decommission.


p-targowicz:
Superseded by the split series starting at #2079 (one suite per PR, python tests kept running in parallel).

PR #2040, opened 2026-07-10
Adjust e-mail requirements for OS installation (e-mail verification feature and require password for creating accounts)

lluiscampos:

PR #2041, opened 2026-07-10
fix(gui): require password and gate reset-link checkbox on OS

lluiscampos:
On the open-source plan the 'Send reset-link' email feature does not work, and users must always have a password set at creation time.


lluiscampos:
@mzedel And this is what I was actually trying to fix in the first plce (I am using v4.1.1 for which #2040 is not an issue). Also feel free to take over the tests 🙏

lluiscampos:
This one is now part of #2040

PR #2043, opened 2026-07-10
chore(gui): updated terminal message decoder dependency

mineralsfree:
Needs: https://github.com/NorthernTechHQ/nt-gui/pull/823


mzedel (approved):
Thanks!

PR #2047, opened 2026-07-10
Dep/replace zxcvbn w ts

mineralsfree:


frontend/src/js/components/login/PasswordReset.test.tsx
```diff
@@ -42,7 +42,8 @@ describe('PasswordReset Component', () => {
     await act(() => vi.runAllTimersAsync());
   });
 
-  it('works as intended', async () => {
+  // password strength is checked on every keystroke and the zxcvbn-ts scoring exceeds the default 5s test timeout
+  it('works as intended', { timeout: 30000 }, async () => {
```
aleksandrychev, line 46: @mineralsfree Is the intent to keep scoring on each keystroke, or was debouncing considered here?

mineralsfree: I think debounce here won't lead to much better UX. It's also performing slower in testing environment than in browser. Since the password creation input is not used frequently, I'd keep it as is.

aleksandrychev (approved):

mzedel (approved):
woks for me, but the `zxcvbn` lib is one of the very few we were asked about from customers & the dropbox ownership was the key reason they were ok with us relying on it... but OTOH now 10 years unmaintained 😬

mineralsfree:
> woks for me, but the `zxcvbn` lib is one of the very few we were asked about from customers & the dropbox ownership was the key reason they were ok with us relying on it... but OTOH now 10 years unmaintained 😬

Yeah, I think those password strength estimates are not reliable without up-to-date dictionaries

PR #2052, opened 2026-07-13
fix(gui): fixed issue that crashed ungrouped distribution widget

mineralsfree:


frontend/src/js/components/dashboard/widgets/Distribution.tsx
```diff
@@ -264,7 +264,8 @@ export const DistributionReport = ({ onClick, onSave, selection = {}, software:
     data: distribution,
     onClick: onSliceClick
   };
-  const couldHaveDevices = !group || groupsById[group]?.deviceIds.length;
+  const { deviceIds = [], filters = [] } = groupsById[group] || {};
+  const couldHaveDevices = !group || deviceIds.length || filters.length;
```
aleksandrychev:
```suggestion
  const couldHaveDevices = !group || deviceIds.length > 0 || filters.length > 0;
```
otherwise `couldHaveDevices` might be a number instead of a boolean, which I assume is the intended type.

mineralsfree: adjusted 👍

aleksandrychev (approved):

PR #2053, opened 2026-07-13
MEN-9796: Rename system attributes to default

mineralsfree:
Needs: https://github.com/NorthernTechHQ/nt-gui/pull/822


mzedel (commented):
commented in the nt-gui PR, but I don't think we can affect the location as it will break existing links (which caused complaints with the switch from the "legacy" attribute references)

mzedel (approved):

PR #2055, opened 2026-07-13
fix(deviceauth): Allow new authset for accepted devices at max limit

bahaa-ghazal:
Ticket: MEN-9955

Will add ng test in enterprise, since there is no limits is OS

### DIFF
--- a/backend/services/deviceauth/devauth/devauth.go
+++ b/backend/services/deviceauth/devauth/devauth.go
@@ -343,7 +343,6 @@ func (d *DevAuth) handlePreAuthDevice(
 	ctx context.Context,
 	aset *model.AuthSet,
 ) (*model.AuthSet, error) {
-	var deviceAlreadyAccepted bool
 	// check the device status
 	// if the device status is accepted then do not trigger provisioning workflow
 	// this needs to be checked before changing authentication set status
@@ -351,6 +350,7 @@ func (d *DevAuth) handlePreAuthDevice(
 	if err != nil {
 		return nil, err
 	}
+	deviceAlreadyAccepted := dev.Status == model.DevStatusAccepted
 
 	// check if the device is in the decommissioning state
 	if dev.Decommissioning {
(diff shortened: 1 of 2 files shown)


merlin-northern (commented):
I would really love an expicit test scenarion for this in the tests-ng, doable?

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -848,25 +852,27 @@ func (d *DevAuth) AcceptDeviceAuth(ctx context.Context, device_id string, auth_i
 	aset.Status = model.DevStatusAccepted
 	dev.AuthSets = []model.AuthSet{*aset}
 
-	reqId := requestid.FromContext(ctx)
+	if !deviceAlreadyAccepted {
```
merlin-northern: cold you double check, I understand the change, in here we used to start the provision no matter what, and if the device had been accepted and we set the auth set status to accepted this seems not necessary to start the provision workflow, could you doble check that we for sure do not rely on that? I do not think so, but this function is a huge trap.

merlin-northern (dismissed):

backend/services/deviceauth/devauth/devauth.go
bahaa-ghazal: The only case I can think of that this might break is that if we try to accept an authset for a existing device but with different tier then the one accepted. Do we allow that change of tiers for accepted device? or the device must be rejected first then re-accepted with the new tier?

merlin-northern: tiers do not exist in os, so on this you have to fall in ent. but the tier change works by accepting a new authset, and yes, the device stays accepted.

bahaa-ghazal: you know what, it wont hurt do add `provision_device` call in OS. we need that in enterprise because it used the provision_device_v2 which gets attributes as param to the workflow. we can just keep the first part of skipping counting limits if device is  accepted

bahaa-ghazal: I hate the sync ... the enterprise would need a different fix because of tiers

alfrunes (approved):

PR #2062, opened 2026-07-14
fix(deviceconnect): enforce tenant access scope for device connections

alfrunes:
The tenant ID was missing from the NATS topic effectively removing the multi-tenancy scope restrictions from the NATS interface. With this commit, any connections made to a device can only reach devices within the same tenant (without external checks/workarounds).


frodeha:
@alfrunes - some units failing :)

frodeha (approved):

PR #2073, opened 2026-07-17
chore: port health check test to integration:ng

p-targowicz:
Part of the suite-by-suite python-to-Go port (see #2079 for the series
plan). Ports test_health.py; the python test keeps running in parallel.
Builds on #2079.


bahaa-ghazal (commented):
LGTM.
I'm just wondering if this test is necessary, since `healthCheckComposeEnvironment` always run when creating the compose env 🤷

PR #2074, opened 2026-07-17
test(deviceauth): port device auth tests to integration:ng

p-targowicz:
Part of the suite-by-suite python-to-Go port (see #2079). Ports
test_devauth.py: preauthorization, device and authset management, auth
requests across RSA, four ECDSA curves and Ed25519. The python tests
keep running in parallel. Builds on #2079.

### DIFF
--- a/backend/tests/runner/tests/opensource/devauth_management_v2_test.go
+++ b/backend/tests/runner/tests/opensource/devauth_management_v2_test.go
@@ -1,0 +1,1463 @@
+//nolint:all // This is all test code
+package opensource
+
+import (
+	"bytes"
+	"cmp"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"net/http"
+	"path"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/mendersoftware/mender-server/pkg/api/client"
+	"github.com/mendersoftware/mender-server/pkg/utils/types"
+	"github.com/mendersoftware/mender-server/services/deviceauth/model"
+	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+// DevauthManagementV2Suite ports (a subset of) the OS device auth tests
+// from mender-server/backend/tests/integration/test_devauth.py. Enterprise
+// branches (useExistingTenant()) and TestDevAuthCli (covered by acceptance
+// suites) are intentionally not ported.
+type DevauthManagementV2Suite struct {
+	suite.Suite
+
+	APIClient *client.APIClient
+	User      common.User
+	Tenant    common.Tenant
+
+	JWT string
+}
+
+func (i *BackendIntegrationSuite) TestDevauthManagementV2() {
+	suite.Run(i.T(), &DevauthManagementV2Suite{
+		APIClient: i.environment.APIClient(),
+		User:      i.user,
+		Tenant:    i.tenant,
+	})
+}
+
+func (s *DevauthManagementV2Suite) SetupSuite() {
+	require := require.New(s.T())
+
+	ctx := common.BasicAuthContext(s.T().Context(), s.User)
+	token, r, err := s.APIClient.UserAdministrationManagementAPIAPI.Login(ctx).Execute()
+
+	require.NoError(err)
+	require.NotNil(r)
+	require.Equal(http.StatusOK, r.StatusCode)
+	require.NotEmpty(token)
+	s.JWT = token
+}
+
+// ---------------------------------------------------------------------
+// TestPreauth (test_devauth.py::TestPreauth)
+// ---------------------------------------------------------------------
+
+func (s *DevauthManagementV2Suite) TestPreauth() {
+	s.Run("Ok", s.preauthOk)
+	s.Run("FailDuplicate", s.preauthFailDuplicate)
+	s.Run("FailBadRequest", s.preauthFailBadRequest)
+}
+
+func (s *DevauthManagementV2Suite) preauthOk() {
+	// ported from test_devauth.py::TestPreauth::test_ok
+	require := require.New(s.T())
+	assert := assert.New(s.T())
+	ctx := common.JWTAuthContext(s.T().Context(), s.JWT)
+
+	count, err := s.countDevices(ctx, model.DevStatusPreauth)
+	require.NoError(err)
+
+	type devEntry struct {
+		mac, sn string
+		idData  string
+		kp      *common.KeyPair
+		id      string
+	}
+
+	devs := make([]*devEntry, 0, len(common.KeyKinds))
+	for _, kind := range common.KeyKinds {
+		kp, err := common.NewKeyPair(kind)
+		require.NoError(err)
+		mac, sn := randMacSn()
+		devs = append(devs, &devEntry{mac: mac, sn: sn, idData: idDataJSON(mac, sn), kp: kp})
+	}
+
+	for _, d := range devs {
+		id, r, err := s.preauthorize(ctx, d.mac, d.sn, d.kp.PublicKeyPEM(), false)
+		require.NoError(err)
+		require.Equal(http.StatusCreated, r.StatusCode)
+		d.id = id
+	}
+
+	newCount, err := s.countDevices(ctx, model.DevStatusPreauth)
+	require.NoError(err)
+	assert.Equal(count+int32(len(devs)), newCount)
+
+	for _, d := range devs {
+		dev, err := s.getDevice(ctx, d.id)
+		require.NoError(err)
+		assert.Equal(model.DevStatusPreauth, dev.GetStatus())
+		require.Len(dev.AuthSets, 1)
+		aset := dev.AuthSets[0]
+		assert.Equal(d.mac, fmt.Sprint(aset.IdentityData["mac"]))
+		assert.Equal(d.sn, fmt.Sprint(aset.IdentityData["sn"]))
+		assert.Equal(model.DevStatusPreauth, aset.GetStatus())
+
+		// the actual device can obtain an auth token
+		device := common.NewDeviceFromKeyPair(d.kp, d.idData)
+		ok, err := device.SubmitAuthRequest(ctx, s.APIClient, s.Tenant.TenantToken)
+		require.NoError(err)
+		assert.True(ok)
+
+		outdev, err := s.getDevice(ctx, d.id)
+		require.NoError(err)
+		assert.Equal(model.DevStatusAccepted, outdev.GetStatus())
+		require.Len(outdev.AuthSets, 1)
+		assert.Equal(model.DevStatusAccepted, outdev.AuthSets[0].GetStatus())
+	}
+
+	// send the preauth requests again with fresh keys: since the devices
+	// show up with new keys they are immediately accepted with a new
+	// authset, and the old one is rejected
+	for _, d := range devs {
+		kp, err := common.NewKeyPair(common.KeyKindRSA)
+		require.NoError(err)
+		d.kp = kp
+
+		_, r, err := s.preauthorize(ctx, d.mac, d.sn, kp.PublicKeyPEM(), true)
+		require.NoError(err)
+		assert.Equal(http.StatusCreated, r.StatusCode)
+	}
+
+	for _, d := range devs {
+		dev, err := s.getDevice(ctx, d.id)
+		require.NoError(err)
+		assert.Equal(model.DevStatusAccepted, dev.GetStatus())
+		require.Len(dev.AuthSets, 2)
+
+		device := common.NewDeviceFromKeyPair(d.kp, d.idData)
+		ok, err := device.SubmitAuthRequest(ctx, s.APIClient, s.Tenant.TenantToken)
+		require.NoError(err)
+		assert.True(ok)
+
+		outdev, err := s.getDevice(ctx, d.id)
+		require.NoError(err)
+		assert.Equal(model.DevStatusAccepted, outdev.GetStatus())
+		require.Len(outdev.AuthSets, 2)
+
+		foundAccepted := false
+		for _, aset := range outdev.AuthSets {
+			if aset.GetStatus() == model.DevStatusAccepted {
+				foundAccepted = true
+				assert.True(comparePubkeysPEM(aset.GetPubkey(), d.kp.PublicKeyPEM()))
+				break
+			}
+		}
+		assert.True(foundAccepted)
+	}
+}
+
+func (s *DevauthManagementV2Suite) preauthFailDuplicate() {
+	// ported from test_devauth.py::TestPreauth::test_fail_duplicate
+	require := require.New(s.T())
+	assert := assert.New(s.T())
+	ctx := common.JWTAuthContext(s.T().Context(), s.JWT)
+
+	countBefore, err := s.countDevices(ctx, "")
+	require.NoError(err)
+
+	mac, sn := randMacSn()
+	idData := idDataJSON(mac, sn)
+	kp, err := common.NewKeyPair(common.KeyKindRSA)
+	require.NoError(err)
+
+	device := common.NewDeviceFromKeyPair(kp, idData)
+	ok, err := device.SubmitAuthRequest(ctx, s.APIClient, s.Tenant.TenantToken)
+	require.NoError(err)
+	assert.False(ok)
+
+	newKP, err := common.NewKeyPair(common.KeyKindRSA)
+	require.NoError(err)
+
+	_, r, err := s.preauthorize(ctx, mac, sn, newKP.PublicKeyPEM(), false)
+	require.Error(err)
+	assert.Equal(http.StatusConflict, r.StatusCode)
+
+	// device list is unmodified: the failed duplicate did not add a device
+	countAfter, err := s.countDevices(ctx, "")
+	require.NoError(err)
+	assert.Equal(countBefore+1, countAfter)
+
+	// existing device has no new auth sets
+	apiDev, err := s.getDeviceByMacSn(ctx, mac, sn)
+	require.NoError(err)
+	require.Len(apiDev.AuthSets, 1)
+	assert.True(comparePubkeysPEM(apiDev.AuthSets[0].GetPubkey(), kp.PublicKeyPEM()))
+	assert.Equal(model.DevStatusPending, apiDev.AuthSets[0].GetStatus())
+}
+
+func (s *DevauthManagementV2Suite) preauthFailBadRequest() {
+	// ported from test_devauth.py::TestPreauth::test_fail_bad_request
+	require := require.New(s.T())
+	assert := assert.New(s.T())
+	ctx := common.JWTAuthContext(s.T().Context(), s.JWT)
+
+	kp, err := common.NewKeyPair(common.KeyKindRSA)
+	require.NoError(err)
+
+	// id data not json: identity_data is a raw string instead of an
+	// object. The generated client can't build this (identity_data is
+	// typed as an object), so this one goes over a raw request.
+	body, err := json.Marshal(map[string]any{
+		"identity_data": `{"mac": "foo"}`,
+		"pubkey":        kp.PublicKeyPEM(),
+	})
+	require.NoError(err)
+	resp, err := common.RawRequest(ctx, s.APIClient, http.MethodPost, "/api/management/v2/devauth/devices", body)
+	require.NoError(err)
+	defer resp.Body.Close()
+	assert.Equal(http.StatusBadRequest, resp.StatusCode)
+
+	// not a valid key
+	mac, sn := randMacSn()
+	r, err := s.APIClient.DeviceAuthenticationManagementAPIAPI.
+		DeviceAuthManagementPreauthorize(ctx).
+		PreAuthSet(client.PreAuthSet{
+			IdentityData: client.IdentityData{Mac: types.Pointer(mac), Sn: types.Pointer(sn)},
+			Pubkey:       "not a public key",
+		}).
+		Execute()
+	require.Error(err)
+	assert.Equal(http.StatusBadRequest, r.StatusCode)
+}
+
+// ---------------------------------------------------------------------
+// devs_authsets fixture (test_devauth.py::make_devs_with_authsets and
+// friends). Kept local to this file, like the Python original keeps it
+// local to test_devauth.py rather than in testutils/common.py.
+// ---------------------------------------------------------------------
+
+type devFixtureAuthset struct {
+	ID     string
+	Status string
+	KP     *common.KeyPair
+}
+
+type devFixture struct {
+	ID       string
+	Mac      string
+	Sn       string
+	IDData   string
+	Status   string
+	Authsets []*devFixtureAuthset
+}
+
+func (s *DevauthManagementV2Suite) createFirstAuthset(ctx context.Context, kind string) (*devFixture, error) {
+	mac, sn := randMacSn()
+	idData := idDataJSON(mac, sn)
+	kp, err := common.NewKeyPair(kind)
+	if err != nil {
+		return nil, err
+	}
+
+	device := common.NewDeviceFromKeyPair(kp, idData)
+	if _, err := device.SubmitAuthRequest(ctx, s.APIClient, s.Tenant.TenantToken); err != nil {
+		return nil, err
+	}
+
+	apiDev, err := s.getDeviceByMacSn(ctx, mac, sn)
+	if err != nil {
+		return nil, err
+	}
+	aset := findAuthsetByPubkey(apiDev.AuthSets, kp.PublicKeyPEM())
+	if aset == nil {
+		return nil, fmt.Errorf("authset not found after auth request for mac=%s sn=%s", mac, sn)
+	}
+
+	return &devFixture{
+		ID:     apiDev.GetId(),
+		Mac:    mac,
+		Sn:     sn,
+		IDData: idData,
+		Status: model.DevStatusPending,
+		Authsets: []*devFixtureAuthset{
+			{ID: aset.GetId(), Status: model.DevStatusPending, KP: kp},
+		},
+	}, nil
+}
+
+func (s *DevauthManagementV2Suite) addAuthset(ctx context.Context, dev *devFixture, kind string) error {
+	kp, err := common.NewKeyPair(kind)
+	if err != nil {
+		return err
+	}
+
+	device := common.NewDeviceFromKeyPair(kp, dev.IDData)
+	if _, err := device.SubmitAuthRequest(ctx, s.APIClient, s.Tenant.TenantToken); err != nil {
+		return err
+	}
+
+	apiDev, err := s.getDevice(ctx, dev.ID)
+	if err != nil {
+		return err
+	}
+	aset := findAuthsetByPubkey(apiDev.AuthSets, kp.PublicKeyPEM())
+	if aset == nil {
+		return fmt.Errorf("new authset not found for device %s", dev.ID)
+	}
+
+	dev.Authsets = append(dev.Authsets, &devFixtureAuthset{ID: aset.GetId(), Status: model.DevStatusPending, KP: kp})
+	return nil
+}
+
+func (s *DevauthManagementV2Suite) makePendingDevice(ctx context.Context, kind string, numAuthsets int) (*devFixture, error) {
+	dev, err := s.createFirstAuthset(ctx, kind)
+	if err != nil {
+		return nil, err
+	}
+	for i := 1; i < numAuthsets; i++ {
+		if err := s.addAuthset(ctx, dev, kind); err != nil {
+			return nil, err
+		}
+	}
+	return dev, nil
+}
+
+func (s *DevauthManagementV2Suite) makeAcceptedDevice(ctx context.Context, kind string, numAuthsets, numAccepted int) (*devFixture, error) {
+	dev, err := s.makePendingDevice(ctx, kind, numAuthsets)
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < numAccepted; i++ {
+		if err := s.setAuthsetStatus(ctx, dev.ID, dev.Authsets[i].ID, model.DevStatusAccepted); err != nil {
+			return nil, err
+		}
+		dev.Authsets[i].Status = model.DevStatusAccepted
+	}
+	dev.Status = model.DevStatusAccepted
+	return dev, nil
+}
+
+func (s *DevauthManagementV2Suite) makeRejectedDevice(ctx context.Context, kind string, numAuthsets int) (*devFixture, error) {
+	dev, err := s.makePendingDevice(ctx, kind, numAuthsets)
+	if err != nil {
+		return nil, err
+	}
+	for _, aset := range dev.Authsets {
+		if err := s.setAuthsetStatus(ctx, dev.ID, aset.ID, model.DevStatusRejected); err != nil {
+			return nil, err
+		}
+		aset.Status = model.DevStatusRejected
+	}
+	dev.Status = model.DevStatusRejected
+	return dev, nil
+}
+
+func (s *DevauthManagementV2Suite) makePreauthdDevice(ctx context.Context, kind string) (*devFixture, error) {
+	kp, err := common.NewKeyPair(kind)
+	if err != nil {
+		return nil, err
+	}
+	mac, sn := randMacSn()
+
+	id, r, err := s.preauthorize(ctx, mac, sn, kp.PublicKeyPEM(), false)
+	if err != nil {
+		return nil, err
+	}
+	if r.StatusCode != http.StatusCreated {
+		return nil, fmt.Errorf("unexpected preauthorize status %d", r.StatusCode)
+	}
+
+	apiDev, err := s.getDevice(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if len(apiDev.AuthSets) != 1 {
+		return nil, fmt.Errorf("expected 1 authset for preauthd device %s, got %d", id, len(apiDev.AuthSets))
+	}
+
+	return &devFixture{
+		ID:     id,
+		Mac:    mac,
+		Sn:     sn,
+		IDData: idDataJSON(mac, sn),
+		Status: model.DevStatusPreauth,
+		Authsets: []*devFixtureAuthset{
+			{ID: apiDev.AuthSets[0].GetId(), Status: model.DevStatusPreauth, KP: kp},
+		},
+	}, nil
+}
+
+func (s *DevauthManagementV2Suite) makePreauthdDeviceWithPending(ctx context.Context, kind string, numPending int) (*devFixture, error) {
+	dev, err := s.makePreauthdDevice(ctx, kind)
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < numPending; i++ {
+		if err := s.addAuthset(ctx, dev, common.KeyKindRSA); err != nil {
+			return nil, err
+		}
+	}
+	return dev, nil
+}
+
+// makeDevsWithAuthsets ports make_devs_with_authsets: a good number of
+// devices, some with >1 authsets, in every status.
+func (s *DevauthManagementV2Suite) makeDevsWithAuthsets(ctx context.Context) ([]*devFixture, error) {
+	var devs []*devFixture
+	add := func(dev *devFixture, err error) error {
+		if err != nil {
+			return err
+		}
+		devs = append(devs, dev)
+		return nil
+	}
+
+	// vanilla 'pending' devices, single authset
+	for i := 0; i < 3; i++ {
+		if err := add(s.makePendingDevice(ctx, common.KeyKindRSA, 1)); err != nil {
+			return nil, err
+		}
+	}
+	for i := 0; i < 2; i++ {
+		if err := add(s.makePendingDevice(ctx, common.KeyKindECP256, 1)); err != nil {
+			return nil, err
+		}
+	}
+	if err := add(s.makePendingDevice(ctx, common.KeyKindEd25519, 1)); err != nil {
+		return nil, err
+	}
+
+	// pending devices with >1 authsets
+	for i := 0; i < 2; i++ {
+		if err := add(s.makePendingDevice(ctx, common.KeyKindRSA, 3)); err != nil {
+			return nil, err
+		}
+	}
+	for i := 0; i < 2; i++ {
+		if err := add(s.makePendingDevice(ctx, common.KeyKindECP256, 3)); err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1467 @@
...
+	require.NoError(err)
+	require.NotNil(r)
+	require.Equal(http.StatusOK, r.StatusCode)
+	require.NotEmpty(token)
+	s.JWT = token
+}
+
+// TearDownSuite decommissions every device this suite created, so this
+// suite doesn't leak devices into other suites that assert exact (not
+// delta) device/inventory counts against the shared environment.
```
bahaa-ghazal: I don't think this is a issue as I mentioned, each suite has its own tenant

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1467 @@
...
+	if id != "" {
+		s.createdDevices = append(s.createdDevices, id)
+	}
+}
+
+// ---------------------------------------------------------------------
+// TestPreauth (test_devauth.py::TestPreauth)
+// ---------------------------------------------------------------------
+
+func (s *DevauthManagementV2Suite) TestPreauthOk() {
```
bahaa-ghazal: Can you refactor the test to use [this](https://github.com/mendersoftware/mender-server/blob/f2f26e370bdd63ad68edc1616d1603ac3446d615/backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go#L88-L125) format? One test and muntliple tests case using `.Run()`

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1467 @@
...
+		{model.DevStatusAccepted, 5, 2},
+		{model.DevStatusPending, 2, 2},
+	}
+
+	for _, tc := range cases {
+		apiDevs, err := s.listDevices(ctx, tc.status, int32(tc.page), int32(tc.perPage), ids...)
+		require.NoError(err)
+
+		refDevs := filterAndPageDevs(devs, tc.page, tc.perPage, tc.status)
+		s.compareDevs(refDevs, apiDevs)
```
bahaa-ghazal: Same here, use `.Run()` so that one failing case doesn't stop the whole test.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1467 @@
...
+
+	for _, id := range []string{"foo", "bar"} {
+		_, r, err := s.APIClient.DeviceAuthenticationManagementAPIAPI.
+			DeviceAuthManagementGetDevice(ctx, id).Execute()
+		require.Error(err)
+		assert.Equal(http.StatusNotFound, r.StatusCode)
+	}
+}
+
+func (s *DevauthManagementV2Suite) TestDeviceMgmtDeleteDeviceOk() {
```
bahaa-ghazal: Here also for `TestDeviceMgmtDeleteDevice`, one test and multiple test cases. Also applies for the rest of tests that are similar.

bahaa-ghazal (changes requested):
looks good, would appreciate if could use the same style as in [here](https://github.com/mendersoftware/mender-server/blob/f2f26e370bdd63ad68edc1616d1603ac3446d615/backend/tests/runner/tests/opensource/deployments_management_v1alpha1_test.go#L88-L125)

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: done

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: done

bahaa-ghazal (commented):
LGTM! 
I would also like @merlin-northern to take a look at it to be more safe I didn't miss anything

p-targowicz:
> LGTM! I would also like @merlin-northern to take a look at it to be more safe I didn't miss anything

absolutely @merlin-northern please take a look whenever you have some spare time 🙏

bahaa-ghazal (dismissed):

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/mendersoftware/mender-server/pkg/api/client"
+	"github.com/mendersoftware/mender-server/pkg/utils/types"
+	"github.com/mendersoftware/mender-server/services/deviceauth/model"
+	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
```
merlin-northern, line 25:
```suggestion
	"github.com/google/uuid"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	"github.com/stretchr/testify/suite"

	"github.com/mendersoftware/mender-server/pkg/api/client"
	"github.com/mendersoftware/mender-server/pkg/utils/types"
	"github.com/mendersoftware/mender-server/services/deviceauth/model"
	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
```
nitpicking.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+}
+
+// ---------------------------------------------------------------------
+// TestPreauth (test_devauth.py::TestPreauth)
+// ---------------------------------------------------------------------
+
+func (s *DevauthManagementV2Suite) TestPreauth() {
+	s.Run("Ok", s.preauthOk)
+	s.Run("FailDuplicate", s.preauthFailDuplicate)
+	s.Run("FailBadRequest", s.preauthFailBadRequest)
```
merlin-northern, line 70: looking for instance at the `compareDevs` and  `preauthOk` it is impossible to distinguish between the test case and the helper function. there should be a clean distinction.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+// local to test_devauth.py rather than in testutils/common.py.
+// ---------------------------------------------------------------------
+
+type devFixtureAuthset struct {
+	ID     string
+	Status string
+	KP     *common.KeyPair
+}
+
+type devFixture struct {
```
merlin-northern, line 258: why is it called a fixture?

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		page++
+	}
+}
+
+// listDevices lists devices, optionally restricted to a set of device ids.
+// The shared environment isn't reset between tests (see
+// docker_compose_environment.go), so callers that need to compare the full
+// listing/paging/count against a locally tracked device set must pass ids
+// to filter out devices left behind by other tests/suites.
+func (s *DevauthManagementV2Suite) listDevices(ctx context.Context, status string, page, perPage int32, ids ...string) ([]client.Device, error) {
```
merlin-northern, line 1394: we have been there. the python tests have 10 tons of helpers, abstraction, and stuff that mask the true purpose of the test and makes it impossible to read now and insanely difficult for future us to maintain. this is what the python tests proved: the insane design patterns and strange abstractions have no place in the test code. the test code should be an easy to read and super simple to understand reference for how the product is expected to work, even so at the cost of duplicating code. what we did with the tests-ng was to write the test cases calling directly the pkg/client everywhere where possible (AcceptWait func is one important exception, about which below). this makes the test maintainable, why? because in 10 years someone reading the test code sees DeviceAuthManagementListDevices().Execute() and immediately sees what is being tested, and why, without jumps to definitions in other places. there are cases where we need to have a common code as is the case with accepting a device (see [AcceptWait](https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/tests/runner/tests/enterprise/feature_devicetiers_test.go#L192)) so some balance is eeded.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		return 0, err
+	}
+	return count.GetCount(), nil
+}
+
+func (s *DevauthManagementV2Suite) getDevice(ctx context.Context, id string) (*client.Device, error) {
+	dev, _, err := s.APIClient.DeviceAuthenticationManagementAPIAPI.
+		DeviceAuthManagementGetDevice(ctx, id).Execute()
+	return dev, err
+}
```
merlin-northern, line 1366: see below, this is another helper that makes things worse.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	ctx := common.JWTAuthContext(s.T().Context(), s.JWT)
+
+	kp, err := common.NewKeyPair(common.KeyKindRSA)
+	require.NoError(err)
+
+	// id data not json: identity_data is a raw string instead of an
+	// object. The generated client can't build this (identity_data is
+	// typed as an object), so this one goes over a raw request.
+	body, err := json.Marshal(map[string]any{
+		"identity_data": `{"mac": "foo"}`,
```
merlin-northern, line 224: this really looks like a trap and a riddle, takes some time to understand that the `` make all the difference.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		"pubkey":        kp.PublicKeyPEM(),
+	})
+	require.NoError(err)
+	resp, err := common.RawRequest(ctx, s.APIClient, http.MethodPost, "/api/management/v2/devauth/devices", body)
```
merlin-northern, line 228: hardcoded endpoint, and method, has nothing to do with the code or spec. I think the correct way to inject the bad payload is to override `RoundTrip` in the `http.RoundTripper` interface and set a new `HTTPClient` in the openapi client configuration with custom `Transport`.

merlin-northern (commented):
okey, I would like to discuss during the MSTM the policy on rewriting the tests. let's have a decision and a defined approach. one way is: leave the python tests as legacy, develop new tests in the new framework, and every time we modify something we write/update the test case in tests-ng. also I would like to discuss the approach to AI generated code and set the policy for that.

I am not blocking this, but I would kindly request no more porting before we discuss&agree as mentioned above.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+// DevauthManagementV2Suite ports (a subset of) the OS device auth tests
+// from mender-server/backend/tests/integration/test_devauth.py. Enterprise
+// branches (useExistingTenant()) and TestDevAuthCli (covered by acceptance
+// suites) are intentionally not ported.
```
frodeha: I'd prefer if we can remove _all_ of these types of AI comments from the file. They serve no purpose (beyond being extremely annoying, which I suppose can be a purpose). I don't want to see anything with ".py" or "Python original" or anything like that personally.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+func (s *DevauthManagementV2Suite) SetupSuite() {
+	require := require.New(s.T())
+
+	ctx := common.BasicAuthContext(s.T().Context(), s.User)
+	token, r, err := s.APIClient.UserAdministrationManagementAPIAPI.Login(ctx).Execute()
+
+	require.NoError(err)
+	require.NotNil(r)
+	require.Equal(http.StatusOK, r.StatusCode)
+	require.NotEmpty(token)
```
frodeha: The client will produce an error on non success status codes, so the handling of `r` here is unnecessary. Replace with:

```suggestion
	token, r, err := s.APIClient.UserAdministrationManagementAPIAPI.Login(ctx).Execute()
	require.NoError(err)
	require.NotEmpty(token)
```

This generally applies to the entire file unless there's a very specific reason to check it.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
frodeha: I agree with Peter's comment - let's fix it. Test member functions (even if private) should start with `test`. So in this case `s.preauthOk` should be `s.testPreauthOk`, `s.preauthFailDuplicate` should be `s.testPreauthFailDuplicate` etc.

I'm also pushing for us converging on a common-ish pattern for the sub test run "names" themselves (the first string arguments here). The pattern we're sort of moving towards is `"Success/<specifier>"` for test cases that is testing happy paths and `"Failure/<specifier>"` for test cases that are testing error paths. I know we're not consistent in this (yet), but I'd like to at least not introduce yet another way here.

See here for examples: https://github.com/mendersoftware/mender-server/blob/main/backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+
+	count, err := s.countDevices(ctx, model.DevStatusPreauth)
+	require.NoError(err)
+
+	type devEntry struct {
+		mac, sn string
+		idData  string
+		kp      *common.KeyPair
+		id      string
+	}
```
frodeha: I'm quite certain we don't need this at all. We can just use `common.Device`. Use `NewDeviceFromKeyPair` to create the device with kp from `common.NewKeyPair`. This of course applies to all test cases in the file not just this one.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+
+	devs := make([]*devEntry, 0, len(common.KeyKinds))
+	for _, kind := range common.KeyKinds {
+		kp, err := common.NewKeyPair(kind)
+		require.NoError(err)
+		mac, sn := randMacSn()
```
frodeha: Use `common.RandomMAC()` instead to make an actual (although fake) mac address instead of a uuid (which is an odd choice).  I see no value in `"sn"` at all so I suggest we keep it simple and remove it. This also applies everywhere.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	// id data not json: identity_data is a raw string instead of an
+	// object. The generated client can't build this (identity_data is
+	// typed as an object), so this one goes over a raw request.
+	body, err := json.Marshal(map[string]any{
+		"identity_data": `{"mac": "foo"}`,
+		"pubkey":        kp.PublicKeyPEM(),
+	})
+	require.NoError(err)
+	resp, err := common.RawRequest(ctx, s.APIClient, http.MethodPost, "/api/management/v2/devauth/devices", body)
+	require.NoError(err)
```
frodeha: No, this is not the way. We can send invalid stuff by assigning to `PreAuthSet.AdditionalProperties["identity_data"]` (which is a map[string]any) instead of `PreAtuhSet.IdentityData`. 

I think something like this should work:
```suggestion
	r, err := s.APIClient.DeviceAuthenticationManagementAPIAPI.
		DeviceAuthManagementPreauthorize(ctx).
		PreAuthSet(client.PreAuthSet{
			AdditionalProperties: map[string]any{"identity_data": `{"mac": "foo"}`},
			Pubkey:               kp.PublicKeyPEM(),
		}).
		Execute()
	require.Error(err)
```

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+type devFixture struct {
+	ID       string
+	Mac      string
+	Sn       string
+	IDData   string
+	Status   string
+	Authsets []*devFixtureAuthset
+}
+
+func (s *DevauthManagementV2Suite) createFirstAuthset(ctx context.Context, kind string) (*devFixture, error) {
```
frodeha: Please, let's have helper functions go below the test functions in the file. I'm still scrolling looking for where this is actually being used.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+}
+
+type devFixture struct {
+	ID       string
+	Mac      string
+	Sn       string
+	IDData   string
+	Status   string
+	Authsets []*devFixtureAuthset
+}
```
frodeha: Pretty sure we don't need these. Use `common.Device`, `client.Device`, `client.Authset` instead. The helper functions are (for some reason) from these types to these "fixtures" (another term I'd prefer not to see here).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	devs := make([]*devEntry, 0, len(common.KeyKinds))
+	for _, kind := range common.KeyKinds {
+		kp, err := common.NewKeyPair(kind)
+		require.NoError(err)
+		mac, sn := randMacSn()
+		devs = append(devs, &devEntry{mac: mac, sn: sn, idData: idDataJSON(mac, sn), kp: kp})
+	}
+
+	for _, d := range devs {
+		id, r, err := s.preauthorize(ctx, d.mac, d.sn, d.kp.PublicKeyPEM(), false)
```
frodeha: This is the entire point of what we're testing in these tests and I'd much prefer to not have to scroll a good 1k lines down to find out what it does. Let us do the API call inline in these tests without all the indirection. At least that way it's easier to see what's going on.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+}
+
+func (s *DevauthManagementV2Suite) waitDeviceGone(ctx context.Context, id string) bool {
+	err := common.RetryUntil(ctx, 10*time.Second, 200*time.Millisecond, func() (bool, error) {
+		_, r, _ := s.APIClient.DeviceAuthenticationManagementAPIAPI.
+			DeviceAuthManagementGetDevice(ctx, id).Execute()
+		return r != nil && r.StatusCode == http.StatusNotFound, nil
+	})
+	return err == nil
+}
```
frodeha: I would inline these everywhere. These also look like cases where we don't care about `r` as `err` will be non-nil unless the status code is success.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
frodeha: I 100% agree with Peter here. This should be in-lined wherever it's used and called with exactly the query arguments necessary. We don't need this helper.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+			if d.IdentityData != nil && d.IdentityData.GetMac() == mac && d.IdentityData.GetSn() == sn {
+				return &d, nil
+			}
+		}
+		if len(devices) < int(perPage) {
+			return nil, fmt.Errorf("device not found by id data (mac=%s sn=%s)", mac, sn)
+		}
+		page++
+	}
+}
```
frodeha: Agree with Peter also on these, let's inline them

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		require.NoError(err)
+		var header map[string]any
+		require.NoError(json.Unmarshal(headerRaw, &header))
+		assert.Equal("JWT", header["typ"])
+	}
+}
+
+// ---------------------------------------------------------------------
+// shared helpers
+// ---------------------------------------------------------------------
```
frodeha, line 862: Remove me and others like me

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1463 @@
...
+	assert.Equal(http.StatusBadRequest, r.StatusCode)
+
+	// bad request - invalid payload: the generated client can't send an
+	// arbitrary body via the typed Status model, so this one goes over a
+	// raw request.
+	reqPath := fmt.Sprintf("/api/management/v2/devauth/devices/%s/auth/%s/status", devs[0].ID, devs[0].Authsets[0].ID)
+	resp, err := common.RawRequest(ctx, s.APIClient, http.MethodPut, reqPath, []byte(`{"foo": "bar"}`))
+	require.NoError(err)
+	defer resp.Body.Close()
+	assert.Equal(http.StatusBadRequest, resp.StatusCode)
```
frodeha: I would rather not have this her at all if we have to hard-code URLs in this way. It's not a high value test either so I say remove it.

frodeha (changes requested):
Alright, @p-targowicz. I've poured over this now and left a bunch of comments. Many of them are general and apply widely across the entire PR (not just in the place I commented). 

I assume many of my comments apply to the other PRs as well, so in interest of time it would be best if you have a look at those and fix any similar things before the first review of the subsequent ones (saves me from making the same comments again!) :)

Have a go (pun intended - hehe) and let me know if something is unclear 👍

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done — applied the same across the chain (#2075–#2078).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done — applied the same across the chain (#2075–#2078).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done — applied the same across the chain (#2075–#2078).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done — applied the same across the chain (#2075–#2078).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done — applied the same across the chain (#2075–#2078).

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: All the leaf cruft is gone (testDevice, devFixture, authSetBy*, compareDev, and the listDevices/countDevices/getDevice/preauthorize wrappers as requested what's left below the divider is -`the device-population setup` (makeDevsWithAuthsets + createPending/Accepted) which Peter asked to keep as shared onboarding, `findDeviceIDByMac` (this is the fix for the CI timeout inlining a full device listing reintroduces it) and two async pollers + samePubkey

Are you OK keeping those or do you want the population setup inlined into each test too ? @frodeha

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
frodeha: We're not using common.Device and `devEntry` (what "this" implies in my previous comment) is still here.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+			mac:    mac.String(),
+			idData: macIDData(mac.String()),
+			kp:     kp,
+		})
+	}
+
+	for _, d := range devs {
+		r, err := devauthm.DeviceAuthManagementPreauthorize(ctx).
+			PreAuthSet(client.PreAuthSet{
+				IdentityData: client.IdentityData{Mac: types.Pointer(d.mac)},
```
frodeha: I must have missed this in my first review, but we're standardizing on using `client.PtrString` and friends instead of `types.Pointer`. Can we please change it everywhere?

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+	require.NoError(err)
+
+	// the environment isn't reset between tests, so restrict the listing to
+	// the devices this test created
+	ids := deviceIDs(devs)
+
+	cases := []struct {
+		status        string
+		page, perPage int
+	}{
```
frodeha, line 270: Here we need some descriptive names for each test case (I have no idea what these are testing without reading the test code). Add a name property to the struct and add a descriptive name to each case (use PascalCase). 

After giving every test case a fitting name, update the for loop on line 289 to run each case as a subtest by calling `s.Run(tc.name, func(){ ... })`. When you do this, it's important that we move the `require := ...` and `assert := ...` that's currently at the top of this function (line 260 and 261) inside the callback function passed to `s.Run`. It's ok and necessary to change line 265 to be `require.NoError(s.T(), err)` as a consequence of this.

Both of these apply everywhere we have table driven tests like this.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+		if tc.page > 0 {
+			req = req.Page(int32(tc.page))
+		}
+		if tc.perPage > 0 {
+			req = req.PerPage(int32(tc.perPage))
+		}
+		apiDevs, _, err := req.Execute()
+		require.NoError(err)
+
+		want := filterAndPage(devs, tc.page, tc.perPage, tc.status)
```
frodeha: We don't need to do this, it's good enough to assert that all the devices returned from the request has the expected status (tc.status) in the loop below

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+
+	devauthm := s.APIClient.DeviceAuthenticationManagementAPIAPI
+
+	devs, err := s.makeDevsWithAuthsets(ctx)
+	require.NoError(err)
+
+	// decommission a pending device
+	pending := filterAndPage(devs, 0, 0, model.DevStatusPending)
+	require.NotEmpty(pending)
+	devPending := pending[0]
```
frodeha: Do a simple for loop over devs to find a pending one instead. Applies to all uses of the `filterAndPage` function afaict, keep it simple and readable.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+			require.NoError(err)
+			require.Equal(http.StatusOK, r.StatusCode)
+			assert.Equal(aset.GetStatus(), status.GetStatus())
+		}
+	}
+
+	for _, tc := range []struct{ did, aid string }{
+		{devs[0].id, "foo"},
+		{"foo", "bar"},
+	} {
```
frodeha: No, extract the test cases from the loop definition, this is hard to read.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+	assert := assert.New(s.T())
+	ctx := common.JWTAuthContext(s.T().Context(), s.JWT)
+
+	devauthm := s.APIClient.DeviceAuthenticationManagementAPIAPI
+
+	type devEntry struct {
+		mac    string
+		idData string
+		kp     *common.KeyPair
+	}
```
frodeha: Replace with `common.Device`

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+
+// keyFor returns the local key pair that produced the given API auth set.
+func (d *managedDevice) keyFor(aset client.AuthSet) *common.KeyPair {
+	for _, kp := range d.keys {
+		if aset.Pubkey != nil && samePubkey(*aset.Pubkey, kp.PublicKeyPEM()) {
+			return kp
+		}
+	}
+	return nil
+}
```
frodeha: Remove me, use `common.Device` instead

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+	lo := (page - 1) * perPage
+	hi := lo + perPage
+	if lo > len(filtered) {
+		lo = len(filtered)
+	}
+	if hi > len(filtered) {
+		hi = len(filtered)
+	}
+	return filtered[lo:hi]
+}
```
frodeha: Delete me

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1250 @@
...
+		for i := range devices {
+			if devices[i].IdentityData != nil && devices[i].IdentityData.GetMac() == mac {
+				return &devices[i], nil
+			}
+		}
+		if len(devices) < int(perPage) {
+			return nil, fmt.Errorf("device with mac %s not found", mac)
+		}
+	}
+}
```
frodeha: I believe this will become unnecessary once we use `common.Device` everywhere

frodeha (changes requested):

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Done.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Removed managedDevice - replaced with  testDevice

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: yep becomes unnecessary once we use common.Device

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1237 @@
...
+	devauthm := s.APIClient.DeviceAuthenticationManagementAPIAPI
+
+	statuses := []string{
+		"",
+		model.DevStatusPending,
+		model.DevStatusAccepted,
+		model.DevStatusRejected,
+		model.DevStatusPreauth,
+		model.DevStatusNoAuth,
+	}
```
frodeha, line 421: These are cases to and should get the name + s.Run treatment.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
frodeha: Not done?

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
```diff
@@ -0,0 +1,1237 @@
...
+// authSetDevice returns the common.Device whose key produced the given API auth
+// set, matched by public key.
+func (d *testDevice) authSetDevice(aset client.AuthSet) *common.Device {
+	for _, ad := range d.authSets {
+		if aset.Pubkey != nil && samePubkey(*aset.Pubkey, common.ExportPublicKeyPEM(ad.PublicKey)) {
+			return ad
+		}
+	}
+	return nil
+}
```
frodeha: More stuff to be replaced with `common.Device` and/or `client.Device` (which yet again should remove the need for a lot of the cruft below).

frodeha (changes requested):
Some more. It's also high time to start running the tests and confirm that stuff is actually working (which is not the case currently). Appreciate if we can have green run before the next review :)

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: replaced `testDevice` with `common.Device`

frodeha:
@p-targowicz - apologies if this isn't ready for review (I had a look despite you not requesting a review), but there's comments that haven't been fixed. I think it's a good idea to review the open comments resolve the ones that have been fixed so it's easier to get an overview for both of us? Wdyt?

Also, we've today merged changes to the common.Device utility that will require this (and other PR's) to be rebased on main.

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: done

frodeha:
@p-targowicz - we need to rebase this on main before we can proceed

backend/tests/runner/tests/opensource/devauth_management_v2_test.go
p-targowicz: Now it should be done :D sorry !

backend/Makefile
```diff
@@ -118,7 +118,7 @@ test-integration:
 
 .PHONY: test-integration-ng
 test-integration-ng:
-	go -C tests/runner test -count=1 -v -p=1 -tags=nopkcs11 ./...
+	go -C tests/runner test -count=1 -timeout 30m -v -p=1 -tags=nopkcs11 ./...
```
frodeha: We don't need this right? The tests complete in less than 5 minutes in CI unless I'm missing something?

frodeha (approved):
Aight, this looks reasonable to me now (bar the one timeout change I commented on), so let's get this thing merged shall we? Thanks for all the work, tokens and patience :)

PR #2079, opened 2026-07-17
chore: add context-aware polling helper for integration tests

p-targowicz:
First of a series porting the python backend integration suite to the Go integration:ng runner, one file per PR. The python tests are NOT removed: both suites run in parallel until the port has proven itself; retiring the python side is a separate follow-up decision.

This PR adds the RetryUntil polling helper the later suites build on. Series order: retry, crypto, device onboarding, artifact builder, raw request helper, two existing-suite refactors, useradm claims, then the ported suites (health, device auth, deployment lifecycle, artifacts, inventory, decommission).

### DIFF
--- a/backend/tests/runner/tests/common/retry.go
+++ b/backend/tests/runner/tests/common/retry.go
@@ -1,0 +1,50 @@
+//nolint:all // This is all test code
+package common
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+// RetryUntil calls fn repeatedly, waiting period between attempts, until fn
+// reports success (true, nil), fn returns an error (which aborts the retry
+// loop immediately), ctx is done, or the timeout budget runs out.
+//
+// fn decides for itself whether an error is fatal (return it, aborting the
+// loop) or should just be treated as "not ready yet" (swallow it and return
+// false, nil).
+//
+// Returns nil on success, fn's error if it aborted the loop, ctx.Err() if
+// ctx was done, or a timeout error if the budget ran out without success.
+func RetryUntil(ctx context.Context, timeout, period time.Duration, fn func() (bool, error)) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


backend/tests/runner/tests/common/retry.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+		if ok {
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("timed out after %s waiting for condition", timeout)
+		}
+
+		select {
+		case <-ctx.Done():
```
bahaa-ghazal, line 42: Could we use a [ticker](https://pkg.go.dev/time#NewTicker) for this? here is an example of how we use a ticker:
https://github.com/mendersoftware/mender-server/blob/7ecd44ab79250c52b78412dc282bf1c8d4d37165/backend/services/deviceconnect/api/http/device.go#L375-L388

backend/tests/runner/tests/common/retry.go
```diff
@@ -0,0 +1,40 @@
...
+// fn decides for itself whether an error is fatal (return it, aborting the
+// loop) or should just be treated as "not ready yet" (swallow it and return
+// false, nil).
+//
+// Returns nil on success, fn's error if it aborted the loop, ctx.Err() if
+// ctx was done, or a timeout error if the budget ran out without success.
+func RetryUntil(ctx context.Context, timeout, period time.Duration, fn func() (bool, error)) error {
+	deadline := time.Now().Add(timeout)
+	for {
+		ok, err := fn()
```
bahaa-ghazal, line 33: It is also nice to check if the context is fine before the first call
```go
if err := ctx.Err(); err != nil {
		return err
	}
```

bahaa-ghazal (changes requested):

backend/tests/runner/tests/common/retry.go
p-targowicz: done valid points 💯

bahaa-ghazal (approved):
Nice, looks good!

PR #2081, opened 2026-07-17
chore: add device onboarding helper for integration tests

p-targowicz:
Part of the one-file-per-PR python-to-Go test port series (see #2079). The python tests keep running in parallel.

### DIFF
--- a/backend/tests/runner/tests/common/device.go
+++ b/backend/tests/runner/tests/common/device.go
@@ -1,0 +1,280 @@
+//nolint:all // This is all test code
+package common
+
+import (
+	"context"
+	"crypto"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


bahaa-ghazal (commented):
FYI, there is [TestDevice](https://github.com/mendersoftware/mender-server-enterprise/blob/770582af0485f1561fc597d43b16c130495d75e4/backend/tests/runner/tests/enterprise/feature_devicetiers_test.go#L50-L51) in enterprise which has simliar methods. Maybe it is better to backport it to open source? Just making sure we don't end up duplicated code

bahaa-ghazal (approved):
Looks good!

PR #2082, opened 2026-07-17
chore: add artifact builder helper for integration tests

p-targowicz:
Part of the one-file-per-PR python-to-Go test port series (see #2079). The python tests keep running in parallel.


bahaa-ghazal (approved):
LGTM

PR #2084, opened 2026-07-17
test(inventory): use shared onboarding helpers in the v2 suite

p-targowicz:
Part of the one-file-per-PR python-to-Go test port series (see #2079). The python tests keep running in parallel. Behavior unchanged; statistics assertions become delta-based with device cleanup.

### DIFF
--- a/backend/tests/runner/tests/opensource/inventory_management_v2_test.go
+++ b/backend/tests/runner/tests/opensource/inventory_management_v2_test.go
@@ -2,26 +2,11 @@ package opensource
 
 import (
 	"context"
-	"crypto"
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/sha256"
-	"crypto/x509"
-	"encoding/base64"
-	"encoding/json"
-	"encoding/pem"
-	"fmt"
-	"net"
 	"net/http"
-	"slices"
 	"time"
 
 	"github.com/mendersoftware/mender-server/pkg/api/client"
-	"github.com/mendersoftware/mender-server/pkg/utils/types"
-	"github.com/mendersoftware/mender-server/services/deviceauth/model"
-	modelinventory "github.com/mendersoftware/mender-server/services/inventory/model"
 	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
-	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


backend/tests/runner/tests/opensource/inventory_management_v2_test.go
```diff
@@ -53,221 +40,97 @@ func (u *InventoryManagementV2Suite) SetupSuite() {
...
-		mac, err := u.randomMAC()
-		require.NoError(err, "failed to generate random mac")
-		macs = append(macs, mac.String())
-	}
-
-	// Create auth requests for all macs
-	for _, m := range macs {
-		_, _, err := u.authRequest(ctx, u.Tenant.TenantToken, m)
-		require.NoError(err)
+func (u *InventoryManagementV2Suite) TearDownSuite() {
```
bahaa-ghazal: This isn't really needed because every suite gets their own tenant, which is isolated from any other tenant. Therefore, there is no need to use time to clean things up after the suite is already done.

backend/tests/runner/tests/opensource/inventory_management_v2_test.go
```diff
@@ -35,6 +20,8 @@ type InventoryManagementV2Suite struct {
 	Tenant    common.Tenant
 
 	JWT string
+
+	createdDevices []string
```
bahaa-ghazal: Is this needed because of other test that  you are migrating?

bahaa-ghazal (commented):
Looks good 👍
Just couple of comments regarding the suite refactoring

backend/tests/runner/tests/opensource/inventory_management_v2_test.go
p-targowicz: yep

backend/tests/runner/tests/opensource/inventory_management_v2_test.go
p-targowicz: true done

bahaa-ghazal (dismissed):
Still think 125cadc34c5cec289125c34bc13a5783d77e6258, is redundant but the rest looks good

bahaa-ghazal (approved):
Thank you!

PR #2088, opened 2026-07-17
few fixes to prevent deviceauth from sending Internal Server Errors on token verification

kjaskiewiczz:
There are few issues in the deviceauth:
- it's possible to accept device without provisioning;
- it's possible to provision same device multiple times;
- when the device has more than one auth set and one of them is a preauth, when removing preauth, we remove the device leaving orphan auth sets behind.
This PR addresses only the last issue.
For the first two issues I'll create separate PR with workaround and I also created bug report, so we'll have opportunity to fix it properly - https://northerntech.atlassian.net/browse/MEN-9993 

I divided fix into smaller commits.
First one moves code responsible for deleting auth sets and removing preauth device into separate methods. First method is also used when verifying token to remove orphan authset.
Second prevents device that has more auth sets than single preauth from being deleted.
Last one adds a bit of cleanup when, during token verification, we see auth set without corresponding device - in this case we remove the auth set.

After spending more time with deviceauth code and tests I also created this one - https://northerntech.atlassian.net/browse/MEN-9997
The fix included in this PR (https://github.com/mendersoftware/mender-server/pull/2088/changes/13f55baf2c2e2ff297dff751fbc9477e3a9dc3bb) is more of a workaround.

### DIFF
--- a/backend/services/deviceauth/devauth/devauth.go
+++ b/backend/services/deviceauth/devauth/devauth.go
@@ -723,12 +723,6 @@ func (d *DevAuth) DeleteDevice(ctx context.Context, devID string) error {
 
 // Deletes device authentication set, and optionally the device.
 func (d *DevAuth) DeleteAuthSet(ctx context.Context, devID string, authId string) error {
-	l := log.FromContext(ctx)
-
-	l.Warnf("Delete authentication set with id: "+
-		"%s for the device with id: %s",
-		authId, devID)
-
 	// retrieve device authentication set to check its status
 	authSet, err := d.db.GetAuthSetById(ctx, authId)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


backend/services/deviceauth/devauth/devauth.go
```diff
@@ -1230,6 +1230,15 @@ func (d *DevAuth) VerifyToken(ctx context.Context, raw string) error {
 	// decommissioning
 	dev, err := d.db.GetDeviceById(ctx, auth.DeviceId)
 	if err != nil {
+		if err == store.ErrDevNotFound {
```
alfrunes, line 1233: Maybe use `errors.Is` in case we get the idea to wrap the error?

backend/services/deviceauth/devauth/devauth.go
alfrunes, line 476: The branches here are a bit unnecessarily complex (there are multiple branches on `err` and different status variables. Could we simplify it a little?
For example:
```go
	newStatus, err := d.db.GetDeviceStatus(ctx, devId)
	switch err {
	case nil:
		if currentStatus == newStatus {
			return nil
		}
		status = newStatus
	case store.ErrAuthSetNotFound:
		status = model.DevStatusNoAuth
	default:
		return errors.Wrap(err, "Cannot determine device status")
	}
```
or are were we intentionally ignoring the error if `status` argument is empty?

kjaskiewiczz: Yeah, I think I tried to modify this part three times already, and every time something was broken, so I ended up adding the smallest change I could and creating follow up ticket - https://northerntech.atlassian.net/browse/MEN-9997

We are not only ignoring error if status is empty, but also newStatus, unless it's the same as currentStatus.
TBH I don't know what are we trying to achieve here, but I left the investigation to the person that will pick MEN-9997.

alfrunes: So it's a feature, not a bug :sparkles:

alfrunes (approved):

PR #2105, opened 2026-07-27
fix(gui): terminal replay library migration adjustments

mineralsfree:
failing job: https://gitlab.com/Northern.tech/Mender/mender-server/-/jobs/15554339392

### DIFF
--- a/frontend/src/js/components/auditlogs/eventdetails/TerminalPlayer.tsx
+++ b/frontend/src/js/components/auditlogs/eventdetails/TerminalPlayer.tsx
@@ -88,8 +88,7 @@ const generateHtml = (versions, content) => {
         <button id="stop" class="disabled" disabled onclick="handleStop()">Stop</button>
       </div>
       <script>
-        const byteArrayToString = body => String.fromCharCode(...body);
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/auditlogs/eventdetails/TerminalPlayer.tsx
```diff
@@ -89,7 +89,7 @@ const generateHtml = (versions, content) => {
       </div>
       <script>
         const byteArrayToString = body => String.fromCharCode(...body);
-        const transfer = '${JSON.stringify(content.map(item => ({ delay: item.delay, content: btoa(JSON.stringify(item.content)) })))}';
+        const transfer = '${JSON.stringify(content.map(item => ({ delay: item.delay, content: btoa(JSON.stringify(item.content ? Array.from(item.content) : item.content)) })))}';
```
mzedel: based on the commit message I would expect 
```suggestion
        const transfer = '${JSON.stringify(content.map(item => ({ delay: item.delay, content: btoa(JSON.stringify(byteArrayToString(item.content))) })))}';
```
to be more in line with the processing of incoming messages... and avoid stringifying the non existent `item.content`

mzedel (approved):

PR #2115, opened 2026-07-30
MEN-9586 - fix(gui): let deployment creation options reflect inability to autogenerate delta deployments for trial users

mzedel:


mineralsfree (approved):
👍 😄

frontend/src/js/components/deployments/CreateDeployment.tsx
```diff
@@ -300,7 +287,7 @@ export const CreateDeployment = ({ deploymentObject = {}, onDismiss, onScheduleS
               <RolloutPatternSelection isEnterprise={isEnterprise} previousPhases={previousPhases} />
               <RolloutOptions isEnterprise={isEnterprise} />
               <ForceDeploy />
-              {hasDeltaEnabled && (
+              {!isTrial && hasDeltaEnabled && (
```
mineralsfree, line 290: or maybe disable the input instead? To make feature visible?

mzedel: yep, but decided against it as the inline notifications have been increasingly removed - it remains visible and disabled in the settings...

PR #2122, opened 2026-07-31
Men 9320: Deployment report theme adjustments

mineralsfree:
After: 
1.image
2. 
image
3.
image

### DIFF
--- a/frontend/src/js/components/deployments/deployment-report/DeviceList.tsx
+++ b/frontend/src/js/components/deployments/deployment-report/DeviceList.tsx
@@ -22,7 +22,7 @@ import {
   Error as ErrorIcon,
   PendingOutlined as PendingIcon
 } from '@mui/icons-material';
-import type { LinearProgressProps, SvgIconOwnProps } from '@mui/material';
+import type { SvgIconOwnProps } from '@mui/material';
 import { Button, LinearProgress, Typography } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 
@@ -33,12 +33,12 @@ import { Link } from '@northern.tech/common-ui/Link';
 import Loader from '@northern.tech/common-ui/Loader';
 import Pagination from '@northern.tech/common-ui/Pagination';
 import { MaybeTime } from '@northern.tech/common-ui/Time';
-import { SynchronizedTwoColumnData } from '@northern.tech/common-ui/TwoColumnData';
+import { TwoColumnData } from '@northern.tech/common-ui/TwoColumnData';
 import MenderTooltip from '@northern.tech/common-ui/helptips/MenderTooltip';
 import {
   DEVICE_LIST_DEFAULTS,
   canAccess as canShow,
-  deploymentSubstates,
+  deploymentStatesToSubstates,
   rootfsImageVersion as rootfsImageVersionAttribute
 } from '@northern.tech/store/constants';
 import { generateReleasesPath } from '@northern.tech/store/locationutils';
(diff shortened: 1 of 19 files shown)


chiachenglu:
Thank you, just some small things I noticed.

1. It should be sentence case for the alert message 
Screenshot 2026-08-03 at 08 11 34

2. Device status "Already installed" should have "Skipped" in the first line. Some of the statuses are missing/not defined in the [design file](https://www.figma.com/design/vewvWAMm9vPomw4YZOHiGt/Follow-up-after-design-theme-update?node-id=10322-45097&t=YVxbhnMPVFVvEFOO-4), e.g., "Pause before committing" and "Decommissioned". @michaelntech Can you help to clarify them? 

3. Remove the arrow next to the "End time" at schedule detail section. In the design, it seems to be moved to closer to the "Start time" too. 
Screenshot 2026-08-03 at 08 34 53

Thank you :) @mineralsfree

mineralsfree:
> Thank you, just some small things I noticed.
> 
> 1. It should be sentence case for the alert message
> 
> Screenshot 2026-08-03 at 08 11 34
> 2. Device status "Already installed" should have "Skipped" in the first line. Some of the statuses are missing/not defined in the [design file](https://www.figma.com/design/vewvWAMm9vPomw4YZOHiGt/Follow-up-after-design-theme-update?node-id=10322-45097&t=YVxbhnMPVFVvEFOO-4), e.g., "Pause before committing" and "Decommissioned". @michaelntech Can you help to clarify them?
> 3. Remove the arrow next to the "End time" at schedule detail section. In the design, it seems to be moved to closer to the "Start time" too.
> 
> Screenshot 2026-08-03 at 08 34 53
> Thank you :) @mineralsfree

Thanks for taking a look. 
1. Fixed
2. This one is a bit more difficult than it might seem. We don't always recieve substate from server (the second string under the status) So we cannot just replace `Already installed` with `Skipped`, because than we will loose context. Maybe this change can be done as part of upcoming report improvement where all the other states will be clarified and reworked?
3. I've mentioned in the current ticket that I am not touching phases, as it's going to be replaced with uniform phase introduction anyway: https://northerntech.atlassian.net/browse/MEN-9099

chiachenglu:
@mineralsfree I see, for the second part, do you mean in the UI second line of the device status cell can't always been received by the frond end? like those? 
Screenshot 2026-08-03 at 10 34 05
Screenshot 2026-08-03 at 10 28 16

mineralsfree:
> @mineralsfree I see, for the second part, do you mean in the UI second line of the device status cell can't always been received by the frond end? like those? Screenshot 2026-08-03 at 10 34 05 Screenshot 2026-08-03 at 10 28 16

yes

frontend/src/js/components/deployments/deployment-report/DeviceList.tsx
```diff
@@ -189,19 +184,11 @@ const deviceListColumns = [
       return (
         <>
           <div className="flexbox align-items-center margin-bottom-x-small">
```
mzedel:
```suggestion
          <div className="flexbox align-items-center">
```
to ensure the icon is center aligned with the rest of the line if there's no substate info - the substate or the progress will need some margin added instead

frontend/src/js/components/deployments/deployment-report/Log.tsx
```diff
@@ -47,10 +49,10 @@ export const LogDialog = ({ canAi, deployment, deviceId, onClose }) => {
   const exportLog = () => createFileDownload(logData, getFilename(context), '');
 
   return (
-    <BaseDialog open title="Deployment log for device" maxWidth="xl" onClose={onClose}>
+    <BaseDialog open title="Deployment log for device" maxWidth="lg" onClose={onClose}>
       <DialogContent className={`${classes.wrapper} ${canAi ? 'ai-enabled' : ''}`}>
         <div className={classes.codeContainer}>
-          <CopyCode code={logData} withDescription />
+          <Code style={{ whiteSpace: 'pre-wrap', wordBreak: 'break-word' }}>{logData}</Code>
```
mzedel: I assume you could also do:
```suggestion
          <Code className="copyable-content">{logData}</Code>
```
as the code is copyable after all (even though it's via the extracted button now) 🤷

mzedel (dismissed):
Thank you & well done 👍 - 2 tiny suggestions, but since the icon alignment has been there for a while I don't think this should prevent a merge

mzedel (approved):

chiachenglu:
@michaelntech We might be a little late here for the deployment status icons discussion, can you include this in the in the deployment page rework task? [https://northerntech.atlassian.net/browse/MEN-6754] so we have make sure we include this in the rework?

PR #2134, opened 2026-08-04
feat(deviceauth): add feature flag to restore legacy provision behavior

alfrunes:
70796e33252db9c085d3a19adbff5dd20ef52da9 had a side effect of changing the iot-manager webhook events to no longer trigger provision_device event more than once. This commit introduces a feature flag to restore this legacy behavior.

Ticket: MEN-10027

### DIFF
--- a/backend/services/deviceauth/devauth/devauth.go
+++ b/backend/services/deviceauth/devauth/devauth.go
@@ -135,6 +135,8 @@ type Config struct {
 	DefaultTenantToken string
 
 	HaveAddons bool
+
+	LegacyProvisionEvent bool
 }
 
 func NewDevAuth(d store.DataStore, co client.WorkflowsOtherAPI,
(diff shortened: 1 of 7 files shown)


mzedel (commented):
Thanks for the prompt adjustment! AFAICT the change looks good!

backend/services/deviceauth/devauth/devauth.go
```diff
@@ -871,7 +883,18 @@ func (d *DevAuth) AcceptDeviceAuth(ctx context.Context, device_id string, auth_i
 		return err
 	}
 
-	if dev.Provisioned {
+	workflowName := "provision_device"
+	if d.config.LegacyProvisionEvent {
```
merlin-northern: since some tenants may rely on this but I guess very very not many and some may rely on the opposite, I would consider creating per-tenant flag (I know). I would also check perhaps somehow how many of the events we process on hosted -- perhaps it is not such a big deal.

alfrunes: Something like this? https://github.com/mendersoftware/mender-server/pull/2134/changes/d21879e872bca4555369353644acefd7b6860d0b

kjaskiewiczz (approved):

PR #2135, opened 2026-08-04
fix(create-artifact-worker): pin single-file-artifact-gen to a release tag

p-targowicz:
Closes QA-345

single-file-artifact-gen was fetched from the mender client repo at master so any rebuild picked up whatever was on the branch head

Now it is pinned to MENDER_CLIENT_VERSION=5.1.0 and renovate tracks that ARG

I checked that the renovate config validates the image builds and artifact generation works inside it

One thing for the reviewer - 5.1.0 does not have master's change from -t to --compatible-types and mender-artifact 4.4.1 already marks --device-type as deprecated so this will need a version bump before that flag goes away

### DIFF
--- a/backend/services/create-artifact-worker/Dockerfile
+++ b/backend/services/create-artifact-worker/Dockerfile
@@ -48,6 +48,7 @@ FROM ${REGISTRY}/library/alpine:3.24.1
 ARG TARGETARCH
 ARG TARGETOS
 ARG USER=65534:65534
+ARG MENDER_CLIENT_VERSION=5.1.0
 RUN apk add libc6-compat xz bash
 USER $USER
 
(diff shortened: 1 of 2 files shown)


p-targowicz:
Closing this on purpose after a second look

Pinning single-file-artifact-gen to 5.1.0 would introduce a risk that floating on master does not have - renovate bumps MENDER_ARTIFACT_VERSION on its own so mender-artifact can move ahead while the script stays put which is exactly the skew that mendersoftware/mender@f120bb497 was written to fix

That commit rewrites -t to --compatible-types and no release tag carries it yet - only master and qnx-preview-v1 - so pinning to 5.1.0 would freeze us on the flag mender-artifact 4.4.1 already reports as deprecated

Leaving it on master until a release carries the fix. The enterprise only half is still going ahead in mendersoftware/mender-server-enterprise#1553 where the pinned version is identical to the branch head so there is no such tradeoff

PR #2138, opened 2026-08-04
MEN-9997: Fix inconsistencies in deviceauth status handlers

alfrunes:
Fixes inconsistencies and removes duplicated lookups and adds race protection for status updates.
It also expands the webhook context for status update events.
Implementation note:
* I have purposely avoid changing the inputs of the `provision_device` workflow to avoid breaking changes / creating a new workflow.
* The inventory "Provision Device" internal API is now unused and can be removed in a future version in favor of the internal status update API.
* provision_device workflow is now a superset of update_device_status workflow and now triggers both provision and status update event in iot-manager (therefore only one of the workflow is called on status changes).

Please follow the commits while reviewing as they also provide more context.


kjaskiewiczz (approved):
looks good

PR #2143, opened 2026-08-07
MEN-8241: app bar (header) adjustments

mineralsfree:


frontend/src/js/components/header/Header.tsx
```diff
@@ -384,29 +375,27 @@ export const Header = ({ isDarkMode }) => {
...
-          <>
-            <div className="flexbox align-items-center">
-              <Chip label="Service Provider" />
-              <Divider className={`margin-left-small margin-right-small ${classes.headerSection}`} orientation="vertical" />
-              <AccountMenu className={classes.headerSection} />
-            </div>
-          </>
+          <div className="flexbox align-items-center">
+            <Chip label="Service Provider" />
+            <Divider className={`margin-left-small margin-right-small ${classes.headerSection}`} orientation="vertical" />
```
mzedel:
```suggestion
            <Divider className={`margin-left-x-small margin-right-x-small ${classes.divider}`} orientation="vertical" />
```
and then the `classes.headerSection` is only applicable to the `AccountMenu` - which I _assume_ should be ok with the default text colours for the menu items (+ what's defined for the actual text via the
```
<Typography variant="subtitle2" className="text-overflow">
  {email}
</Typography>
```
section...

frontend/src/js/components/header/DeploymentNotifications.tsx
```diff
@@ -11,28 +11,20 @@
...
-  const { classes } = useStyles();
-  return (
-    <Link to={DEPLOYMENT_ROUTES.active.route} className={`flexbox align-items-center ${classes.root} ${className}`}>
-      <RefreshIcon className="flip-horizontal margin-right-x-small" fontSize="small" />
-      <Typography variant="subtitle2">{inprogress}</Typography>
-    </Link>
-  );
-};
+const DeploymentNotifications = ({ className = '', inprogress }) => (
+  <Tooltip title="Deployments in progress">
```
mzedel:
```suggestion
  <Tooltip arrow title="Deployments in progress">
```

mzedel (changes requested):
there's some slight misalignment left between the designed menuitem heights & the hover state on the tenant switch and I guess we could remove the `headerSection` class, but apart from that 👍

### here:

Image

### design:

Image

frontend/src/js/components/header/Header.tsx
```diff
@@ -199,57 +202,62 @@ const AccountMenu = ({ className }) => {
...
         open={Boolean(anchorEl)}
         anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
         transformOrigin={{ horizontal: 'right', vertical: 'top' }}
+        slotProps={{ paper: { className: classes.menuPaper } }}
       >
         <MenuItem component={RouterLink} to="/settings/my-profile" onClick={handleClose}>
           My profile
         </MenuItem>
-        <Divider />
+        <Divider className={classes.menuDivider} />
```
mzedel: only now I've checked the docs for the menuitem & noticed:
```
        <MenuItem component={RouterLink} to="/settings/my-profile" onClick={handleClose} divider>
          My profile
        </MenuItem>
```
to drop the need for the `<Divider />` altogether + get the expected `margin` for the use case... the _My organization_ item below would then also get the divider, the one above the tenant switch would move below the switch (and not have a margin there) - all to drop the `menuDivider` class 🙃

frontend/src/js/components/header/Header.tsx
```diff
@@ -199,57 +202,62 @@ const AccountMenu = ({ className }) => {
...
         {!!(multitenancy && name) && (
-          <MenuItem component={RouterLink} dense to="/settings/organization" onClick={handleClose} className={classes.organization}>
-            <div>
-              <Typography variant="caption" className="muted">
+          <MenuItem component={RouterLink} dense to="/settings/organization" onClick={handleClose}>
+            <div className={classes.organization}>
```
mzedel:
```suggestion
            <div className="text-overflow">
```

frontend/src/js/components/header/Header.tsx
```diff
@@ -105,69 +108,70 @@ const cookies = new Cookies();
...
     border: 'none',
-    '&:hover': {
-      border: 'none'
-    }
+    [`& .${accordionSummaryClasses.root}`]: {
+      minHeight: 'unset',
+      padding: theme.spacing(0.75, 2),
+      '&:hover': { backgroundColor: theme.palette.action.hover }
+    },
+    [`& .${accordionSummaryClasses.content}`]: { margin: 0 }
```
mzedel: instead of this you could also have a `slotProps={{ content: { className: 'margin-none' } }} on the `AccordionSummary`

mzedel (dismissed):
looks good, I'd prefer to remove the boilerplate to get there a bit, but that shouldn't stand in the way of closing this IMO

mzedel (approved):

PR #2146, opened 2026-08-10
Qa 1702/testutils docker compose fixes

rewanrashid-boop:


backend/tests/testutils/infra/container_manager/docker_manager.py
```diff
@@ -49,14 +52,30 @@ def _cp(self, source, destination):
...
+        Matched on the compose labels rather than by grepping `docker ps` output.
+        A substring match also hits any project whose name merely *contains* this
+        one -- "mender" against "mender_enterprise", or a primary namespace
+        against its own failover backend -- and returns several ids on one line.
+        """
+        cmd = (
+            "docker ps -q "
+            "--filter label=com.docker.compose.project={project} "
+            "--filter label=com.docker.compose.service={service}"
+        ).format(project=self.name, service=service)
```
alfrunes: What about using `docker compose` CLI?
```suggestion
        cmd = ["docker", "compose", "-p", project, "ps", "-q", service]
```

backend/tests/testutils/util/websockets.py
```diff
@@ -25,11 +26,21 @@
...
         self.insecure = insecure
         self.retry_connect = retry_connect
+        # Optional (host, port) to open the TCP connection to, instead of the
+        # host in 'url'. Needed because Traefik routes on the Host header while
+        # tests can only reach it by container IP: putting the IP in the URL
+        # would send the wrong Host, and passing Host via additional_headers
+        # emits it twice (the library already derives one from the URL).
+        # Connecting a socket ourselves keeps the URL -- and therefore both the
+        # Host header and the TLS SNI name -- canonical.
+        self.connect_to = connect_to
```
alfrunes, line 43: Arguably something seems wrong with the test environment if you cannot resolve traefik by the right host alias.

rewanrashid-boop:
### Backround and Info
These changes are upstream fixes in mender-server as part of the making the future pip package used for integration in all client repos the standard way to test both locally and in CI. The current prototype of the package can be found here: https://github.com/mendersoftware/mender-testkit

These tasks are all part of the epic https://northerntech.atlassian.net/browse/QA-1390 which is exactly this. Current situation is that in the client repos I have made open PRs for migrating each test suite to use compose files in server. All suites are functional and have been tested against both the sumbodule and now recently the new package in mender-testkit. The goal is to sever this submodule and rely solely on the package but currently the pakcage points to the sha of this pr because main does not have the recent changes.  I think however we should point the package at a release tag instead so that in the next release of the server the package will have these changes. Once this is ok to merge for you guys, I will update the ref to point to main and later when the next release of mender-server is ready will switch renovate.json to be tag release based. 

Any input on this is appreciated and if this sounds good to you let me know then I'll create follow up task in the epic specifying the renovate changes.  ദ്ദി(˵ •̀ ᴗ - ˵ ) ✧

backend/tests/testutils/util/websockets.py
rewanrashid-boop: It isn't an environment problem: these suites  run several backends at once where integration gives each test namespace its own compose project  so there's no single Traefik for a hostname to point at. docker.mender.io names one address, since only one stack can hold the port. Picking a specific stack out of N means using its container IP.

As for the alias, itself it is fine and does resolve, just not from where the tests run: 127.0.0.11 is namespace-local, so the host can't query it. But even with an /etc/hosts entry it still wouldn't disambiguate the stacks.
Once you're on an IP, Traefik still routes on the Host header, so the URL has to stay canonical while the TCP connection goes elsewhere that's all connect_to does. Putting the IP in the URL sends the wrong Host; passing Host via additional_headers sends it twice, since websockets already derives one from the URL.

alfrunes (approved):

PR #2152, opened 2026-08-12
fix(deviceauth): check limit only if device status transitions to acc…

alfrunes:
…epted

This commit refactors the limit check outside of the updateAuthSetStatus since the check depends on the current status of the device (not authset).


kjaskiewiczz (approved):
looks good

PR #2154, opened 2026-08-13
Men 9328: PAT creation dialog adjustments

mineralsfree:


mineralsfree:
@mender-test-bot start review app enterprise

mineralsfree:

frontend/src/js/components/settings/AccessTokenManagement.tsx
```diff
@@ -86,120 +76,124 @@ const columnData = [
...
+                render={({ field: { onChange, value } }) => (
+                  <Select labelId="token-expiration-label" label="Expiration" disabled={!!token} onChange={onChange} value={value}>
+                    {Object.entries(expirationTimes).map(([title, value]) => (
+                      <MenuItem key={value} value={value}>
+                        {title}
+                      </MenuItem>
+                    ))}
+                  </Select>
+                )}
+              />
```
mzedel: Considering that the `expirationTimes` also needs to be reverted to the prior setup, I'm not sure if that really saves that much, but maybe gain us some more consistency 🤷
```suggestion
              <ControlledSelect
                name="expiresIn"
                disabled={!!token}
                labelAttribute="title"
                labelId="token-expiration-label"
                options={expirationTimes}
                selectionAttribute="value"
              />
```

mzedel (dismissed):

frontend/src/js/common-ui/forms/ControlledSelect.tsx
```diff
@@ -11,9 +11,27 @@
...
+  name: string;
+  options?: Option[];
+  placeholder?: string;
+  renderOption?: (option: Option) => ReactNode;
+  selectionAttribute?: string;
+
+  width?: number | string;
+};
+
+export const objToOptions = (obj: Option): Option[] => Object.entries(obj).map(([title, id]) => ({ id, title }));
```
mzedel: This + the `type Option = Record<string, string | number>;` seems to be very tailored to the current usage, in contrast to the mui `any` for value (https://mui.com/material-ui/api/select/#select-prop-value) IOW `type Option = Record<string, any>;` 😕... I guess for the type that's ok, but I can't see the `objToOptions` age well for any `object` that's not a pure map/ as a util along the `ControlledSelect` vs. the usage in the PAT dialog

mzedel (dismissed):

frontend/src/js/common-ui/forms/ControlledSelect.tsx
mineralsfree: Good point. moved helper to component, relaxed type

mzedel (approved):

PR #2155, opened 2026-08-13
fix(gui): change wrong UI status for aborted deployments

LudvigAnderson:
Ticket: MEN-9983
Changelog: Title


LudvigAnderson:
@mineralsfree This is part of something Michael found and shared in Slack a while ago. I got reminded of it because it showed up today in a support ticket :smile: 

Do you think there is a need for tests for this, maybe? Ideally, a test could have caught that the deployment status label was very different from the actual deployment status. What do you think?

mineralsfree (approved):
Nice catch 👍

mineralsfree:
> Do you think there is a need for tests for this, maybe? Ideally, a test could have caught that the deployment status label was very different from the actual deployment status. What do you think?

I see your point, but I don't think it's common to test things like this (mapping between keys and title, icons, etc.) Usually things like this are caught with snapshot testing and PR review, but here the changes to the report that brought the issue where too broad and it was easy to miss it

PR #2156, opened 2026-08-13
ci: added temporary workaround to allow gitlab registry image publishing w/ new docker

mzedel:


merlin-northern (approved):
wow. that is some advanced gitlab magic.

PR #2160, opened 2026-08-17
MEN-9976: Add API spec for search inventory by identity

frodeha:
**Description**
Added API specification for the new `search inventory by identity` endpoint as per the ticket description.

Ticket: MEN-9976


backend/docs/api/dist/openapi.yaml
```diff
@@ -5839,6 +5841,72 @@ paths:
           $ref: '#/components/responses/InternalServerError'
       security:
         - ManagementJWT: []
+  /api/management/v2alpha1/inventory/identities/search:
```
mzedel, line 5844: considering the presence of the deviceauth service that so far got "identity" as scope, this feels slightly wrong (I guess as the `tags` scope is covered too here), but OTOH the `alpha1` nature & the upcoming changes that shouldn't cause too much confusion 👌

mineralsfree (dismissed):
lgtm

mzedel (commented):
while the design doesn't seem to show this, I'm quite sure we are expected to list devices with their selected device identity attribute, so I'd say the `List of attributes to select and return` in the response would be helpful here as well...
OR: ensure the searched for attribute is always included (the examples imply that's not the case)

backend/docs/api/dist/openapi.yaml
frodeha: Yeah, I'm not in love with the name either I have to admit (I'm sort of half inventing a new thing, but not really). My challenge is that the semantics of it doesn't map cleanly, e.g you can search for device id (which is not even an attribute and technically speaking has no scope), the one specific tags scoped attribute "name" (but no others) and all identity scoped attributes. That's a tricky resource to define if you ask me (especially in a single term).

Hmm, what about some alternatives:
- /api/management/v2alpha1/inventory/id-attributes/search (inspired by the key in the tenant settings object)
- /api/management/v2alpha1/inventory/device-identity-attributes/search (inspired by the UI text itself)

Do you find any of these preferable? Or maybe you've got something better for me...? 😇

frodeha:
> while the design doesn't seem to show this, I'm quite sure we are expected to list devices with their selected device identity attribute, so I'd say the `List of attributes to select and return` in the response would be helpful here as well...

Oh I was just gonna return all attributes 😅 That being said, I do like the idea of adding the `List of attributes to select and return` so I'll add that instead. 

> OR: ensure the searched for attribute is always included (the examples imply that's not the case)

Yeah, thanks for pointing this out. The example is misleading here. What you searched for is returned, but I'll tweak it to be the `List of attributes to select and return` instead.

backend/docs/api/dist/openapi.yaml
mzedel: the `identities` is fine & as long as there's the `alpha1` I would be happy with `/bacon/search` - I'm holding out for `/v1/devices/search` 🤗

frodeha: > I would be happy with /bacon/search

Great. Now I'm hungry... :joy:

mzedel (dismissed):

mzedel (approved):

PR #2161, opened 2026-08-17
fix(deviceconnect): proper graceful shutdown with ws handshake

alfrunes:
This commit makes the server perform a proper close handshake on graceful shutdown when receiving SIGTERM, SIGQUIT and SIGUSR1. The latter also adds a configurable grace period to rate limit the shutdown period. The internal shutdown API calls SIGUSR1 to trigger shutdown and properly waits for the shutdown handler to finish before returning. Repeated SIGQUIT and SIGTERM will abort the graceful shutdown and stop the server immediately.


backend/services/deviceconnect/api/http/device.go
```diff
@@ -175,12 +175,12 @@ func (h DeviceController) Connect(c *gin.Context) {
 	logCtx := accesslog.GetContext(ctx)
 	logCtx.SetField("status", 101)
 	conn.SetReadLimit(int64(app.MessageSizeLimit))
+	shutdown, stop := h.app.GetShutdownNotification()
```
kjaskiewiczz, line 178: It's possible I'm a bit lost, but with RegisterConnectionCancelHandle we had exclusive parameter with which the method was making sure there is only one connection per device. Is it still the case?

backend/services/deviceconnect/server/server.go
```diff
@@ -78,30 +78,43 @@ func InitAndRun(conf config.Reader, dataStore store.DataStore) error {
...
-	l.Info("server exited")
-
-	if recvSignal == unix.SIGUSR1 {
-		deviceConnectApp.ShutdownDone()
-		l.Info("graceful shutdown completed")
+	select {
+	case <-deviceConnectApp.Done():
+		// Give the last websocket sessions a second to complete the handshake
+		select {
+		case err = <-errChan:
```
kjaskiewiczz, line 111: I think this one will happen immediately (would need a bit more time to test it to be sure though) and we'll never give the last websocket session a second.

backend/services/deviceconnect/api/http/status.go
```diff
@@ -70,16 +69,20 @@ func (h StatusController) Health(c *gin.Context) {
...
 		c.JSON(http.StatusServiceUnavailable, gin.H{
 			"error": err.Error(),
 		})
 		return
 	}
-	time.Sleep(h.gracefulShutdownTimeout)
-	c.Writer.WriteHeader(http.StatusAccepted)
+	select {
+	case <-h.app.Done():
+		c.Writer.WriteHeader(http.StatusOK)
```
kjaskiewiczz, line 84: we say it returns 202 here - https://github.com/mendersoftware/mender-server/blob/24d19f2df9f4c0a37df9dbdb60de3b0eb224d314/backend/docs/api/deviceconnect/internal_v1.yaml#L44, but with this change it will return 200

backend/services/deviceconnect/server/server.go
```diff
@@ -78,30 +78,43 @@ func InitAndRun(conf config.Reader, dataStore store.DataStore) error {
...
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			err = fmt.Errorf("timeout waiting for server to shutdown: %w", err)
```
kjaskiewiczz, line 114: not a big deal, but I think err here will always be nil

backend/services/deviceconnect/api/http/router.go
kjaskiewiczz, line 66: do we still use it?

backend/services/deviceconnect/api/http/management.go
```diff
@@ -217,6 +217,8 @@ func (h ManagementController) Connect(c *gin.Context) {
 	logCtx := accesslog.GetContext(ctx)
 	logCtx.SetField("status", http.StatusSwitchingProtocols)
 
+	done, stop := h.app.GetShutdownNotification()
```
kjaskiewiczz, line 220: I think this will reassign stop from line 151

backend/services/deviceconnect/api/http/device.go
```diff
@@ -345,6 +345,9 @@ func (h DeviceController) handleManagementMessages(
 				}
 				err = conn.WriteMessage(websocket.BinaryMessage, data)
 				if err != nil {
+					if errors.Is(err, websocket.ErrCloseSent) {
```
kjaskiewiczz, line 348: should we add the same check in management? https://github.com/mendersoftware/mender-server/blob/24d19f2df9f4c0a37df9dbdb60de3b0eb224d314/backend/services/deviceconnect/api/http/management.go#L464

kjaskiewiczz (commented):
not sure I fully understand it, so it's possible my comments are simply stupid ;)

backend/services/deviceconnect/api/http/device.go
alfrunes: Ohh, I probably did not notice when implementing this. Good you highlighted this, because the previous implementation would not work in a multi-replica setup! I will see if I can create something here...

backend/services/deviceconnect/server/server.go
alfrunes: Good catch! Should be `s/err/ctx.Err()/`

backend/services/deviceconnect/api/http/router.go
alfrunes: Nope :)

backend/services/deviceconnect/api/http/management.go
alfrunes: Since it's creating a new variable it doesn't have any effect (both `stop` are deferred). But indeed a very poor choice of variable name.

backend/services/deviceconnect/server/server.go
alfrunes: The context timeout is 5s + gracePeriod. So it depends if it was hanging during shutdown. But it is waiting for the minimum of the two. The idea of the last select was to give another second if we have it to wait for close handshake from the peers.

backend/services/deviceconnect/api/http/status.go
alfrunes: It used to be async, but now it's synchronous, waiting for the app to terminate.

alfrunes: Well, that is not entirely true now that I review it. The shutdown of hijacked connections happens async, while the standard library will terminate all non-hijacked connections (including the one initiating this request. Will revert to status code 202.

backend/pkg/stream/nats.go
```diff
@@ -422,46 +514,59 @@ func (l *natsListener) Close(ctx context.Context) error {
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case msg := <-l.msgChan:
+			rest, msgType, _ := cutLast(msg.Subject, ".")
+			switch msgType {
+			case msgTypeClose:
+				// Another ListenNATS caller is taking over this address:
+				// close before acknowledging, so the requester is
+				// guaranteed the subject is free once it gets the reply.
+				_ = l.Close(ctx)
```
kjaskiewiczz, line 530: I'm lost here - we call close from accept, there is no locking, so would it be possible to have race condition here - concurrent modifications of openConns if the device will try to reconnect?

backend/pkg/stream/nats.go
```diff
@@ -386,22 +425,75 @@ type natsListener struct {
...
-	}, nil
+	}
+
+	if cfg.closeIfOccupied {
+		nonce := make([]byte, 16)
+		_, _ = rand.Read(nonce)
+		selfNonce := hex.EncodeToString(nonce)
+
+		subject := fmtListenSubject(addr, addr, msgTypeClose)
+		if err := nc.Publish(subject, []byte(selfNonce)); err != nil {
```
kjaskiewiczz, line 467: Shouldn't we use Request here instead of Publish?
Publish sets no reply, so I guess
```
if msg.Reply != "" {
_ = msg.Respond(nil)
}
```
from Accept will never happen.

backend/docs/api/deviceconnect/internal_v1.yaml
```diff
@@ -41,8 +41,8 @@ paths:
       summary: Shutdown the service.
       operationId: DeviceConnect Internal Shutdown
       responses:
-        202:
-          description: The service started the graceful shutdown procedure.
+        '200':
```
kjaskiewiczz, line 44: shouldn't this be 202?

alfrunes: After some testing I realized that the correct status should be 200. The call to `http.Server.Shutdown` does not kill the connection as the state is Active and not Idle. Therefore, it does not return before the application has shut down.

backend/pkg/stream/nats.go
alfrunes: Yeah, I'm also not super confident about this implementation. Perhaps we should use a separate mechanism to ensure mutual exclusion of the websocket connection. Perhaps I leave it out of this PR for now and create a ticket to actually fix this?

backend/pkg/stream/nats.go
alfrunes: Since the newer listener is already is subscribed before sending the close message with the self nonce. All other connections should terminate when seeing this message, and also all messages received before reading it's own nonce should be discarded as "handled by another listener".
Anyway, I'm not 100% confident in this implementation, perhaps it's better to use a different mechanism (versioned subject, a redis lock or something else?) to ensure mutual exclusion to the devices/connect API?

PR #2162, opened 2026-08-17
feat(deviceconnect): add thundering herd rate limit to ws APIs

alfrunes:
A new configuration `websocket_burst_limit` sets the number of burst requests to accept per second to prevent trembling herd resource exhaustion.

Ticket: MEN-10052

### DIFF
--- a/backend/services/deviceconnect/api/http/rate_test.go
+++ b/backend/services/deviceconnect/api/http/rate_test.go
@@ -1,0 +1,43 @@
+// Copyright 2026 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package http
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/time/rate"
+)
+
+func TestWebsocketRatelimitMiddleware(t *testing.T) {
+	router := gin.New()
+	router.Use(websocketRatelimitMiddleware(rate.NewLimiter(rate.Every(rate.InfDuration), 1))).
+		Any("/", func(*gin.Context) {})
+
+	r, _ := http.NewRequest("GET", "/", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, r)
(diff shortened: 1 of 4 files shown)


backend/services/deviceconnect/config.yaml
```diff
@@ -77,3 +77,8 @@
 # Overwrite with environment variable: DEVICECONNECT_REQUEST_SIZE_LIMIT
 
 # request_size_limit: 1048576
+
+# Number of websocket that can be accepted per second for trembling herd protection.
```
frodeha: I might be wrong, but I think the terms is "thundering herd"...? Not that it matters of course, just thought I'd mention it 🤷

backend/services/deviceconnect/api/http/router.go
```diff
@@ -62,6 +66,18 @@ type RouterConfig struct {
 	GracefulShutdownTimeout time.Duration
 	MaxRequestSize          int64
 	MaxFileSize             int64
+
+	StartupConnectionBurst int
```
frodeha: It's technically not just during startup this applies though is it? A sufficiently large burst of connection requests at any time during the runtime of the service would apply right? Should we maybe rename this to reflect so?

Something like this maybe?
```suggestion
	WebsocketConnectionBurst int
```

backend/services/deviceconnect/api/http/router.go
```diff
@@ -62,6 +66,18 @@ type RouterConfig struct {
...
+}
+
+func startupRatelimitMiddleware(rl *rate.Limiter) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		if !rl.Allow() {
+			rest.RenderError(c, http.StatusServiceUnavailable, fmt.Errorf("server is starting up"))
```
frodeha: Same as above, not only applicable on startup I'd say.

frodeha (dismissed):
Implementation looks good to me, just have some minor (and opinionated) comments you can consider. Nice 👍

backend/services/deviceconnect/api/http/router.go
alfrunes: Changed to `WebsocketBurstLimit`  to match config key.

backend/services/deviceconnect/config.yaml
alfrunes: I like the sound of thunder :cloud_with_lightning:

backend/services/deviceconnect/api/http/rate_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/time/rate"
+)
+
+func TestStartupRatelimit(t *testing.T) {
```
frodeha: Sorry to do this, but should we maybe rename the test as well...? 🙈

backend/services/deviceconnect/api/http/router.go
frodeha: On second read (sorry!), should we maybe produce `429` so we are consistent with other rate limit implementations? Same with the message (`"server is starting up"` -> `"too many requests"`)?

frodeha (approved):
Looks great to me! Nicely done 👍

PR #2164, opened 2026-08-17
reduce page rerenders.

mzedel:


mzedel:
@mender-test-bot sync

PR #2171, opened 2026-08-18
test(gui): added e2e tests for test device actions

mineralsfree:
Ticket: MEN-9989


mzedel (approved):
Not sure if we want to have the full duration impact, but maybe a run through a deployment cycle with the expectation that it won't get picked up within a few minutes would get closer to the intended functionality of the test devices? But for the device actions alone this should be enough...

mineralsfree:
> Not sure if we want to have the full duration impact, but maybe a run through a deployment cycle with the expectation that it won't get picked up within a few minutes would get closer to the intended functionality of the test devices? But for the device actions alone this should be enough...

Hmm, I think such a test might be quite unstable on staging. I don't know if there is an alternative place to test it though... Maybe we could discuss it during STM?

PR #2174, opened 2026-08-18
ci: tune static backend check resource usage

mzedel:


alfrunes (approved):

mzedel:
not sure if this is worth the cost: ~10min with this vs ~1min before 😬 - WDYT @alfrunes?

alfrunes:
10m might be a bit long, but since this job is fairly independent, it doesn't matter if it runs a little bit slower as long as it doesn't become a bottleneck.

PR #2178, opened 2026-08-19
chore(gui): removed now unused release tag retrieval & rely on injected commit or tag

mzedel:
I'll follow up in nt-gui too...


alfrunes (approved):

mineralsfree (approved):
Ready for offline use 😄 👍

PR #2183, opened 2026-08-19
Men 9986: Software tags validations

mineralsfree:


frontend/src/js/common-ui/forms/helpers.tsx
```diff
@@ -13,6 +13,16 @@
 //    limitations under the License.
 import { Typography } from '@mui/material';
 
+import validator from 'validator';
+
+export const invalidCharactersError = 'Valid characters are a-z, A-Z, 0-9, ., _ and -';
+
+export const hasValidTagCharacters = (value: string) => validator.isWhitelisted(value, 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-.');
```
mzedel: since we have some error messages & range definitions already - could you maybe move the `getErrorMsg` to the helpers file and use the validation from there?
OTOH looking at the `isAlphanumericLocator` that's used for role name validation and doesn't allow dots - that might either not be aligned with the backend or handled differently there too... but nevertheless I'd say keeping the validations & errors somewhat close would be good.

mzedel (dismissed):
looks good, just consider the adjustment to not spread the validation logic in too many places

frontend/src/js/common-ui/forms/helpers.tsx
mineralsfree: on top of that we have `validateGroupName` defined in `GroupDefiniton`. I hope a more focused effort on the unification will be done as part of newly emerged i3 limits epic. As of now, moved validatons to a separate file

mzedel: yeah that `GroupDefinition` logic is yet another can of worms 😆 and yeah apparently the role name validation is indeed slightly different 🫠
as to the move 👌, we can surely align the validation & error referencing later on...

mzedel (approved):

PR #2185, opened 2026-08-20
MEN-9976: Search inventory by identity endpoint

frodeha:
**Description**
Added the new alpha versioned identity search endpoint as per the ticket description.


merlin-northern (commented):
I am reading this but I have a general question: how come there is no entry for the docs/api in this PR? 🥹

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
```diff
@@ -0,0 +1,289 @@
...
+	"fmt"
+	"net/http"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/mendersoftware/mender-server/pkg/api/client"
+	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
```
merlin-northern, line 14:
```suggestion
	"github.com/stretchr/testify/require"
	"github.com/stretchr/testify/suite"

	"github.com/mendersoftware/mender-server/pkg/api/client"
	"github.com/mendersoftware/mender-server/tests/runner/tests/common"
```
I know, nitpicking. but it is your own fault! you produce impeccable code so I have to grasp at straws.

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
```diff
@@ -0,0 +1,289 @@
...
+	var (
+		ctx           = common.JWTAuthContext(u.T().Context(), u.JWT)
+		deviceCount   = 4
+		devices       = make([]*common.Device, 0, deviceCount)
+		nameTagPrefix = rand.Text()
+	)
+
+	{
+		var (
+			wg      sync.WaitGroup
```
merlin-northern, line 58: what stands behind this wg dance? I mean why do it like that?

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
```diff
@@ -0,0 +1,289 @@
...
+				_, err = device.SubmitAuthRequest(ctx, u.APIClient, u.Tenant.TenantToken)
+				require.NoError(err)
+
+				err = device.Accept(ctx, u.APIClient)
+				require.NoError(err)
+
+				_, err = u.APIClient.DeviceInventoryManagementAPIAPI.
+					AddTags(ctx, device.ID).
+					Tag([]client.Tag{
+						{Name: "name", Value: fmt.Sprintf("%s-%d", nameTagPrefix, idx+1)},
```
merlin-northern, line 77:
```suggestion
						{Name: DbDevAttributesName
, Value: fmt.Sprintf("%s-%d", nameTagPrefix, idx+1)},
```
perhaps?

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	pickAttributeValue := func(attrs []client.AttributeResponse, scope client.Scope, name string) string {
+		for _, a := range attrs {
+			if a.Scope == scope && a.Name == name && a.Value.String != nil {
+				return *a.Value.String
+			}
+		}
+		return ""
+	}
```
merlin-northern, line 99: so all this dance with wg and lots of inline funcs will make it so much more difficult to dlv debug. and I really do run dlv on tests at least once a week 🥹

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
```diff
@@ -0,0 +1,289 @@
...
+			require        = require.New(u.T())
+			expectedDevice = devices[0]
+			valuePrefix    = expectedDevice.ID[:6]
+		)
+
+		results, _, err := u.APIClient.InventoryV2alpha1ManagementAPIAPI.
+			SearchInventoryByIdentity(ctx).
+			SearchIdentityParams(client.SearchIdentityParams{
+				Scope:       string(client.IDENTITY),
+				Name:        "id",
```
merlin-northern, line 112: I am kind of sure we do have a const for id in like all the places ;-)

merlin-northern:
> I am reading this but I have a general question: how come there is no entry for the docs/api in this PR? 🥹

okey just ignore, I know, the docs are in place already.

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -63,8 +63,10 @@ const (
 	urlInternalAttributesNoScope = "/tenants/:tenant_id/device/:device_id/attributes"
 	urlInternalReindex           = "/tenants/:tenant_id/devices/:device_id/reindex"
 	apiUrlManagementV2           = "/api/management/v2/inventory"
+	apiUrlManagementV2Alpha1     = "/api/management/v2alpha1/inventory"
```
merlin-northern, line 66: this first thing I always think of is: RBAC. just make me calm: the access control is under control here, right?

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -932,6 +934,25 @@ func (i *ManagementAPI) FiltersSearchHandler(c *gin.Context) {
...
+	}
+
+	devs, totalCount, err := i.App.SearchDevicesByIdentity(ctx, searchParams)
+	if err != nil {
+		rest.RenderError(c, http.StatusInternalServerError, errors.New("internal error"))
+		return
+	}
+
+	// the response writer will ensure the header name is in Kebab-Pascal-Case
+	c.Writer.Header().Add(hdrTotalCount, strconv.Itoa(totalCount))
```
merlin-northern, line 952: the docs state:
```
Link:
  Standard header used for page navigation ...
```
could we double check that this is the case?

backend/services/inventory/api/http/api_inventory.go
```diff
@@ -1163,6 +1184,28 @@ func parseSearchParams(c *gin.Context) (*model.SearchParams, error) {
 	return &searchParams, nil
 }
 
+func parseSearchIdentityParams(c *gin.Context) (model.SearchIdentityParams, error) {
```
merlin-northern, line 1187: could we double check that:
* no regex injections are possible
* the format of name and attribute is plain text with no funny binary stuff
* we restrict the number of results in page and per page as in: not to specify page MAX_INT+1 and per_page MAX_INT*2
* we clearly document the formats allowed for all the input fields and counts restrictions in the spec in `SearchIdentityParams`

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
 	return devices, int(count), nil
 }
 
+func (db *DataStoreMongo) SearchDevicesByIdentity(
+	ctx context.Context,
+	searchParams model.SearchIdentityParams,
+) ([]model.Device, int, error) {
```
merlin-northern, line 1226: I would vote for returning empty array (`[]model.Device{}` explicit) instead of nil. we have been down this road and we once agreed that it is easier on the caller. perhaps that has changed, but take under consideration.

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
...
+	defer cursor.Close(ctx)
+
+	devices := []model.Device{}
+	if err = cursor.All(ctx, &devices); err != nil {
+		return nil, -1, errors.Wrap(err, "failed to search devices")
+	}
+
+	count, err := c.CountDocuments(ctx, query)
+	if err != nil {
+		return nil, -1, errors.Wrap(err, "failed to search devices")
```
merlin-northern, line 1277: not technically failed to search here, right?

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
...
+
+	cursor, err := c.Find(ctx, query, options)
+	if err != nil {
+		return nil, -1, errors.Wrap(err, "failed to search devices")
+	}
+	defer cursor.Close(ctx)
+
+	devices := []model.Device{}
+	if err = cursor.All(ctx, &devices); err != nil {
+		return nil, -1, errors.Wrap(err, "failed to search devices")
```
merlin-northern, line 1272: also not strictly failure at search. (refering to the error message)

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
...
+
+	if len(searchParams.Attributes) > 0 {
+		projection := bson.M{makeAttrField(DbDevUpdatedTs, model.AttrScopeSystem): 1}
+		for _, a := range searchParams.Attributes {
+			projection[makeAttrField(a.Attribute, string(a.Scope))] = 1
+		}
+		options.SetProjection(projection)
+	}
+
+	cursor, err := c.Find(ctx, query, options)
```
merlin-northern, line 1264: so below we have just All, and here we have just find. do you think we could have some general restriction on the number of results, like 10000 or anything, to have a final bar-rier preventing Alling all the db, in case something slipped through?

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
...
+		// Using the `_id` index gives better performance than using
+		// the `identities` index, so we leverage it if we can
+		query = bson.M{DbDevId: prefixMatcher}
+		sort = bson.D{{Key: DbDevId, Value: 1}}
+	} else {
+		// Search both the `identities` index _and_ the `attributes` array
+		// results in a significant performance increase. See more technical
+		// details in `MEN-9700`
+		field := makeAttrField(searchParams.Name, string(searchParams.Scope), DbDevAttributesValue)
+		query = bson.M{ /*"identities": prefixMatcher,*/ field: prefixMatcher}
```
merlin-northern, line 1247: this comment is on purpose like this in here on this line? looks like a trap ;-)

backend/services/inventory/store/mongo/datastore_mongo.go
```diff
@@ -1219,6 +1220,66 @@ func (db *DataStoreMongo) SearchDevices(
...
+		prefixMatcher = bson.Regex{Pattern: "^" + regexp.QuoteMeta(searchParams.ValuePrefix)}
+	)
+
+	if searchParams.Name == model.AttrNameID {
+		// Using the `_id` index gives better performance than using
+		// the `identities` index, so we leverage it if we can
+		query = bson.M{DbDevId: prefixMatcher}
+		sort = bson.D{{Key: DbDevId, Value: 1}}
+	} else {
+		// Search both the `identities` index _and_ the `attributes` array
```
merlin-northern, line 1243: help me where is the `DbDevIdentitiesName` here?

merlin-northern (commented):
looks lovely, I dropped some comments mainly for my own peace of mind.

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: We return nil in error cases only here afaict, do you mean to return an empty array also in error cases?

backend/services/inventory/api/http/api_inventory.go
frodeha: Well, I'll have to go back and add it after the sync to enterprise, but indeed it needs to be handled (which I will do)

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
frodeha: Fixed in 12afb22f470744fb7f588d09e7fbaa5b16a3ece6

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
frodeha: It creates the devices concurrently instead of sequentially which cuts the runtime of the test in approximately half. I realize that it's not a pattern we use much though so I made it sequential in 12afb22f470744fb7f588d09e7fbaa5b16a3ece6

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
frodeha: Fixed

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
frodeha: As you wish, the function is no longer in-lined (12afb22f470744fb7f588d09e7fbaa5b16a3ece6) - dlv as much as you like

backend/tests/runner/tests/opensource/inventory_management_v2alpha1_test.go
frodeha: Fixed

backend/services/inventory/api/http/api_inventory.go
frodeha: Sure, extended one of the tests to check it (12afb22f470744fb7f588d09e7fbaa5b16a3ece6), do note that since this is a POST request the paging parameters are sent in the body and this (X-Total-Count) is therefore the only relevant header.

backend/services/inventory/api/http/api_inventory.go
frodeha: > no regex injections are possible

I believe  should be covered by `QuoteMeta` [used as part of building the database query](https://github.com/mendersoftware/mender-server/pull/2185/changes#diff-285981e3ba8d06375be4a2140133d7d02b939140c45db6c3bfa1cbf96342611dR1234)

> the format of name and attribute is plain text with no funny binary stuff

I don't see how honestly. We allow _anything_ as attribute name and value (even funny binary stuff) so we're kinda stuck with it unless we also want to restrict on attribute creation (which is definitely outside the scope of this task imo).

> we restrict the number of results in page and per page as in: not to specify page MAX_INT+1 and per_page MAX_INT*2

This is covered by validating per_page [here](https://github.com/mendersoftware/mender-server/pull/2185/changes#diff-68d9b882b18e612f471d3583e17151bad52aa5ae8d2c517cb01afbebf09807b3R163). page is unbounded, but requesting page 100000000 should be of no consequence surely?

> we clearly document the formats allowed for all the input fields and counts restrictions in the spec in SearchIdentityParams

Fixed in 19f4a85db89ec3b07e39b1352a3b252d6fa86daf

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: Fixed

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: Oh it's certainly not on purpose. I was being an idiot and forgot to fix it after rebasing. Thanks for catching it, fixed in aa2f576c677db156c36847341df783dbd7cad5f4

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: It was lost in transmission from your PR to mine. I got my head fixed and recovered it here: aa2f576c677db156c36847341df783dbd7cad5f4

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: Fixed

backend/services/inventory/store/mongo/datastore_mongo.go
frodeha: I'm not sure about this one. The paging options are always set (courtesy of default values in the api package), so reading the entire database shouldn't happen. Now, what you're talking about is "in case something slipped through" (which is fair enough), but I'm hesitant to add stuff like this only in one query (we have nothing like this anywhere else). 

Could we leave it as-is and consider a general practice/pattern as part of other general work maybe?

backend/services/inventory/store/mongo/datastore_mongo.go
merlin-northern: in the error, yes.

backend/services/inventory/store/mongo/datastore_mongo.go
merlin-northern: sure, I just had this vision about protection. you know I practice safe IT.

merlin-northern (approved):
have I mentioned that I do trust you? it is lovely. (both the trust and the patch)

PR #2191, opened 2026-08-21
chore(gui): replace generic redux dispatch with typed appDispatch

mineralsfree:


mzedel (dismissed):
👍 - you think we could also add the blanket `.unwrap()` for awaited thunks now 😜

mineralsfree:
> 👍 - you think we could also add the blanket `.unwrap()` for awaited thunks now 😜

Let's do the release first 😆

mzedel (approved):

PR #2203, opened 2026-08-26
MEN-10046 + MEN-10047 - preauth dialog, kv editor, file upload

mzedel:


frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
```diff
@@ -25,6 +25,10 @@ import { useAppDispatch } from '@northern.tech/store/store';
 import { preauthDevice } from '@northern.tech/store/thunks';
 import { isEmpty } from '@northern.tech/utils/helpers';
 
+const publicKeyEnvelope = /-----BEGIN (?:[A-Z0-9 ]+ )?PUBLIC KEY-----[\s\S]+-----END (?:[A-Z0-9 ]+ )?PUBLIC KEY-----/;
```
mineralsfree: Maybe something like this (I think backend rejects everything but  `BEGIN PUBLIC KEY`) 
```
const publicKeyEnvelope = /^-----BEGIN PUBLIC KEY-----([A-Za-z0-9+/=\s]+)-----END PUBLIC KEY-----$/;

export const isPublicKey: Validator = value => {
  const [, body] = publicKeyEnvelope.exec(value.trim()) ?? [];
  if (!body) {
    return 'Please provide a public key in PEM format, starting with -----BEGIN PUBLIC KEY-----';
  }
  return validator.isBase64(body.replace(/\s/g, '')) || 'This public key appears to be malformed';
};
``` 
and also move to common validations file?

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
```diff
@@ -44,25 +48,45 @@ export const DeviceLimitWarning = ({ acceptedDevices, deviceLimit, hasContactInf
...
-    setErrortext(null);
+    setSubmitError('');
     setJsonIdentity(jsonIdentity);
   }, []);
 
-  const onHandleSubmit = shouldClose => {
+  const onKeyChange = (content?: string) => {
+    setSubmitError('');
+    if (!content) {
+      setKeyError('');
```
mineralsfree: Maybe set an error ("This file is empty") or something similar?

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
mineralsfree: This error is not a string, but object. Supplying a malformed (but looking alright) key will crash the UI still. 
```
Objects are not valid as a React child (found: object with keys {name, message, stack, code})
```

mineralsfree: maybe something like: 

```
  setSubmitError(typeof error === 'string' ? error : (error as Error).message);
```

frontend/src/js/common-ui/forms/FileUpload.tsx
```diff
@@ -11,35 +11,68 @@
...
+import { makeStyles } from 'tss-react/mui';
 
 import storeActions from '@northern.tech/store/actions';
 import { useAppDispatch } from '@northern.tech/store/store';
 
 const { setSnackbar } = storeActions;
 
-export const FileUpload = ({ enableContentReading = true, fileNameSelection, onFileChange, onFileSelect = () => undefined, placeholder, style = {} }) => {
+const useStyles = makeStyles()(theme => ({
+  dropzone: { ['&.dropzone']: { padding: theme.spacing(4, 2) } },
```
mineralsfree: It's 
```
  dropzone: { ['&.dropzone']: { padding: theme.spacing(3, 2) } },
```
in design

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
```diff
@@ -25,72 +25,91 @@ import { useAppDispatch } from '@northern.tech/store/store';
...
-            <>
-              Drag here or <Link>browse</Link> to upload a public key file
-            </>
-          }
-          onFileChange={setPublicKey}
-        />
-        <h4 className="margin-bottom-none margin-top">Identity data</h4>
-        <KeyValueEditor errortext={errortext} onInputChange={convertIdentityToJSON} />
+      <DialogContent>
+        <p className="margin-top-none">Add a device&apos;s authentication set to authorize it automatically as soon as it connects.</p>
```
mineralsfree: maybe use Typography here?

frontend/src/js/common-ui/forms/KeyValueEditor.test.tsx
```diff
@@ -36,8 +35,7 @@ describe('KeyValueEditor Component', () => {
     const { rerender } = render(ui);
     await user.type(screen.getByPlaceholderText(/key/i), 'testKey');
     await user.type(screen.getByPlaceholderText(/value/i), 'testValue');
-    expect(document.querySelector(fabSelector)).not.toBeDisabled();
-    await user.click(document.querySelector(fabSelector));
+    await screen.getByLabelText('add-editor-line-button').click();
```
mineralsfree: why not `user.click(screen.getByLabelText('add-editor-line-button'))` ? Isn't it a recommended way

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
```diff
@@ -29,22 +29,10 @@ const publicKeyEnvelope = /-----BEGIN (?:[A-Z0-9 ]+ )?PUBLIC KEY-----[\s\S]+----
...
-      support portal
-    </Link>{' '}
-    to request a higher limit.
-  </p>
-);
-
-export const DeviceLimitWarning = ({ acceptedDevices, deviceLimit, hasContactInfo }) => (
-  <div className="margin-bottom-small margin-top-small warning">
-    <InfoIcon style={{ marginRight: 2, height: 16, verticalAlign: 'bottom' }} />
+export const DeviceLimitWarning = ({ acceptedDevices, deviceLimit }) => (
```
mineralsfree, line 34: DeviceLimitWarning is also used in device auth passing the `hasContactInfo` prop. We need to either remove it or keep it if it's needed there

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
```diff
@@ -92,35 +80,36 @@ export const PreauthDialog = ({ acceptedDevices, deviceLimit, limitMaxed, onCanc
...
+        <ContentSection title="Identity data">
+          <KeyValueEditor onInputChange={convertIdentityToJSON} />
+        </ContentSection>
         {!!limitMaxed && <DeviceLimitWarning acceptedDevices={acceptedDevices} deviceLimit={deviceLimit} />}
         {!!submitError && <Alert severity="error">The device could not be added: {submitError}</Alert>}
       </DialogContent>
       <DialogActions>
         <Button onClick={onCancel}>Cancel</Button>
-        <Button variant="contained" disabled={isSubmitDisabled} onClick={() => onHandleSubmit(false)} color="primary" style={{ marginLeft: 10 }}>
+        <Button variant="outlined" disabled={isSubmitDisabled} onClick={() => onHandleSubmit(false)}>
```
mineralsfree, line 114: Clicking this should probably reset the form. Now it only shows Snackbar

mineralsfree (changes requested):
I'd fix the error handling first, some other smaller issues and suggestions included

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
mzedel: AFAIR the idea was to add multiple devices that only differ in their identity attributes, but get seeded the same certificate in the factory => kept it

frontend/src/js/components/devices/dialogs/PreauthDialog.tsx
mzedel: I don't think anyone is aware this is there + we have the notification in the header => removed the prop

frontend/src/js/common-ui/forms/KeyValueEditor.test.tsx
mzedel: I guess I've been looking at too much e2e tests 🙃

mineralsfree (dismissed):
I think you might've missed the empty file comment, otherwise looks good 👍

mineralsfree (approved):

PR #2205, opened 2026-08-26
feat(deployments): add request timeout middleware to devices APIs

alfrunes:
Adds configuration `request_timeout` (duration) for setting the timeout for reading request headers as well as request handler context.

Ticket: MEN-10042

### DIFF
--- a/backend/pkg/middlewares/timeout.go
+++ b/backend/pkg/middlewares/timeout.go
@@ -1,0 +1,44 @@
+// Copyright 2026 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package middlewares
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"time"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+// Timeout wraps the http.Request.Context with a fixed timeout.
+func Timeout(duration time.Duration) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		ctx, cancel := context.WithTimeout(c.Request.Context(), duration)
+		defer cancel()
+		c.Request = c.Request.WithContext(ctx)
+		c.Next()
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) &&
+			!c.Writer.Written() {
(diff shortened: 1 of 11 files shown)


backend/pkg/middlewares/timeout.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/gin-gonic/gin"
+
+	"github.com/mendersoftware/mender-server/pkg/rest.utils"
+)
+
+// Timeout wraps the http.Request.Context with a fixed timeout.
+func Timeout(duration time.Duration) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		ctx, cancel := context.WithTimeout(c.Request.Context(), duration)
+		defer cancel()
```
frodeha, line 32: This might be YAGNI territory, but do you think we maybe should respect an existing deadline in the context (aka only wrap the context if it doesn't have a timeout already)? Just a thought, I'm not sure myself so feel free to disregard.

backend/services/deployments/config/config.go
```diff
@@ -155,6 +156,9 @@ const (
 	// Max Request body size
 	SettingMaxRequestSize        = "request_size_limit"
 	SettingMaxRequestSizeDefault = 1024 * 1024 // 1 MiB
+
+	SettingRequestTimeout        = "request_timeout"
+	SettingRequestTimeoutDefault = time.Minute
```
frodeha, line 161: This should be added to the config.yaml as documentation no? Also, should we maybe make it clear somehow that this only applies to device APIs (which is the case afaict?)?

backend/services/deployments/api/http/api_deployments.go
```diff
@@ -159,6 +161,7 @@ func NewConfig() *Config {
 		MaxImageSize:        DefaultMaxImageSize,
 		MaxGenerateDataSize: DefaultMaxGenerateDataSize,
 		MaxRequestSize:      dconfig.SettingMaxRequestSizeDefault,
+		RequestTimeout:      dconfig.SettingRequestTimeoutDefault,
```
frodeha, line 164: Not setting this in the [defaults of the config package](https://github.com/mendersoftware/mender-server/blob/main/backend/services/deployments/config/config.go#L290) strikes me as a loaded foot-gun just waiting to take someone's toe off. I realize it doesn't make any difference currently, but could we please consider setting this in the config package defaults as well (or instead?) and save some unfortunate future engineer from a lifetime of poor balance and nine toes? 🦶

backend/services/deployments/config/config.go
alfrunes: The intention is to extend this to all APIs... But maybe we should have separate configs for devices and management APIs? :thinking:

backend/services/deployments/api/http/api_deployments.go
alfrunes: Oops, that was not intentional :face_with_peeking_eye:

backend/services/deployments/config/config.go
alfrunes: Currently it's also setting the (global) server timeout for reading request headers too.

alfrunes:
@frodeha I split the configuration into two for read_headers (server config) and devices API handler timeout.

frodeha (approved):
Looks great. Nice! 👍

PR #2215, opened 2026-08-29
Fix/phased deployments followups

mzedel:


frontend/src/js/components/deployments/deployment-wizard/RolloutOptions.tsx
mineralsfree, line 94: Maybe use this const in Global.tsx as well? Or define there and use here?

mineralsfree (approved):
LGTM, just a small suggestion to make sure the limit is aligned in both places

frontend/src/js/components/deployments/deployment-wizard/RolloutOptions.tsx
mzedel: created a follow up PR -> https://github.com/mendersoftware/mender-server/pull/2218

PR #2216, opened 2026-08-31
test: do not use deprecated endpoints and fields

kjaskiewiczz:


backend/services/deployments/tests/tests/test_release_v2.py
```diff
@@ -115,20 +115,23 @@ def test_get_all_releases(self, mongo, cli):
...
-                [
-                    ("foo", "device-type-1", "app"),
-                    ("foo", "device-type-2", "single-file"),
-                    ("bar", "device-type-2", "directory"),
-                ]
-            ):
-                types = management_v2_client(jwt=self.d.get_jwt()).list_release_types()
-                assert len(types) > 0
-                assert types == ["rootfs-image", "app", "single-file", "directory"]
+            with artifacts_added_from_data([("baz", "device-type-1")]):
```
alfrunes: Why was this added? (does not seem to align with commit message)

backend/tests/pyproject.toml
```diff
@@ -1,5 +1,9 @@
 [tool.pytest.ini_options]
 junit_family = "xunit1"
 minversion = "6.0"
-markers = ["storage_test: marks tests that uses artifact storage", "slow: marks tests as slow running"]
+markers = [
+    "storage_test: marks tests that uses artifact storage",
+    "slow: marks tests as slow running",
+    "deprecated_api: marks tests that deliberately exercise a deprecated API; delete together with the endpoint/field they cover",
```
alfrunes: This marker does not seem to be in use here, but I believe we have integration tests using deprecated APIs.

backend/services/deployments/tests/tests/test_release_v2.py
kjaskiewiczz: Sorry, that was leftover from some debugging session (unnecessary, but still correct though). When I was moving tests around this test started failing, and it was failing because it was using data from different tests  - it didn't use "clean_db" fixture. After I moved some tests and before I figured out it was expecting data from other tests, I added this artifact, which created "rootfs-image" (expected by this test, but no longer provided by previous test).
Long story short - removed.

backend/tests/pyproject.toml
kjaskiewiczz: I don't think we have integration tests directly testing deprecated APIs (and with this PR, we don't have integration tests using deprecated APIs at all) in mender-server. We have few in enterprise.
I removed this marker (I will only introduce it in enterprise).

backend/tests/integration/test_deployments.py
```diff
@@ -640,8 +642,10 @@ def status_verify(
...
 
         if deployment_status != "":
-            resp = self.deploymentsm.with_auth(user_token).call(
-                "GET", deployments.URL_DEPLOYMENTS.format(id=deployment_id)
+            resp = self.deploymentsm_v2.with_auth(user_token).call(
+                "GET",
+                deployments_v2.URL_DEPLOYMENTS,
+                qs_params={"id": deployment_id},
             )
             resp.status_code == 200
```
p-targowicz: I know its not your code but this does nothing - if we want to validate then 
```
assert resp.status_code == 200
```

can we pls change it too ?

backend/tests/integration/test_deployments.py
```diff
@@ -626,7 +626,9 @@ def status_verify(
 
         if device_deployment_status != "":
             resp = self.deploymentsm.with_auth(user_token).call(
-                "GET", deployments.URL_DEPLOYMENTS_DEVICES.format(id=deployment_id)
+                "GET",
+                deployments.URL_DEPLOYMENTS_DEVICES_LIST.format(id=deployment_id),
+                qs_params={"per_page": 500},
             )
             resp.status_code == 200
```
p-targowicz: same as below https://github.com/mendersoftware/mender-server/pull/2216/changes#r3908474556 can you pls change it ?

p-targowicz (dismissed):
Overall its looking good apart these kinda pointless `resp.status_code == 200` if you'd have some time to check and change it to either assertion or remove as now it does nothing it'd be great !

backend/tests/pyproject.toml
alfrunes: It's ok to have it for future deprecations and alignment. But also ok not to have it in this case :)

alfrunes (dismissed):

p-targowicz:
@kjaskiewiczz legend tahnk you for fixing it 💯

p-targowicz (approved):

PR #2217, opened 2026-09-01
Fix/reduce theme flash

mzedel:
~now depends on: https://github.com/NorthernTechHQ/nt-gui/pull/905 #2219~

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -19,7 +19,7 @@
         "@mui/material": "9.3.1",
         "@mui/x-charts": "9.12.0",
         "@mui/x-date-pickers": "9.12.0",
-        "@northern.tech/store": "0.46.0",
+        "@northern.tech/store": "0.46.1",
         "@northern.tech/themes": "0.16.9",
         "@northern.tech/utils": "0.14.0",
         "@reduxjs/toolkit": "2.12.0",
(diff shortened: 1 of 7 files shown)


mineralsfree:
This will move so many people to the dark mode. Maybe we should finish theme migration first to ensure we are not missing any area? 😄

mzedel:
> This will move so many people to the dark mode. Maybe we should finish theme migration first to ensure we are not missing any area? 😄

We have the dark mode available already & they would have the possibility to toggle it - for me it's mostly a question if this might be helpful already instead of adding the loading screen 🙃
also note: the ones with dark mode browser preference + light mode in Mender will now see a brief dark version first - although that could also be handled if it actually is an issue...

mineralsfree:
> > This will move so many people to the dark mode. Maybe we should finish theme migration first to ensure we are not missing any area? 😄
> 
> We have the dark mode available already & they would have the possibility to toggle it - for me it's mostly a question if this might be helpful already instead of adding the loading screen 🙃 also note: the ones with dark mode browser preference + light mode in Mender will now see a brief dark version first - although that could also be handled if it actually is an issue...

I use light mode in Mender UI because I don't like dark mode in Mender, but I use dark mode elsewhere. The change is fine, but maybe it's better if @chiachenglu or @michaelntech approve it

michaelntech:
I do like the idea of using the browser/system theme like this, but I think it would be better to plan it a bit before making the change. Ideally I think we should go for a combination of doing this, plus: 

- making the theme switch easier to find (in the user dropdown?)
- adding a 'use system default' option to it
- using local storage to show the login screen in the chosen theme when the user setting is unavailable (which requires providing a dark mode login screen; actually it would be good to have this even just for the browser preference change )
- and using the splash screen to cover the initial load and any theme flashing not covered by the other changes.

So from my POV, good idea but let's coordinate a bit first!

mzedel:
@michaelntech & @mineralsfree the user preference dependency is gone now, now it's purely reliant on the user theme settings & makes the selected theme show before the settings have arrived on reloads/ when opening the page in a new tab

frontend/src/js/components/header/Header.tsx
```diff
@@ -192,7 +192,12 @@ const AccountMenu = () => {
 
   const onLogoutClick = () => {
     setAnchorEl(null);
-    dispatch(logoutUser()).then(() => window.location.replace('/ui/'));
+    dispatch(logoutUser())
+      .unwrap()
+      .then(() => {
+        window.localStorage.removeItem(settingsKeys.colorScheme);
```
mzedel: this would likely be better suited in the `store` package...

mzedel:

frontend/src/js/components/header/Header.tsx
mzedel: it now is!

mzedel:

mzedel:

mineralsfree (approved):

PR #2220, opened 2026-09-01
Request logger improvements

alfrunes:
* Increase time resolution to millisecond
* Quote empty fields to ensure correct logfmt parsing
* Remove redundant `ts` field from accesslogs
* Accesslog shows 499 for clients closing connections.


oldgiova (commented):
Thanks for dealing with this! Can't honestly approve, but the approach seems good to me

frodeha (approved):
LGTM 👍

PR #2232, opened 2026-09-03
fix(gui): fetch tenant token in device onboarding modal

mineralsfree:
regression after migrating from the old get-organization endpoint:
image


mzedel (approved):
🤦 and I even commented on it in the ticket - thank you!

PR #2233, opened 2026-09-04
chore(gui): integration settings theme adjustments

mineralsfree:
Ticket: MEN-9334


frontend/src/js/components/settings/Integrations.tsx
```diff
@@ -13,185 +13,177 @@
...
+        label={label}
+        multiline
+        onBlur={onBlur}
+        onChange={onChange}
+        value={value}
+        helperText={hint}
+      />
+    )}
+  />
+);
```
mzedel: Instead of this, the `TextInput` from `common-ui` would help reusing what we have 👌

frontend/src/js/components/settings/Integrations.tsx
```diff
@@ -287,6 +288,11 @@ export const Integrations = () => {
   }
   return (
     <IntegrationsContainer>
+      {!isConfiguring && (
+        <Alert severity="info" className="margin-bottom-small" style={{ width: SETTINGS_CONTENT_MAX_WIDTH }}>
```
mzedel: Maybe move the `style={{ width: SETTINGS_CONTENT_MAX_WIDTH }}` into the `IntegrationsContainer` to have any other content capped at that the width...

mzedel (commented):
Nice! only the ConfigField seems a bit too general to not be using `TextInput`

mzedel (approved):

PR #2240, opened 2026-09-07
Redis cluster hardening

oldgiova:
Hardening redis cluster connection, dealing different topology changes

Co-authored-with: Claude

### DIFF
--- a/backend/pkg/redis/redis.go
+++ b/backend/pkg/redis/redis.go
@@ -186,7 +186,7 @@ func parseRedisExtraOptions(ctx context.Context, redisurl *url.URL) (
 	// Use cluster mode if `cluster` querystring is truthy or additional
 	// addr parameters are supplied.
 	if _, ok := q["cluster"]; ok {
-		isCluster, _ = strconv.ParseBool("cluster")
+		isCluster, _ = strconv.ParseBool(q.Get("cluster"))
 		delete(q, "cluster")
 	} else {
 		_, isCluster = q["addr"]
--- a/backend/pkg/redis/redis.go
+++ b/backend/pkg/redis/redis.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/redis/go-redis/v9"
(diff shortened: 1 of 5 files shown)


backend/pkg/redis/redis.go
```diff
@@ -183,32 +196,80 @@ func parseRedisExtraOptions(ctx context.Context, redisurl *url.URL) (
...
+}
+
+func (h *clusterHardeningHook) startBackgroundReload() {
+	go func() {
+		ticker := time.NewTicker(clusterBackgroundReloadInterval)
+		defer ticker.Stop()
+		for range ticker.C {
+			h.rdb.ReloadState(context.Background())
+		}
+	}()
```
alfrunes, line 271: I believe this does not belong to the hook (it's starting an async go routine) and should already be covered by the existing cluster option (ClusterStateReloadInterval) and the rest of the hook implementation (reload on X failures).

oldgiova: Moved `startBackgroundReload` out of the hook.

Regarding the option, as far as I understood, the passive interval checks only runs as a side effect of a command being dispatched (`clusterStateHolder.Get()`), and the forced reload only fires after a command has already failed. 
The ticker, instead, is the only piece that refreshes proactively regardless of traffic...

Happy to investigate more, though

backend/pkg/redis/redis.go
```diff
@@ -186,7 +186,7 @@ func parseRedisExtraOptions(ctx context.Context, redisurl *url.URL) (
 	// Use cluster mode if `cluster` querystring is truthy or additional
 	// addr parameters are supplied.
 	if _, ok := q["cluster"]; ok {
-		isCluster, _ = strconv.ParseBool("cluster")
+		isCluster, _ = strconv.ParseBool(q.Get("cluster"))
```
alfrunes, line 189: 🫣

backend/pkg/redis/redis.go
alfrunes: My question is more whether it's necessary to reload the cluster state unconditionally every 5 seconds. It seems like the hook would do the job when it's necessary.

alfrunes (approved):

PR #2259, opened 2026-09-15
MEN-9331: Webhook adjustments

mineralsfree:


frontend/src/js/components/settings/webhooks/Webhooks.tsx
```diff
@@ -50,14 +51,8 @@ export const Webhooks = () => {
...
-      <h2>Webhooks</h2>
-      {webhooks.length ? (
-        <DetailsTable columns={columns} items={mappedWebhooks} onItemClick={setSelectedWebhook} />
-      ) : (
-        <div className="flexbox centered">
-          No webhooks are configured yet. Learn more about webhooks in our <DocsLink path="server-integration" title="documentation" />
-        </div>
-      )}
+      <Typography variant="subtitle1">Webhooks</Typography>
+      <DetailsTable columns={columns} items={mappedWebhooks} onItemClick={setSelectedWebhook} />
```
mzedel: I guess this _could_ be a `<ContentSection/>`

frontend/src/js/components/settings/webhooks/Management.tsx
```diff
@@ -95,7 +92,15 @@ export const WebhookManagement = ({ onCancel, onRemove, webhook }) => {
...
   const dispatch = useAppDispatch();
   const { classes } = useStyles();
-  const containerRef = useRef();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const lastSelectedEvent = useRef<Event>(undefined);
+
+  const onEventSelect = useCallback((event?: Event) => {
+    if (event) {
+      lastSelectedEvent.current = event;
+    }
```
mzedel: maybe use separate state to handle the visibility/ detach the content rendering in the details from the return to the events list?

frontend/src/js/components/settings/webhooks/Management.tsx
```diff
@@ -139,31 +144,33 @@ export const WebhookManagement = ({ onCancel, onRemove, webhook }) => {
...
-              classes={classes}
-              columns={columns}
-              events={events}
-              eventsTotal={eventsTotal}
-              getWebhookEvents={dispatchedGetWebhookEvents}
-              setSelectedEvent={setSelectedEvent}
-              webhook={webhook}
-            />
+      <div className={classes.slideContainer} ref={containerRef}>
+        <Slide appear={false} in={!selectedEvent} container={() => containerRef.current} direction="right">
```
mzedel: More out of curiosity: is there any specific reason the container is now returned in a function (I think I took this straight from the docs at the time https://mui.com/material-ui/transitions/#slide-relative-to-a-container)?

frontend/src/js/components/settings/webhooks/Configuration.tsx
```diff
@@ -112,45 +114,56 @@ const WebhookConfiguration = ({ onCancel, onSubmit }: { onCancel: () => void; on
...
+        <Typography variant="subtitle1">Target URL</Typography>
+        <UrlInput
+          hint="URL"
+          InputLabelProps={{ shrink: true }}
+          required
+          requiredRendered={false}
+          id="url"
+          validations="isLength:1,isURL"
+          helperText="This URL will receive the events."
+        />
```
mzedel: I guess these and the other inputs _could_ also go into `ContentSection`s

frontend/src/js/components/settings/webhooks/Management.tsx
```diff
@@ -95,7 +92,15 @@ export const WebhookManagement = ({ onCancel, onRemove, webhook }) => {
...
   const { classes } = useStyles();
-  const containerRef = useRef();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const lastSelectedEvent = useRef<Event>(undefined);
+
+  const onEventSelect = useCallback((event?: Event) => {
+    if (event) {
+      lastSelectedEvent.current = event;
+    }
+    setSelectedEvent(event);
```
mzedel: maybe use separate state to handle the visibility/ detach the content rendering in the details from the return to the events list? at least to me the ref usage always is a bit of a workaround if there's another more react way around 🤷...
(only noting on this as I think we'll get to adjust more such situations in the near future e.g. https://northerntech.atlassian.net/browse/MEN-10048 🫣)

mzedel (dismissed):
Looks nice, a bit verbose - but to me fits in more than good enough while we wait for the design alignment 👌.

comments more out of curiosity...

frontend/src/js/components/settings/webhooks/Management.tsx
mineralsfree: It's to make sure the ref holds a real value (access defered until slide transition happens). Since containerRef initialized during first render, the null is passed to container initially. Probably not a real issue here, but maybe a bit more robust in general...

mzedel (approved):

PR #2263, opened 2026-09-16
MEN-10082

mzedel:


frontend/pipeline.yml
mineralsfree: nitpick, but maybe `   - npm run script --prefix frontend/tests/e2e_tests -- --environment enterprise --variant ${TEST_VARIANT}
`
And then acceptance:enterprise:orchestrator
can become 
```
test:frontend:acceptance:enterprise:orchestrator:
  extends: test:frontend:acceptance:enterprise:qemu
  variables:
    TEST_VARIANT: orchestrator
```

frontend/tests/e2e_tests/integration/06-orchestrator-dependent/01-orchestrator.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+import * as fs from 'fs';
+
+import test, { expect } from '../../fixtures/fixtures';
+import { timeouts } from '../../utils/constants';
+import { acceptPendingDevice, selectReleaseByName, triggerDeploymentCreation } from '../../utils/utils.ts';
+
+const rtosArtifactUrl = 'https://raw.githubusercontent.com/mendersoftware/mender-orchestrator-support/main/demo/premade-artifacts/rtos-v2.mender';
```
mineralsfree: Is there an issue of keeping those files in fixtures instead? I think we should rely on gh services availability less 😄

frontend/tests/e2e_tests/integration/06-orchestrator-dependent/01-orchestrator.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    await page.getByText('finished').click();
+    await page
+      .getByRole('listitem')
+      .first()
+      .waitFor({ timeout: 10 * timeouts.sixtySeconds });
+
+    await expect(async () => {
+      await navbar.getByRole('link', { name: /devices/i }).click();
+      await page.getByRole('tab', { name: /system/i }).click();
+      await expect(page.getByRole('cell', { name: targetComponentVersion }).first()).toBeVisible();
```
mineralsfree: won't it also pass on failed deployment?

frontend/tests/e2e_tests/utils/utils.ts
```diff
@@ -13,7 +13,35 @@
...
+  let hasAcceptedDevice = false;
+  try {
+    await page.waitForSelector(`css=${selectors.deviceListItem}`, { timeout: timeouts.default });
+    hasAcceptedDevice = await page.isVisible(selectors.deviceListItem);
+  } catch {
+    console.log(`no accepted device present so far`);
+  }
+  if (!hasAcceptedDevice) {
+    const pendingMessage = page.getByText(/pending authorization/i);
+    await pendingMessage.waitFor({ timeout: 3 * timeouts.sixtySeconds });
```
mineralsfree: this timeout is longer than one of the consumer test timeout 😄 maybe worth unifying

mineralsfree (dismissed):

frontend/tests/e2e_tests/integration/06-orchestrator-dependent/01-orchestrator.spec.ts
mzedel: the main reason used to be that using the client components from there or `main`/`master` in case of images etc. that we got notified early about unexpected changes or movements (as we had e.g. unexpected changes in the demo artifact content)... I guess the process for this has gotten better over time, although I think there's still no way we could track changes with renovate or such - nevertheless we could give it a try with local fixtures 👍 (but I'll put it in a follow up PR)

PR #2264, opened 2026-09-17
MEN-9305: split dynamic and static groups creation

mineralsfree:
Ticket: MEN-9305


frontend/src/js/components/devices/group-management/CreateDynamicGroup.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { BaseDialog } from '@northern.tech/common-ui/dialogs/BaseDialog';
+import Form from '@northern.tech/common-ui/forms/Form';
+import TextInput from '@northern.tech/common-ui/forms/TextInput';
+import { groupNameValidationRules } from '@northern.tech/common-ui/forms/validations';
+import { getGroups } from '@northern.tech/store/selectors';
+
+export const CreateDynamicGroup = ({ onClose, onCreate }) => {
+  // a dynamic group can't reuse the name of any existing group, static or dynamic
+  const { dynamic, static: staticGroups } = useSelector(getGroups);
+  const groups = [...staticGroups, ...dynamic].map(g => g.groupId);
```
mzedel, line 27: maybe
```suggestion
  const groups = useSelector(getGroupNames);
```

frontend/src/js/common-ui/forms/validations.ts
```diff
@@ -116,3 +117,36 @@ export const hasValidTagCharacters = (value: string) => validator.isWhitelisted(
 export const tagValidationRules = {
   validate: (tags = []) => tags.every(hasValidTagCharacters) || invalidCharactersError
 };
+
+const GROUP_NAME_LENGTH_LIMIT = 256;
```
mzedel, line 121: this and the other validation functionality are _very_ group specific and so far the rest of the validations are more general purpose focused (although the tag validation things would likely be better with a different name)...
I would imagine you can get the length requirement with `isLength:1:256`, the character validation with `isAlphanumericLocator` (which would also be more appropriate given the backend definition there: https://github.com/mendersoftware/mender-server-enterprise/blob/main/backend/pkg/rules/patterns.go#L14-L16), the ungrouped group with `isNot:${UNGROUPED_GROUP.name}` - IOW:

```
validations={`isLength:1:256,isNot:${UNGROUPED_GROUP.name},isAlphanumericLocator`}
```

 and then only process the existing groups & `selectedDevices` overlap?

mzedel (dismissed):
Looks good, but I'm not convinced moving all the group specific validation logic into common-ui territory makes sense here as it's not that _common_...

frontend/src/js/components/devices/group-management/GroupDefinition.test.tsx
```diff
@@ -11,43 +11,51 @@
...
 //    limitations under the License.
-import { render } from '@/testUtils';
+import { formRenderWrapper } from '@/testUtils';
 import { undefineds } from '@northern.tech/testing/mockData';
 
-import GroupDefinition, { validateGroupName } from './GroupDefinition';
+import GroupDefinition, { getGroupNameError, groupNameValidationRules } from './GroupDefinition';
 
-const selectedDevices = [{ id: 'test' }];
+const invalidCharactersError = 'This please only enter valid characters. Valid characters are a-z, A-Z, 0-9, _ and -';
```
mzedel: 🙈

mzedel (dismissed):

frontend/src/js/components/devices/group-management/GroupDefinition.test.tsx
mineralsfree: You are first to notice it since 2024 😄

mzedel (approved):

PR #2270, opened 2026-09-18
chore(gui): display buttons on the left side in forms

mineralsfree:
+ remove unused cancelButton class

Ticket: MEN-10023


frontend/src/js/common-ui/forms/Form.tsx
```diff
@@ -39,7 +38,7 @@ export const Form = ({
   submitRef,
   validationMode = 'onChange'
 }) => {
-  const { classes: internalClasses } = useStyles();
+  const { classes: internalClasses, cx } = useStyles();
```
mzedel: This is similar to the `disableMargin` approach for the `ContentSection` & not really motivated by anything than personal preference AFAICT

I guess if you want to ensure the class name stays in sync, we could do:

```
export const buttonClasses = {
  wrapper: 'button-wrapper'
};
```

and then rely on this in the consumers the same way we do for the MUI style overrides...

mzedel (dismissed):
Looks good in general - not a fan of yet another round of personal preference induced churn, especially as increased `cx` usage could end up with more stylesheets added to the page/ increased rendering cost...

frontend/src/js/common-ui/forms/Form.tsx
mineralsfree: Dropped the commit with this change 👌

mzedel (approved):

PR #2276, opened 2026-09-21
MEN-9361: use select for determenistic lists

mineralsfree:


frontend/src/js/common-ui/forms/Select.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const selectionRenderer = renderValue ?? (!children && (hasPlaceholder || multiple) ? renderSelection : undefined);
+
+  return (
+    <FormControl disabled={disabled} error={error} style={{ width }}>
+      {!!label && <InputLabel id={labelId}>{label}</InputLabel>}
+      <MuiSelect
+        autoWidth={autoWidth}
+        displayEmpty={displayEmpty ?? hasPlaceholder}
+        label={label}
+        labelId={label ? labelId : labelIdProp}
```
mzedel:
```suggestion
        labelId={labelId}
```
since the `labelIdProp` might otherwise get a chance to be unset & used? OTOH if there's no `label`, there's likely also no `labelIdProp` 🤷

frontend/src/js/common-ui/forms/Select.tsx
```diff
@@ -0,0 +1,115 @@
...
+        MenuProps={{ ...defaultMenuProps, ...MenuProps }}
+        multiple={multiple}
+        renderValue={selectionRenderer}
+        value={selectValue}
+        {...remainder}
+      >
+        {children}
+        {!children && hasPlaceholder && !hideEmptyOption && !multiple && (
```
mzedel: seeing this, I'm wondering if we shouldn't rather get rid of the `hideEmptyOption` and align with MUI on `displayEmpty`... (IOW get rid of the `placeholder` in the `<ControlledSelect />`s in `ReleaseArtifactFilter` & instead use `displayEmpty` + a `<MenuItem value="" ...` there...

mzedel (approved):
Looks good as is, only the `hideEmptyOption` seems to grow a bit too involved in comparison to the default MUI approach...

frontend/src/js/common-ui/forms/Select.tsx
mineralsfree: Good suggestion. I will follow up in the next PR

PR #2277, opened 2026-09-21
fix(gui): submit only roles when changing another user's roles

nickanderson:
Changing another user's role submitted the whole user object, so an unchanged
`email` came along and useradm rejected the update with 422 `updating user email
addresses is not permitted` (MEN-10010, shipped in 4.2.0-saas.15).

Pushing the failing test on its own first to show it red; the fix follows in a
second commit.

Ticket: ME-762


mzedel (approved):

mzedel (changes requested):
as @alfrunes rightly pointed out with https://github.com/mendersoftware/mender-server-enterprise/pull/1698 - on a PUT this would be wrong & I should have checked the endpoint here 😕

nickanderson:
OK so it's fixed in the backend now then? Closing.

